PR #12, opened 2018-04-18
Allowed queued requests. 

lakinwecker:


main.py
```diff
@@ -100,6 +110,11 @@ def play_game(li, game_id, weights, threads):
                 get_engine_stats(info_handler)
 
     print("Game over!")
+    try:
```
lakinwecker: This try/except should probably also exist in the event loop of the start method. It is possible for the workers to be full, but not the queue and for this try loop to not get the new challenge, and the challenge is inserted into the queue right after this worker exits.

So checking the results to see if we have room to play and accepting any waiting challenges at the end of that loop is probably better.

PR #22, opened 2018-04-19
XBoard Support Redux

ZirconiumX:
This took quite a bit of rebasing, but I got there in the end.

This code is quite messy because the XBoard and UCI APIs are different (I don't see why they should be), hopefully it can be cleaned up once merged.

### DIFF
--- a/config.yml.default
+++ b/config.yml.default
@@ -1,9 +1,10 @@
 token: "xxxxxxxxxxxxxxxx"    # lichess OAuth2 Token
 url: "https://lichess.org/"  # lichess base URL
 
-engine:                       # engine settings
+engine:                      # engine settings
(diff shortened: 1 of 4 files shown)


careless25:
Hmm, this will get messy. Can you make a class EngineWrapper? Then have 2 other classes UCI and Xboard inherit from it. Will make the code cleaner and easier to read

ZirconiumX:
Sure, I can try that.

ornicar:
After that's merged, we should consider renaming the repo.

ZirconiumX:
How does this look?

ornicar:
Looking good. Please make sure UCI engines still work.

ZirconiumX:
Yeah, they work fine.

PR #27, opened 2018-04-19
XBoard option compatibility layer

ZirconiumX:
This tries to convert UCI options into the equivalent XBoard commands.

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -22,14 +22,14 @@ def create_engine(config, board):
         commands.append(str(threads))
 
     if engine_type == "xboard":
-        return XBoardEngine(board, commands)
+        return XBoardEngine(board, commands, config.get("xboardoptions"))
 
-    return UCIEngine(board, commands, ucioptions)
+    return UCIEngine(board, commands, config.get("ucioptions"))
 
 
 class EngineWrapper:
 
-    def __init__(self, board, commands):
+    def __init__(self, board, commands, options=None):
         pass
 
     def pre_game(self, game):
@@ -97,7 +97,7 @@ class UCIEngine(EngineWrapper):
 
 class XBoardEngine(EngineWrapper):
 
-    def __init__(self, board, commands):
+    def __init__(self, board, commands, options=None):
         commands = commands[0] if len(commands) == 1 else commands
         self.engine = chess.xboard.popen_engine(commands)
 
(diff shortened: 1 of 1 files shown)


ZirconiumX:
@careless25 @ornicar Can you check this?

careless25:
@ZirconiumX Have you tested it?

ZirconiumX:
Rebased and squashed to master.

careless25:
Can you fix the merge conflicts?

ZirconiumX:
Done.

careless25:
Can anyone test this? Or link me to a xboard engine?

mroh69:
how about adding something like "xboardoptions" to the config instead of trying to convert (and guess) the uci options to xboard?

ZirconiumX:
That feels like trying to reinvent the wheel, because having separate
xboard/UCI sections will cause code duplication.

On 21 Apr 2018 21:27, "Michael Rohleder" <notifications@github.com> wrote:

how about adding something like "xboardoptions" to the config instead of
trying to convert (and guess) the uci options to xboard?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/careless25/lichess-bot/pull/27#issuecomment-383327321>,
or mute the thread

mroh69:
I think its much less code duplication. maybe its bit duplication in the config, but the code is much less.
for example, the config could look like this:

```yml
# this holds the option for uci *or* xboard
# user has to comment out the options he doesnt want (uci or xboard)
engineoptions:     
  # commented as current "ucioptions:"
  # Threads: 2
  # Hash: 2048
  # xboard options
  cores: 2
  memory: 2048
  egtpath: "path"
  some other option: "bla"
```

and in the code its just (as for uci)
```python
engine.xboard.option(CONFIG['engineoptions'])
```

instead of
```python
if option == "could be that name":
  engine.xboard.egtpath(value)
if option == "could be other name"
  engine.xboard.egtpath(value)
if option == "Threads"
  engine.xboard.memory(value)
if option == "Move Overhead"   # uci name in some engines
...
if option == "Overhead"   # uci name of other engines
...
if option == "an option I dont know"
...
etc pp for so many uci options in so many variants...
```

careless25:
+1 For @mroh69 suggestion. Keep the engine config separate instead of the if statement conversions.

ZirconiumX:
Uh, no. That's not even slightly true, and my code should demonstrate that.

If I use a separate `xboardoptions` config, then I *still* need to use if statement conversions to convert from options to XBoard commands.

It gains nothing other than the config being bigger.

gbtami:
@ZirconiumX can you update this patch please? I'm using  it and I hope maintainers will merge it :)

ZirconiumX:
I kinda ballsed it up, but I think I know how to fix it; first I need to
get home from school though. Hopefully Thibault and Shail can avoid giving
me another 60 patches to rebase against for a few hours.

ZirconiumX:
This should be cleaner, even if it took manually editing a patch file to get something that looked reasonable.

gbtami:
Thx!

PR #40, opened 2018-04-20
FIxing references to global CONFIG. Adding engine factory

lakinwecker:
The Engine Factory and config should probably be updated so that the specific engine options are a subconfig of the engine key. The specific class for each engine should probably do the args parsing for itself, rather than this method, but that can come in a separate request. 

I also removed the useless pool from the watch_control_stream

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -3,31 +3,28 @@ import chess
 import chess.xboard
 import chess.uci
 
-class EngineFactory:
-    def __init__(self, config):
-        self.config = config
-    def __call__(self, board):
-        # print("Loading Engine!")
-        cfg = self.config["engine"]
-        engine_path = os.path.join(cfg["dir"], cfg["name"])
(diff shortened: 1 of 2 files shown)


engine_wrapper.py
```diff
@@ -1,7 +1,35 @@
+import os
 import chess
 import chess.xboard
 import chess.uci
 
+class EngineFactory:
+    def __init__(self, config):
+        self.config = config
+    def __call__(self, board):
```
ornicar: Great demonstration that "factory" objects are in fact just partially applied functions #haskell

lakinwecker: Yes, and I could have written it that way in python even.

lakinwecker: https://github.com/lakinwecker/lichess-uci-bot/commit/b91b7a19b998ca19e60fc530ca27eb3500e5ca16

PR #42, opened 2018-04-20
Use the same session for Lichess API requests

fli:
Changes non-streaming Lichess requests to use a Session object to improve performance with connection pooling.

Future changes can move the auth header into the Session object too, but I'll keep this PR small.

PR #47, opened 2018-04-21
Stop challenges from going through backoff if they have been withdrawn

fli:
Due to a side-effect of #42 challenges which are either:
- next to be accepted in the queue but have been withdrawn, or
- withdrawn before the bot can decline them
will return a 404 when the POST request is sent to the lichess api.

Because of the backoff code in #42, each withdrawn challenge can take 20s to process.
This is a kinda ugly fix but this PR excludes them from the backoff code.

### DIFF
--- a/lichess.py
+++ b/lichess.py
@@ -40,9 +40,18 @@ class Lichess():
         response.raise_for_status()
         return response.json()
 
+    # don't want to wait for max_time to accept/decline challenges if they have
+    # been withdrawn, so skip these and handle elsewhere
+    def check_challenge_404(exception):
+        return (isinstance(exception, HTTPError) and
+                exception.response.status_code == 404 and
+                "challenge" in exception.response.url)
+
     @backoff.on_exception(backoff.expo,
         (RemoteDisconnected, ConnectionError, ProtocolError, HTTPError),
-        max_time=20)
+        max_time=20,
+        giveup=check_challenge_404
+        )
     def api_post(self, path, data=None):
         url = urljoin(self.baseUrl, path)
         response = self.session.post(url, data=data)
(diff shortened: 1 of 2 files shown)


lichess.py
```diff
@@ -40,9 +40,18 @@ def api_get(self, path):
         response.raise_for_status()
         return response.json()
 
+    # don't want to wait for max_time to accept/decline challenges if they have
+    # been withdrawn, so skip these and handle elsewhere
```
ornicar: Yes. More generally, only `5**` responses should be retried, never the `4**` responses.

main.py
```diff
@@ -61,8 +61,12 @@ def start(li, user_profile, max_games, max_queued, engine_factory, config):
...
                     print("    Queue {}".format(chlng.show()))
                 else:
-                    print("    Decline {}".format(chlng.show()))
-                    li.decline_challenge(chlng.id)
+                    try:
+                        response = li.decline_challenge(chlng.id)
+                        print("    Decline {}".format(chlng.show()))
+                    except HTTPError as exception:
+                        if exception.response.status_code == 404:
+                            print("    Withdrawn {}".format(chlng.show()))
```
ornicar: this blackholes other exceptions

main.py
```diff
@@ -75,12 +79,15 @@ def start(li, user_profile, max_games, max_queued, engine_factory, config):
...
-                    # TODO: Probably warrants better checking.
+                try:
+                    response = li.accept_challenge(chlng.id)
+                    print("    Accept {}".format(chlng.show()))
                     queued_processes += 1
                     print("--- Process Queue. Total Queued: {}. Total Used: {}".format(queued_processes, busy_processes))
+                except HTTPError as exception:
+                    if exception.response.status_code == 404:
+                        print("    Withdrawn {}".format(chlng.show()))
+
```
ornicar: this locks the program because it doesn't look up the next challenge

PR #54, opened 2018-04-21
Engine.Threads needs be type string

JABirchall:
Adding quotes to default config


ornicar:
I pushed this 20 minutes ago https://github.com/careless25/lichess-bot/commit/a34c8a58271b40031061f2082eb8bf389fc94611

PR #61, opened 2018-04-22
Add wait command

ddugovic:
Please review.
Deployed at https://lichess.org/@/GodelEscherBot/tv

### DIFF
--- a/conversation.py
+++ b/conversation.py
@@ -1,3 +1,5 @@
+from time import time
+
 class Conversation():
     def __init__(self, game, engine, xhr):
         self.game = game
@@ -14,7 +16,7 @@ class Conversation():
 
     def command(self, line, game, cmd):
         if cmd == "wait":
-            game.abort_at = game.abort_at + 60
+            game.abort_at = time() + 60
             self.send_reply(line, "Waiting 60 seconds...")
         if cmd == "name" or cmd == "engine":
             self.send_reply(line, self.engine.name())
--- a/conversation.py
+++ b/conversation.py
@@ -15,7 +15,7 @@ class Conversation():
         pass
 
     def command(self, line, game, cmd):
-        if cmd == "wait":
+        if cmd == "wait" and game.is_abortable():
             game.abort_at = time() + 60
             self.send_reply(line, "Waiting 60 seconds...")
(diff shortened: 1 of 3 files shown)


conversation.py
```diff
@@ -6,13 +6,16 @@ def __init__(self, game, engine, xhr):
...
         print("*** {} [{}] {}: {}".format(self.game.url(), line.room, line.username, line.text.encode("utf-8")))
         if (line.text[0] == self.command_prefix):
-            self.command(line, line.text[1:].lower())
+            self.command(line, game, line.text[1:].lower())
         pass
 
-    def command(self, line, cmd):
+    def command(self, line, game, cmd):
+        if cmd == "wait":
+            game.abort_at = game.abort_at + 60
```
ornicar: should be `time.time() + 60` to avoid cumulating wait time.

conversation.py
```diff
@@ -6,13 +8,16 @@ def __init__(self, game, engine, xhr):
...
         if (line.text[0] == self.command_prefix):
-            self.command(line, line.text[1:].lower())
+            self.command(line, game, line.text[1:].lower())
         pass
 
-    def command(self, line, cmd):
+    def command(self, line, game, cmd):
+        if cmd == "wait":
+            game.abort_at = time() + 60
+            self.send_reply(line, "Waiting 60 seconds...")
```
ornicar: the feedback is a good idea, but it should only be sent when the game is actually abortable.

model.py
```diff
@@ -53,6 +53,9 @@ def __init__(self, json, username, base_url, abort_at):
         self.white_starts = self.initial_fen == "startpos" or self.initial_fen.split()[1] == "w"
         self.abort_at = abort_at
 
+    def is_abortable(self):
+        return len(self.state["moves"]) < 6
```
careless25: Can we make "6" a config variable or define a global. No Magic numbers

ornicar: I don't see the harm of it. It's based on the UCI format, which doesn't allow moves longer than 5 chars.

If we don't want that, then we also don't want to split `moves` with a space char, because it's just a similar assumption. Are we also going to put the space in config or global?

PR #67, opened 2018-04-22
Play rated challenges first

ddugovic:
Discourages hording of rating points, and keeps opponents on their toes.


ornicar:
Are you sure this sorts rated (true) before casual (false)? I thought `false < true`

ddugovic:
Oops, you are correct and I mis-read:
https://github.com/careless25/lichess-bot/blob/6cad8fb2ffbba6986db7ec02b90b7e874b1addae/model.py#L20

I'll correct although I don't have an easy way to test this.

PR #71, opened 2018-04-24
fix first_search

wakamex:
fix crash with engine that requires opponent's time on first_search


ZirconiumX:
This is not a fix. This is not even *close* to a fix. This is a workaround for an engine bug, and it needs to be fixed in the engine.

Why do you even need `otim` anyway? You should use your own clock only.

ddugovic:
I agree this looks like an XY problem, perhaps motivated by #72 but even so, this doesn't look like a fix.

PR #74, opened 2018-04-25
display winrate

wakamex:
show winrate calculated from cp (inverse from here: https://github.com/glinscott/leela-chess/blob/27bc580e57a8e01d4d137ed65c0f325c4b7b4661/src/UCTSearch.cpp)

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -62,6 +62,12 @@ class EngineWrapper:
                 if stat == "score":
                     for k,v in info[stat].items():
                         str = "score: {}".format(v.cp)
+                        feval = 0.322978*math.atan(0.0034402*v.cp) + 0.5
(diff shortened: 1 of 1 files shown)


ddugovic:
Stockfish has a different formula: https://github.com/ornicar/lila/issues/1494

wakamex:
maybe that one could be dded to the main branch, while this one would fit under the leela branch. also stockfish estimates its winrate as a byproduct, while leela trains directly on it, estimating cp as a byproduct, making it more relevant in her case.

ddugovic:
Indeed, Leela's estimate should be more accurate, whereas SF estimates need to be retrained periodically.

careless25:
Can you resolve the conflict?

careless25:
Closing. I fixed it in the new version.

PR #83, opened 2018-04-26
Add -d (logging.DEBUG) command line argument

gbtami:


ornicar:
how about `-v`? It's a standard flag to increase verbosity.

gbtami:
I'm fine with it.

PR #87, opened 2018-04-26
Added options to select polyglot best move (update with the new commit)

tttony:
Solves https://github.com/careless25/lichess-bot/issues/84

I tested only with the default option and it worked, so test to let us know

### DIFF
--- a/config.py
+++ b/config.py
@@ -11,15 +11,15 @@ def load_config():
             raise e
 
         #[section, type, error message]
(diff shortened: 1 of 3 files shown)


careless25:
How would I test this?

tttony:
Sorry @careless25  that message was supposed to be in the issue, anyway this update is deprecated, I will close and see if I can update it later

Issue #72, opened 2018-04-24
Bad Request for url


gbtami:
PyChessBot losed this game on time because it's latest move can't get through in this position:
https://lichess.org/cv25KskE#58

The relevant log was:
```
DEBUG:chess.engine:<PopenProcess at 0x7f368a5782e8 (pid=2500)> >> move Rxh7
DEBUG:requests.packages.urllib3.connectionpool:https://lichess.org:443 "POST /bot/game/cv25KskE/move/b7h7 HTTP/1.1" 200 None
DEBUG:chess.engine:<PopenProcess at 0x7f368a5782e8 (pid=2500)> << force
DEBUG:chess.engine:<PopenProcess at 0x7f368a5782e8 (pid=2500)> << setboard 6r1/p6R/2p5/8/4Pk2/5P1B/PP1N1P1P/2K5 w - - 1 30
DEBUG:chess.engine:<PopenProcess at 0x7f368a5782e8 (pid=2500)> << time 81.0
DEBUG:chess.engine:<PopenProcess at 0x7f368a5782e8 (pid=2500)> << otim 339.0
DEBUG:chess.engine:<PopenProcess at 0x7f368a5782e8 (pid=2500)> << go
DEBUG:chess.engine:<PopenProcess at 0x7f368a5782e8 (pid=2500)> >> # Time left: 0.81 s; Planing to think for 2.00 s
DEBUG:chess.engine:<PopenProcess at 0x7f368a5782e8 (pid=2500)> >> 1 2853 0 39 Rxa7
DEBUG:chess.engine:<PopenProcess at 0x7f368a5782e8 (pid=2500)> >> 2 2846 4 302 Rf7+ Ke5 Rxa7
DEBUG:chess.engine:<PopenProcess at 0x7f368a5782e8 (pid=2500)> >> 3 2902 14 1240 Rf7+ Kg5 Rxa7
DEBUG:chess.engine:<PopenProcess at 0x7f368a5782e8 (pid=2500)> >> 4 2853 57 4636 Rf7+ Kg5 Rxa7 Kf4
DEBUG:chess.engine:<PopenProcess at 0x7f368a5782e8 (pid=2500)> >> move Rf7+
DEBUG:requests.packages.urllib3.connectionpool:https://lichess.org:443 "POST /bot/game/cv25KskE/move/h7f7 HTTP/1.1" 400 None
[...]

PR #88, opened 2018-04-26
Removed deceptious st (time control) command caused engine losing on …

gbtami:
…time in #72

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -147,7 +147,6 @@ class XBoardEngine(EngineWrapper):
 
     def first_search(self, board, movetime):
         self.engine.setboard(board)
-        self.engine.st(movetime / 1000)
         self.engine.time(movetime / 10)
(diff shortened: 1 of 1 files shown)


ZirconiumX:
This is not how it should have been fixed, I think.

Instead, the code should call `pre_game` (which should be renamed) after the first move, so that the `level` command overrides the `st` command.

gbtami:
Telling the truth this whole "go movetime 2000" (and xboard workaround attempts) should be dropped and apply auto abort on timeout for all BOT games.

ZirconiumX:
My own solution would be simply to require polyglot books, and then use those for at the very least the first move.

ddugovic:
How do polyglot books work for variants?

ZirconiumX:
As long as people can agree on a unique set of hash keys for a given
variant that are sufficiently random, Polyglot format can be used for
variants, too. It's just a key/value dictionary, with a weight.

PR #91, opened 2018-04-27
Revert "fix first_search"

ddugovic:
This reverts commit 6400e782f26c59c1552a9d9fbd55bb2d826f8d64.

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -82,7 +82,7 @@ class UCIEngine(EngineWrapper):
 
     def first_search(self, board, movetime):
         self.engine.position(board)
-        best_move, _ = self.engine.go(wtime=movetime,btime=movetime)
+        best_move, _ = self.engine.go(movetime=movetime)
         return best_move
 
     def search(self, board, wtime, btime, winc, binc):
@@ -147,8 +147,7 @@ class XBoardEngine(EngineWrapper):
 
     def first_search(self, board, movetime):
         self.engine.setboard(board)
-        self.engine.time(movetime / 10)
-        self.engine.otim(movetime / 10)
+        self.engine.st(movetime / 1000)
         return self.engine.go()
 
     def search(self, board, wtime, btime, winc, binc):
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -36,10 +36,7 @@ class EngineWrapper:
     def __init__(self, board, commands, options=None):
         pass
 
-    def pre_game(self, game):
-        pass
-
-    def first_search(self, movetime):
+    def first_search(self, game, board, movetime):
         pass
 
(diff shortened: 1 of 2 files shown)


gbtami:
This commit causes to send "st 2" after "level" and engines can lose on time in short time control games thinking they have to think 2 sec on every move.

From http://hgm.nubati.net/CECP.html
level MOVESPERSESSION BASETIME INCREMENT
st T

Two mutually exclusive commands (erasing each other's effect) for specifying the time control. In 'level' either MOVESPERSESSION equals 0 (for incremental and sudden-death TC), or INCREMENT is 0 (for classical TC). With 'st' the time control will be T seconds per move (not accumulating leftovers). The BASETIME parameter is in minutes, and can have the MIN:SEC notation. INCREMENT and T are in seconds. Engines are encouraged to understand a floating-point number for the INCREMENT, or at least to not choke on it.

ddugovic:
What happens after a subsequent:

https://github.com/ddugovic/lichess-bot/blob/fcf0d450209b5715775c953c0acec411f3ef1550/engine_wrapper.py#L153-L161

gbtami:
Time control and clock are two different thing. If a CECP engine got "st T" it will ignore time/otime because it was told that it have to use T for all moves.

ddugovic:
OK, what if a subsequent `level` command were sent (after a best move is received)?

gbtami:
I think both "st 2" then "level" or "level" then "level" can do the job. Maybe the latter is better  regarding in ultrabullet/bullet games thinking 2 sec on the very first move may be luxury.

ZirconiumX:
I think `st 2`, `go` then issuing a `level` afterwards is the most idiomatic way to do it.

On the other hand, XBoard engines tend to be much looser with the specification than UCI engines, and some workarounds may be required at some point.

ddugovic:
I agree with all comments and will update this PR to use `level` twice (and not use `st`) since XBoard engines likely support that protocol usage.

gbtami:
st 2 on first move in an ultrabullet game seems rather big handicap.

ZirconiumX:
Then just pass `game` to `first_search` and scale first move with time control, perhaps?

ornicar:
So do I merge this?

ZirconiumX:
I mean, for Dorpsgek, it'd actually search for only a third of a second with that TC, so it should be fine.

ddugovic:
@ornicar Looks good to me & I hear no objections to merging it.  It seems to issue correct commands although my `gnuchess` I'm testing with isn't respecting the `xboard` protocol.

ddugovic:
Successfully tested at https://listage.ovh/CpIv3rn9

gbtami:
Works OK with PyChessBot also.

Issue #79, opened 2018-04-25
Feature: Add option to resign


fsmosca:
Perhaps if uci bot wants to resign, it will send to api,

`bestmove 0000`

Followeb by

`info string resign`

api can terminate the game and award 1 point to the opponent.

I can revise my uci engine to support this.

For xboard engine, it has command to resign.

Or perhaps if api can access the score of the bot and the bot is sending the score properly to api, api can terminate the game based on settable score and number of moves that the bot has been in a bad score.

careless25:
@fsmosca How does a bot know when to resign? What if opponent makes a mistake in the future if the position was played out?

ddugovic:
Perhaps resignation (and draw?) could be configured similar to [cutechess adjudication](https://github.com/cutechess/cutechess/blob/20dca5b19e8450ba20fb21e5b1a42d6f2942ba39/projects/lib/src/gameadjudicator.cpp) ?

fsmosca:
@careless25 
If my bot score is already 1 queen down or worst, I can resign the game but let the api knew it.  Note I programmed the bot I run in lichess.

Now if I know the rating of opponent and I know that his time is still say 2 minutes or more (thru wtime btime for example), I can resign the game too, this is more reliable resignation.

As a bot author I don't care about later moves that my opponent may blunder, I try to be a good sport to human opponent or even bot opponent.

fsmosca:
True UCI does not explicitly specify to support resign, but an engine can in a way resign itself by sending
`bestmove 0000`

So api/gui would mostly likely penalize the engine with illegal move. 

But what I proposed is for the engine to send additional info string,
`info string resign`
To make it clear that the uci engine would like to resign.

xboard engine can resign of course, search resign word.
https://www.gnu.org/software/xboard/engine-intf.html

ddugovic:
Hundreds of [UCI engines](https://chessprogramming.wikispaces.com/UCI+Engines) currently do not resign; hence, cutechess adjudicates when engines evaluate below a configurable threshold for a configurable number of consecutive moves.  (Likewise for draws, instead of waiting for the 50-move rule to take effect.)

Stockfish uses `bestmove (none)` when there are no legal moves or `bestmove e2e4 ponder (none)` when the opponent has no legal response (in this case, to `e2e4`).

fsmosca:
@ddugovic 
A UCI engine does not resign by itself because resigning is not specified in the protocol.

[...]

Issue #21, opened 2018-04-19
Bot stalls when opponent offers draw


ana-ka:
See: 

https://lichess.org/A3aYRGi7
https://lichess.org/5PhGQKAV
https://lichess.org/yQ8yLEmU/black
...

careless25:
The Lichess Bot API doesn't support draw offers. How would an engine know when to offer a draw? Or when to accept a draw?

ana-ka:
I was watching these games and my point was that the users bot repeatedly lost on time and didn't make a move when the opponent offered a draw and played on. That's clearly not intended. I suppose the engine should either just play on or accept a draw offer. I'd prefer the first one..

ddugovic:
The UCI protocol doesn't define a way to notify UCI-compliant engines about a draw offer.

ana-ka:
Alright. But still the bot should not lose on time in this case and just play on as if nothing happened, no?

drbeco:
I think this issue may be related to #128 

Xboard protocol allows "resign" and "offer draw"

PR #103, opened 2018-05-02
Next

careless25:
Add resign endpoint and eval command.


drbeco:
I have being tracking a bug reported at #128 and I'm reading about "resign" and "offer draw" support.

* To link issues related, #79 is already linked, and number #21 is also on the matter.

Thanks

PR #104, opened 2018-05-05
Uci ponder

smartchessguiapp:
Think on opponent time. Only works with protocol uci.

Tested by third party https://lichess.org/@/main_line :

https://lichess.org/forum/team-lichess-bots/pondering-bot--help-me-with-tests

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -16,6 +16,8 @@ from conversation import Conversation, ChatLine
 from functools import partial
 from requests.exceptions import ChunkedEncodingError, ConnectionError, HTTPError
 from urllib3.exceptions import ProtocolError
+import threading
+import time
 
 try:
     from http.client import RemoteDisconnected
@@ -98,6 +100,8 @@ def start(li, user_profile, engine_factory, config):
     control_stream.terminate()
     control_stream.join()
 
+ponder_results = {}
+
 @backoff.on_exception(backoff.expo, BaseException, max_time=600)
 def play_game(li, game_id, control_queue, engine_factory, user_profile, config):
     updates = li.get_game_stream(game_id).iter_lines()
@@ -111,6 +115,8 @@ def play_game(li, game_id, control_queue, engine_factory, user_profile, config):
     print("+++ {}".format(game))
 
     engine_cfg = config["engine"]
+    is_uci = engine_cfg["protocol"] == "uci"
+    is_uci_ponder = is_uci and engine_cfg.get("uci_ponder", False)
     polyglot_cfg = engine_cfg.get("polyglot", {})
 
     if not polyglot_cfg.get("enabled") or not play_first_book_move(game, engine, board, li, polyglot_cfg):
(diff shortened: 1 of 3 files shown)


gbtami:
In theory python-chess supports pondering for xboard engines from 0.23.4 also.

smartchessguiapp:
This solution uses an elementary approach: make the ponder move on the board and start a normal search with this board. This is not dependent on engine option settings and works for experimental engines, that don't support the ponder option for the go command.

It could be easily made to work with the xboard protocol if it were possible to extract a ponder move from the xboard search. I'm neither familiar enough with xboard nor with python-chess to know how to do this safely.

gbtami:
Why reinventing the wheel if python-chess supports pondering?
See https://github.com/niklasf/python-chess/blob/master/chess/uci.py#L899

Maybe I'm wrong but implementing pondering for engines which doesn't support it shouldn't be the lichess-bot task.

ddugovic:
Stockfish's [parameter list](http://support.stockfishchess.org/kb/advanced-topics/engine-parameters) specifies that `Ponder` has a default value of true, although actually running Stockfish the `uci` command reveals:

```
option name Ponder type check default false
```

smartchessguiapp:
With my solution you have to be aware of nothing.

Being aware of the uci command amounts to being aware of something.

Nothing is less than something.

So my solution is better :)

ddugovic:
Ha, that reminds me of how Windows once had dozens of ways of doing the same thing, and had so many service packs.

careless25:
So can anyone test this change?

ddugovic:
I wonder long-term (as other changes are made) what is the process for testing both ways of pondering, and what happens if you accidentally combine them.

gbtami:
Again. Uci has pondering. python-chess has pondering. Why reinvent pondering in lichess-bot ?

smartchessguiapp:
In the OP I said it has been tested and have given a link both to the party who tested it and the respective thread discussion in the bot team forum.

In terms of implemtation I can only speak for uci: the way pondering is defined in the specification, makes no sense.

If the opponent makes the expected move, then fine, you can send the `ponderhit` command and the engine will come up with a move. However this is the same as in my solution, because I already started the search with this move on the board.

However if the opponent does not make the expected move, there is no `pondermiss` command, to switch the ongoing search to the actual move. You have to shut down the search and start a new one ( just as in my solution ).

The whole official uci ponder machinery does not offer a single advantage compared to my method, but introduces a lot of risks. So it is strictly worse. Python-chess's support for uci pondering that I don't use is only in that it sends the ponder option in the go command. All other pondering support is for returning a ponder move ( which I do fully use ) from the search.

I cannot speak for xboard.

smartchessguiapp:
In the OP I said it has been tested and have given a link both to the party who tested it and the respective thread discussion in the bot team forum.

In terms of implemtation I can only speak for uci: the way pondering is defined in the specification, makes no sense.

If the opponent makes the expected move, then fine, you can send the ponderhit command and the engine will come up with a move. However this is the same as in my solution, because I already started the search with this move on the board.

However if the opponent does not make the expected move, there is no pondermiss command, to switch the ongoing search to the actual move. You have to shut down the search and start a new one ( just as in my solution ).

The whole official uci ponder machinery does not offer a single advantage compared to my method, but introduces a lot of risks. So it is strictly worse. Python-chess's support for uci pondering that I don't use is only in that it sends the ponder option in the go command. All other pondering support is for returning a ponder move from the search ( which I do fully use ).

I cannot speak for xboard.

gbtami:
"the way pondering is defined in the specification, makes no sense"
Maybe this explanation from Richard Vida can help: http://www.open-chess.org/viewtopic.php?f=5&t=2146

ddugovic:
> However if the opponent does not make the expected move, there is no pondermiss command, to switch the ongoing search to the actual move. You have to shut down the search and start a new one ( just as in my solution ).

Correct, however `stop` (or start) of a search does not empty the transposition table.  Of course there's the possibility of the ponder move being bad, so sometimes `go infinite` might make more sense than `go ponder`; but in either case `stop` followed by starting a new search (with a full TT) may be useful.

I'm confused about the apparent need for a separate `ponder_thread` different from the engine subprocess and later this week can delve further into this...

smartchessguiapp:
Wait a moment. We are in an event processing queue. We take an event from the queue and process it. If I start blocking this queue by starting a blocking search in the main thread, than I'm never notified about the opponent making a move.

The thread is started in the subprocess. Now it is a multi threaded subprocess. The only thing with that is that in the `finally` clause I shut the the thread gracefully, so the process is cleaned up and only the main thread remains and can exit as it wishes.

gbtami:
"The whole official uci ponder machinery does not offer a single advantage compared to my method, but introduces a lot of risks. So it is strictly worse."

Worse? @smartchessguiapp as I know the uci/xboard ponder machinery works well in several professional GUI programs. What kind of risks are you talking about here?

ddugovic:
OK, thanks for explaining usage of `ponder_thread`.

If I understand this PR correctly, it supports "search on my opponent's turn" even if the engine failed to provide a ponder move.  It does the following:

* If `lichess-bot` has no ponder move, join to the engine thread to extract the ponder move.  But surely the previous `first_search` or `search` could have just as easily returned the ponder move?
* Now if there is a ponder move, if there is time remaining, do something other than `ponderhit`, perhaps because `python-chess` doesn't support it?

gbtami:
I think this patch is inferior to uci protocol regarding it doesn't enable engine to decide it's own pondering and time management strategy.

smartchessguiapp:
As to the risk of using the official uci mechanism I can give a sort of topological answer:

I have shown that it has no advantage compared to my method.

Once it has no advantage, the only thing that can happen other than being equal to it, is that it is worse.

May be there is an extra setting that the engine needs ( what on earth does "Ponder" uci option in Stockfish? could it override the protocol? do I have to set it to finally be able to make work it as it is described in the protocol? strange! - the user may easily mess this setting up or simply not know about it ), may be the engine that works perfectly with normal search has some weirdness with ponder search. There is a nonzero possibility for all the above. Why take this risk?

smartchessguiapp:
What better strategy is there other than thinking about the move that was actually made?

As to time management: the engine was told about the thinking times of the players ( a make a correction for that if you look at the code, I subtract the time used to come up with the ponder move from the thinking time of the engine ). If the move played was the ponder move, than it just started the same think as it would otherwise earlier.

lichess-bot.py
```diff
@@ -111,29 +115,71 @@ def play_game(li, game_id, control_queue, engine_factory, user_profile, config):
...
+                    if is_uci_ponder and not ( ponder_move is None ):
+                        mwtime = wtime
+                        mbtime = btime
+                        elapsed_thinking_time_ms = int ( ( time.time() - thinking_started_at ) * 1000 )
+                        if game.is_white:
+                            mwtime = wtime - elapsed_thinking_time_ms
+                        else:
+                            mbtime = btime - elapsed_thinking_time_ms
+                        if ( mwtime > 0 ) and ( mbtime > 0 ):
+                            ponder_board = board.copy()
```
ddugovic: What's wrong with [ponderhit](https://python-chess.readthedocs.io/en/latest/uci.html#chess.uci.Engine.ponderhit) ?

gbtami:
"I have shown that it has no advantage compared to my method."
No you have not. I gave you a link where Vida explains how pondering works in uci. You think your out of protocol implementation is better than engine authors own pondering implementation (using the protocol).
Sorry, but I give up here.

ddugovic:
OK, upon re-re-reading the code maybe I correctly understand what it's doing?

* If `uci_ponder` is enabled, `search_with_ponder`, else just `search` (or `first_search`) discarding the ponder value
* If (upon receiving the opponent's move) there's a ponder hit, instead of using the python-chess `ponderhit` instead do `ponder_thread.join()` and wait for `ponder_thread_func` to return a move (which seems to issue another `go` instead of `ponderhit`)

smartchessguiapp:
I merged changes since my pull request. Now there are no conflicts any more.

Also made pondering canonical. Now it fully implements the uci specification.

In default.config I warn the user to send appropriate UCI option, should this be necessary for their engine ( the example I gave is what works for Stockfish ).

I tested it for myself in actual game, but I don't have third party test for it.

PR #105, opened 2018-05-06
Added go commands

careless25:


engine_wrapper.py
```diff
@@ -77,8 +77,9 @@ class UCIEngine(EngineWrapper):
 
     def __init__(self, board, commands, options, silence_stderr=False):
         commands = commands[0] if len(commands) == 1 else commands
-        self.engine = chess.uci.popen_engine(commands, stderr = subprocess.DEVNULL if silence_stderr else None)
+        self.go_commands = options.get("go_commands") if options.get("go_commands") else {}
```
ornicar: or `self.go_commands = options.get("go_commands", {})`

config.yml.default
```diff
@@ -15,6 +15,13 @@ engine:                      # engine settings
     Move Overhead: 100       # increase if your bot flags games too often
     Threads: 2               # max CPU threads the engine can use
     Hash: 256                # max memory (in megabytes) the engine can allocate
+#   go_commands:
+#     searchmoves: "c2c4"    # (Optional) Restrict search to moves in this list.
```
ddugovic: I assume that in most positions `c2c4` is illegal.  I'm trying to imagine in what context this configuration setting would be useful.

PR #108, opened 2018-05-09
Added engines folder to gitignore

krs013:
Sorry I know this is trivial but it's a little QoL change I made and I don't want to sneak it in with a normal PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 config.yml
 __pycache__
+/engines/*
+!/engines/README.md


typescriptstudies:
`README.md` in `engines` folder is there so that the folder will be created for the user. If you put the whole folder it into gitignore, then it will ignore `README.md` as well and folder won't be created.

You should add:

`!engines/README.md`

typescriptstudies:
`README.md` in `engines` folder is there so that the folder will be created for the user. If you put the whole folder into `.gitignore`, then it will ignore `README.md` as well and the folder won't be created.

You should add:

`!engines/README.md`

krs013:
Oh great point, @typescriptstudies -- I hadn't noticed that. I'll do that real quick. Thanks!

ornicar:
Add the Mac/Emacs crap to your global gitignore. https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

Environment specific files should not pollute projects gitignore files.

krs013:
Thanks--really glad to know about the global config. I reset the commits and changed it to just have the engines folder, if you still want that.

PR #113, opened 2018-05-11
Graceful exit

smartchessguiapp:
Allows the user to gracefully exit the application by typing "x" or "X" then pressing ENTER.

Adresses complaints about stopping the application using Ctrl+C being cumbersome.

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -17,6 +17,7 @@ from conversation import Conversation, ChatLine
 from functools import partial
 from requests.exceptions import ChunkedEncodingError, ConnectionError, HTTPError
 from urllib3.exceptions import ProtocolError
+import threading
 
 try:
     from http.client import RemoteDisconnected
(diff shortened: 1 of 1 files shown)


ornicar:
Can we have that by pressing Ctrl+c instead?

gbtami:
I suggest to use this pattern instead:
```
import signal

def signal_handler(signal, frame):
        print('You pressed Ctrl+C!')
signal.signal(signal.SIGINT, signal_handler)
```

typescriptstudies:
This generates a long error log. My solution prints "exiting lichess-bot...", then exits smoothly without error messages.

smartchessguiapp:
This generates a long error log. Not graceful enough.

careless25:
Can you implement it as `CTRL+C` like @ornicar mentioned?

HalmonLui:
since ctrl-c will always raise a KeyboardInterrupt, instead of using signal you could use a try/except and listen for a KeyboardInterrupt:
```python
def user_input(prompt): 
  try: 
    return input(prompt)
  except KeyboardInterrupt:
    print("If you want to exit, press ctrl-c again to quit")
  return input(prompt) 
```

krs013:
@HalmonLui I think the trouble with that is that the program might not be in that thread when the signal is sent.

smartchessguiapp:
https://github.com/careless25/lichess-bot/commit/359ad213dbcfac9ba2b84834dc0f096f08defd69

PR #121, opened 2018-06-04
Fix #75 Opening Books for Different Variants

### LINKED ISSUE #75: Opening Books for Different Varients
Currently, if you use an opening book, it will play the moves for all variants. However, this is undesirable as different variants have different moves played in the opening for example atomic most commonly games begin with 1. Nf3 f6 but in standard if 1. Nf3 is played the most common continuation is 1. ... Nf6 .  Thanks, TheM492017.

gbtami:


careless25:
@gbtami Can you make the fixes I suggested above?

gbtami:
Sure! But where is this "above"? :)

careless25:
Oops I forgot to "Submit Review". My bad

gbtami:
Done.

careless25:
Oops I think there was some miscommunication. I do want the book section but not the suffix. It should look like: 

```
book:
    standard: ...
    kingOfTheHill: ...
...
```

gbtami:
Done.

Issue #125, opened 2018-06-08
gate timeout error messages


OhJayGee:
I am seeing a number of errors from lichess-bot following the same pattern (see below).
All of them lead to a abort/game over or loss on time.

Does anybody else see that? is that actual connectivity, because the URLs seem correct, and the machines' internet connectivity was fine as well...

best regards,
OJG

`
+++ https://lichess.org/bG43Aj81/black Bullet vs BOT F_Me_Im_Famous(1500?)
ERROR:backoff:Backing off api_post(...) for 0.9s (requests.exceptions.HTTPError: 504 Server Error: Gateway Time-out for url: https://lichess.org/api/bot/game/bG43Aj81/move/d1e2)
ERROR:backoff:Giving up api_post(...) after 2 tries (requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://lichess.org/api/bot/game/bG43Aj81/move/d1e2)
Abandoning game due to connection error
Traceback (most recent call last):
  File "lichess-bot.py", line 150, in play_game
    li.make_move(game.id, best_move)
  File "/ChessEngines/lichess-bot/lichess.py", line 71, in make_move
    return self.api_post(ENDPOINTS["move"].format(game_id, move))
  File "/ChessEngines/lichess-bot/.venv/lib/python3.6/site-packages/backoff/_sync.py", line 99, in retry
    ret = target(*args, **kwargs)
  File "/ChessEngines/lichess-bot/lichess.py", line 61, in api_post
    response.raise_for_status()
  File "/ChessEngines/lichess-bot/.venv/lib/python3.6/site-packages/requests/models.py", line 935, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://lichess.org/api/bot/game/bG43Aj81/move/d1e2
--- https://lichess.org/bG43Aj81/black Game over
`

ddugovic:
I've seen similar messages, but only the case where the opponent hastily aborts the game before the first move.

careless25:
This happens because: 
1. Lichess sends game started notification
2. Lichess-bot starts the engine. While engine is starting: 
a. Opponent aborts
b. Lichess send game ended notification
3. Engine boots up and tries to play a move

If anyone has a good idea of how to fix this, make a PR or I will get to it once I have some time.

OhJayGee:
[...]

PR #129, opened 2018-06-27
Do not retry on HTTPError due to opponent abort #125

ddugovic:
This won't prevent the HTTP 400 response, but should reduce the number of retries.

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -27,6 +27,9 @@ except ImportError:
 
 __version__ = "1.1.0"
 
+def is_final(exception):
+    return isinstance(exception, HTTPError) and exception.response.status_code < 500
+
 def upgrade_account(li):
     if li.upgrade_to_bot_account() is None:
         return False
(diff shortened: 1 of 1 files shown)


drbeco:
I believe this is too radical. Internet in general needs a bit of redundancy. Ok that 12 tries is too much, but if I'm not reading this wrong, it would abort at the first error.

Maybe let it try something like 3 tries with an interval of a (few) second in between?

ddugovic:
Why is there a need for redundancy when the server returns an HTTP 400 response?  HTTP 400 means the client made an error.

drbeco:
You may be right, I'm being over defensive. Without reading further the code I can't say, but surely it is good to double check if this error is issued only in such controlled way.

ddugovic:
What are your thoughts on:
https://github.com/careless25/lichess-bot/blob/63568da558e46360990843e079ecbcf05276d568/lichess.py#L42

which is currently deployed on all `lichess-bot` instances and appears to be based upon the example provided by `backoff` itself?
https://github.com/litl/backoff/commit/cc8dc948ac7bbf5f2af3926545285ece8aba1491#diff-88b99bb28683bd5b7e3a204826ead112R49

ddugovic:
@drbeco Given that this PR only handles `HTTPError` and only for error codes below `500` and is based upon the above example, is there any objection to merging this (so @ornicar 's server doesn't get blasted with ~10 retries against `HTTP 400` responses every time an opponent aborts)?

ddugovic:
@drbeco @careless25 Please review as this should reduce 90% of the traffic to the Lichess server in cases where the opponent has aborted or resigned out of turn.

PR #132, opened 2018-07-01
add increment setting

quillford:


careless25:
How does someone specify any/infinite increment?

quillford:
180 seconds is the max on lichess so 180 or above will be unlimited. That should definitely be added to the comment and maybe it should be the default.

ddugovic:
How does someone specify a minimum increment?

careless25:
Can you make the changes @ddugovic and I suggested. Thanks :)

quillford:
Alright. I added a min increment setting and made the default more sensible.

PR #143, opened 2018-11-01
UCIHttpEngine

jeremylorino:
added implementation and `config.yaml` entries to request next move from a http endpoint

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ config.yml
 __pycache__
 /engines/*
 !/engines/README.md
-# lichess-bot.py
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
+.venv
 config.yml
 __pycache__
 /engines/*
 !/engines/README.md
+# lichess-bot.py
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -1,4 +1,6 @@
+.venv
 config.yml
 __pycache__
 /engines/*
 !/engines/README.md
+# lichess-bot.py
```
careless25: Remove this

config.yml.default
```diff
@@ -44,6 +44,9 @@ engine:                      # engine settings
 #   log: "./engines/log.txt" # save log of lczero debug output
   silence_stderr: false      # some engines (yes you, leela) are very noisy
 
+http_engine:
+  url: "http://play.lczero.org/getMoveHardcore"
```
ddugovic: There's no `#` comment?

jeremylorino: `http_engine` section is only used if `protocol=http` and I am the site owner ;)

.gitignore
jeremylorino: @careless25, removed

careless25:
So...I don't want to merge this into master because there is no standard HTTP protocol for chess engines. We could merge this onto another branch that you can maintain

jeremylorino:
I hate to accept that response. Regardless of whether you merge or not I will still standardize my rest api interface. Hopefully you are willing to support it as noone out there has decided to.

If we push this to another branch I may as well just keep it on my fork ☹️

gbtami:
@jeremylorino if you want to provide another leela BOT to lichess users you can use UCI. Evangelizing a new chess engine protocol in a github PR form seems a strange idea to me.

PR #152, opened 2018-11-30
Possible fix for various ChunkedEncodingError related issues

gbtami:
Seems requests lib has an interesting behavior regarding incomplete HTTP reads. See [https://blog.petrzemek.net/2018/04/22/on-incomplete-http-reads-and-the-requests-library-in-python/](https://blog.petrzemek.net/2018/04/22/on-incomplete-http-reads-and-the-requests-library-in-python/)
This leads to ChunkedEncodingError sometimes while reading lichess-bot control stream. This patch tries to fix this by calling the worker function itself if this exception occurs.

Feedback and assistance testing this would be much appreciated.

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -39,12 +39,16 @@ def upgrade_account(li):
 
 @backoff.on_exception(backoff.expo, BaseException, max_time=600, giveup=is_final)
 def watch_control_stream(control_queue, li):
-    for evnt in li.get_event_stream().iter_lines():
-        if evnt:
-            event = json.loads(evnt.decode('utf-8'))
-            control_queue.put_nowait(event)
-        else:
-            control_queue.put_nowait({"type": "ping"})
+    try:
+        for evnt in li.get_event_stream().iter_lines():
+            if evnt:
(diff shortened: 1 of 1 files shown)


lichess-bot.py
```diff
@@ -39,12 +39,16 @@ def upgrade_account(li):
...
-            control_queue.put_nowait({"type": "ping"})
+    try:
+        for evnt in li.get_event_stream().iter_lines():
+            if evnt:
+                event = json.loads(evnt.decode('utf-8'))
+                control_queue.put_nowait(event)
+            else:
+                control_queue.put_nowait({"type": "ping"})
+    except ChunkedEncodingError as exception:
+        logger.error("Restarting to whatch control stream due to incomplete read error.")
```
ddugovic: `error` seems a bit severe considering this client didn't cause the problem.

I think "incomplete read" isn't necessarily an accurate description of `ChunkedEncodingError`?  Also there's a spelling typo.

gbtami: Thx for the feedback, Dan! I think I have to think more about this.

PR #154, opened 2018-12-03
Changed "lichess-bot" to "lichessbot" and added option to enable color logging

QueensGambit:
Hello lichess.org developers,

great thanks for the official bot API on [lichess.org](lichess.org) which is a unique feature. 

I found out that the lichess client couldn't answer the command `!name` due to the symbol `-`,
so I changed the printout from `lichess-bot` to `lichessbot` in `conversation.py`.
 
Also I added a class which gives the developer the option to have coloured logging messages and works cross-plattform.
The class implementation is based on this stackoverflow post:
* https://stackoverflow.com/questions/384076/how-can-i-color-python-logging-output
If this does violate your LICENSE please reject this class.

Lastly I added a `\n` to the lichess-ASCII pony for nicer formatting in my opinion.
 
Best wishes,
~QueensGambit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,7 @@ config.yml
 __pycache__
 /engines/*
 !/engines/README.md
+
+# avoid pushing project files when people use PyCharm
+.idea/
+
--- a/ColorLogger.py
+++ b/ColorLogger.py
@@ -1,0 +1,150 @@
+"""
+@file: Colorer.py
+
+Script which allows are colored logging output multiplattform.
+The script is based on this post and was slightly adjusted:
+# https://stackoverflow.com/questions/384076/how-can-i-color-python-logging-output
+
+'Here is a solution that should work on any platform. If it doesn't just tell me and I will update it.
+
+How it works: on platform supporting ANSI escapes is using them (non-Windows) and on Windows
+it does use API calls to change the console colors.
+
+The script does hack the logging.StreamHandler.emit method from standard library adding a wrapper to it.'
+
+by Sorin & Dave
+"""
+
+import logging
+import sys
+import platform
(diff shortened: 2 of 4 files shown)


careless25:
I am confused about the `!name` command. How does the `-` interfere?

ddugovic:
I assume there's no problem in `lichess-bot` code itself, but I've seen the `!name` command be executed without Lichess displaying the output.  Probably Lichess has a "bad words" blacklist which doesn't catch `lichessbot`?

QueensGambit:
Daniel Dugovic seems to assume rightly that the word `lichess-bot` is on a blacklist.
I tried out changing `lichess-bot` to `lichess--bot` and it worked, but whenever `lichess-bot` appears in the message, [lichess.org](lichess.org) doesn't show it.

ddugovic:
If we had some way of verifying that, I could ask the lichess developers about their blacklist...

ornicar:
Don't change anything. I'll check the lichess blacklist and fix it.

.gitignore
```diff
@@ -2,3 +2,7 @@ config.yml
 __pycache__
 /engines/*
 !/engines/README.md
+
+# avoid pushing project files when people use PyCharm
+.idea/
```
ornicar: No. Your environment specific excludes must go to your own global gitignore file.

ornicar:
Please make atomic PRs that do only one thing each.

QueensGambit:
Thanks [Thibault](https://github.com/ornicar) for removing the word `lichess-bot` from the blacklist.
The main reason why I created a PR was because the `!name` command was broken.
I'll create two pull-requests.

PR #161, opened 2018-12-14
Various bugfixes and improvements

noobpwnftw:
I recently saw this project and decided to run a bot myself.
During the experiments I found many bugs so I attempted to fix them, however I am not proficient in Python, nor I am very good at writing good-looking code.

There are a few things in this patch that I'd rather not to separate them again.
1. I borrowed the pondering code from https://github.com/smartchessguiapp/lichess-bot.
2. Handled various ReadTimeout exceptions and make them not mess up with the counters in case of failure.
3. A brief hook up to my online book API where you can safely ignore.

Tested via a bot account under my nickname, and it seems to not timing out often with a 300 ping latency to the lichess API when playing at bullet.

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -17,9 +17,10 @@ import backoff
 from config import load_config
 from conversation import Conversation, ChatLine
 from functools import partial
-from requests.exceptions import ChunkedEncodingError, ConnectionError, HTTPError
+from requests.exceptions import ChunkedEncodingError, ConnectionError, HTTPError, ReadTimeout
 from urllib3.exceptions import ProtocolError
 from ColorLogger import enable_color_logging
+import threading
 
 try:
     from http.client import RemoteDisconnected
@@ -58,7 +59,7 @@ def watch_control_stream(control_queue, li):
                 control_queue.put_nowait(event)
             else:
                 control_queue.put_nowait({"type": "ping"})
-    except (RemoteDisconnected, ChunkedEncodingError, ConnectionError, ProtocolError) as exception:
+    except (RemoteDisconnected, ChunkedEncodingError, ConnectionError, ProtocolError, ReadTimeout) as exception:
         logger.error("Terminating client due to connection error")
         traceback.print_exception(type(exception), exception, exception.__traceback__)
         sys.exit()
@@ -101,10 +102,10 @@ def start(li, user_profile, engine_factory, config):
                     logger.debug("Something went wrong. Game is starting and we don't have a queued process")
                 else:
                     queued_processes -= 1
-                game_id = event["game"]["id"]
-                pool.apply_async(play_game, [li, game_id, control_queue, engine_factory, user_profile, config])
                 busy_processes += 1
                 logger.info("--- Process Used. Total Queued: {}. Total Used: {}".format(queued_processes, busy_processes))
+                game_id = event["game"]["id"]
+                pool.apply_async(play_game, [li, game_id, control_queue, engine_factory, user_profile, config])
             while ((queued_processes + busy_processes) < max_games and challenge_queue): # keep processing the queue until empty or max_games is reached
                 chlng = challenge_queue.pop(0)
                 try:
@@ -121,6 +122,8 @@ def start(li, user_profile, engine_factory, config):
     control_stream.terminate()
     control_stream.join()
 
+ponder_results = {}
+
 @backoff.on_exception(backoff.expo, BaseException, max_time=600, giveup=is_final)
 def play_game(li, game_id, control_queue, engine_factory, user_profile, config):
     response = li.get_game_stream(game_id)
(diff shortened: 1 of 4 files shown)


ddugovic:
If this is stable don't worry about conflicts with #158 , I can always reintroduce that later...

lichess-bot.py
```diff
@@ -214,28 +279,18 @@ def play_first_book_move(game, engine, board, li, config):
...
-            if selection == "weighted_random":
-                move = reader.weighted_choice(board).move()
-            elif selection == "uniform_random":
-                move = reader.choice(board, config.get("min_weight", 1)).move()
-            elif selection == "best_move":
-                move = reader.find(board, config.get("min_weight", 1)).move()
-        except IndexError:
-            # python-chess raises "IndexError" if no entries found
             move = None
+            result = li.api_query("http://api.chessdb.cn:81/cdb.php?action=query&board=" + board.fen())
```
careless25: Can we have this activated via a flag in the config?

noobpwnftw: I have removed the online book and restored polyglot book code.

gbtami:
I just note that another UCI ponder PR was rejected before .
See https://github.com/careless25/lichess-bot/pull/104

noobpwnftw:
I didn't know that happened, it seems to me that where I borrowed the pondering code(by 
https://github.com/careless25/lichess-bot/pull/104) has correctly implemented pondering according to UCI specifications, while `engine.ponder` implemented here https://github.com/careless25/lichess-bot/blob/Leela/engine_wrapper.py is incomplete due to missing "ponderhit" mechanism.

gbtami:
Read all comments in #104 also, please!

noobpwnftw:
How would one interpret this pondering mechanism otherwise, I do not understand.

Here is a brief example:
```
position fen [FEN]
go [time, depth, etc]
  bestmove [move] ponder [ponder]
position fen [FEN + move + ponder]
go ponder [time, depth, etc]
```
In case of ponder hit:
```
ponderhit
  bestmove [move] ponder [ponder]
```
In case of ponder not hit:
```
stop
  bestmove [move] ponder [ponder] <- ignored
position fen [FEN + move]
go [time, depth, etc]
  bestmove [move] ponder [ponder]
```
The UCI option named "Ponder" is an indicator to the engine that the user will use the above pondering mechanism during play, so that the engine can adjust accordingly especially time management, it does not change any behavior of the "go" command(strictly speaking, if "Ponder" is off, engine should only output best move without ponder move because it was told that the pondering mechanism is not used, but unfortunately in real world most users often fail to comprehend), and none of this ponder hit/not hit handling logic exists in python-chess, and it's also expected that the user may decide what move to ponder on, while the engine only suggests what move it thinks is worth pondering.

OhJayGee:
The pondering code is fine and is fulfilling the uci spec fully - including handling ponderhit.
I have been running smartchessgui's code for 6+ month without issue.

I think the last time this PR was introduced smartchessgui's attitude towards uci standards rather ruffled too many feathers - so in the end it was overlooked that he ended up modifying the implementation to fully comply with the the uci standard.  (in 
https://github.com/careless25/lichess-bot/pull/104/commits/c58558836f1c28334b0228bffed0a2c24dfa5f87 late in the thread)

The use of multithreading is required IMHO as ponder is basically running a background process, while continuing to listen to the incoming queue, waiting for the actual opponents move - which then either creates ponderhit or new search. 

Whether python-chess supports pondering or not (it does) is not the question here - there would still be required rigging/threading in the code to utilize python-chess's implementation of pondering.

just my 0.02

gbtami:
@noobpwnftw thx for the explanation! 
@ddugovic yep, seems it will crash on line https://github.com/careless25/lichess-bot/blob/ec34902d967acc7c36be89e46319bbab12db59cd/lichess-bot.py#L206 for CECP engines.
Anyhow, polluting lichess-bot.py with uci specific code seems problematic when we have engine_wrapper.py

noobpwnftw:
`search_with_ponder` defaults to ponder=False, unless otherwise specified.

gbtami:
@noobpwnftw yes, but its not implemented in XBoardEngine class.

noobpwnftw:
@gbtami I have resolved conflicts and added wrapper function for XBoardEngine, hope it works.

gbtami:
@noobpwnftw I'm not the repo owner here, but I suggest some things to easy the acceptance of your work.
1. Create different PRs for commits having different goals (bugfix, new feature, other).
2. If you change anything not closely related to the paritcular bugfix/new feature, add some comment about why you want to change it.
3. Avoid using
```python
     except:
        pass 
```
This PR wants to change too many things at once IMO.

noobpwnftw:
It is quite some work to separate these things.

Reason for that to be a while loop is because `next(lines)` is the one actually triggering exceptions in case of a network problem, and it wasn't guarded by the try catch block, so backoff kicks in and the whole engine is restarted, causing timeouts in bullet games.

gbtami:
Create a separate PR for that then, please!

PR #181, opened 2019-07-13
Send Xboard engines last move instead of board FEN

### LINKED ISSUE #178: Xboard engine repeats moves
Previous moves are required to detect 3rd repeated position. In Xboard protocol only last position is send to the engine. And no "usermove" command is used. 

UCI correctly sends previous moves. Source code is in ucy,py, class Engine, method "def position(self, board, async_callback=None):".

Similar code should be placed in engine_wrapper.py, class XBoardEngine, method "def search(self, board, wtime, btime, winc, binc):".

Example: https://lichess.org/ETxoz4EQ#78 White (Phalanx) plays Re1 and continue 3rd repetition sequence, but normally it plays Re8+ and wins.

MarkZH:
This will allow Xboard engines to track the game history for move
repetitions and other purposes.

Fixes #178

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -170,6 +170,7 @@ class XBoardEngine(EngineWrapper):
         return bestmove
 
     def search(self, board, wtime, btime, winc, binc):
+        self.engine.force()
         try:
             self.engine.usermove(board.peek())
         except IndexError:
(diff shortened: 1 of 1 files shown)


MarkZH:
The latest commit fixes the problem where the "go" command is issued when the Xboard engine is already thinking due to the previous "usermove" command (assuming the engine is not in force mode). It is probably an error for a "go" command to be sent to a thinking engine.

PR #187, opened 2020-01-01
Enable strategies written directly in Python as an alt to engines.

namin:
I thought it convenient, for prototyping, to be able to transform any strategy (represented as a Python function from board to move) into a Lichess bot.

WDYT?

Thanks.


ShailChoksi:
I like the idea but I am not convinced this will be used extensively. I will leave this PR open to see how many users are interested in something like this.

ShailChoksi:
Closing as #304 was merged

PR #235, opened 2020-11-15
Allow multiple books

hyperbotauthor:
Users incrementally updating their book with their own games wish that they don't have to merge changes to a main book, rather compile a small book of their own games add this as a complementary book to the main book.


hyperbotauthor:
Forgot about commit which added books.

PR #236, opened 2020-11-15
Allow multiple books

hyperbotauthor:
Users incrementally updating their book with their own games wish that they don't have to merge changes to a main book, rather compile a small book of their own games add this as a complementary book to the main book.

### DIFF
--- a/config.yml.default
+++ b/config.yml.default
@@ -9,14 +9,14 @@ engine:                      # engine settings
   polyglot:
     enabled: false           # activate polyglot book
     book:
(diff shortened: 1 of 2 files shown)


zzhangb4:
This multiple book setting works very well! Currently I'm using three books with lichess-bot, and it's very easy to reach up to 30 moves and still be in-book :P I haven't had any issues with it so far, and I expect it to work perfectly for anyone else. 

Using many opening books saves more time for the middlegame and endgame. Very useful for fast time controls.

ShailChoksi (approved):

Issue #241, opened 2021-01-04
lichess-bot occasionally crashes


sander-NULL:
From time to time lichess-bot crashes resulting in the bot going offline. The python script cannot be terminated with CTRL + C anymore and I have to kill the process manually. It seems that this happens only after games have been played and not in the middle of them. I start the bot with
`virtualenv .venv -p python3`
`source .venv/bin/activate`
`python lichess-bot.py -v --logfile=bot-log.txt`
and I'll attach the last five logs.
[bot-log.txt](https://github.com/ShailChoksi/lichess-bot/files/5766530/bot-log.txt)
[bot-log2.txt](https://github.com/ShailChoksi/lichess-bot/files/5766548/bot-log2.txt)
[bot-log3.txt](https://github.com/ShailChoksi/lichess-bot/files/5766552/bot-log3.txt)
[bot-log4.txt](https://github.com/ShailChoksi/lichess-bot/files/5766554/bot-log4.txt)
[bot-log5.txt](https://github.com/ShailChoksi/lichess-bot/files/5766555/bot-log5.txt)
[bot-log6.txt](https://github.com/ShailChoksi/lichess-bot/files/5786989/bot-log6.txt)
[bot-log7.txt](https://github.com/ShailChoksi/lichess-bot/files/5829874/bot-log7.txt)

Edit: lichess-bot crashed again and I'll attach the log.
Edit 2: It now happens roughly once a day if not more often. I'll attach the (relevant) log of the very last time. There it seems to have happened after a declined bullet challenge. The logs don't show many details and I would provide more if only I knew how.

sander-NULL:
Thanks for your answer but I don't understand how this should fix the problem. I already have lichess-bot running and it does so quite well except for the above mentioned problem. Maybe I am misunderstanding you here but how should installing lichess-bot again remedy this?

chessBOTneeder:
This is actually not installing lichess-bot again. This is just putting your files online to repl.it and if you prefer, you might be able to let it run 24/7. You can run it online instead of on your computers cmd which can limit some code.  Just go to https://repl.it and create a new repl, then, choose a language, like "bash" and then after the repl is created, just upload your bot's files. Then run your own code for it. It probably won't crash. And like said before, you may be able to find a way to make it run 24/7 without even opening your online repl. Just Google some ways to make a repl.it repl run 24/7. **I don't know how to solve the computer cmd problem, sorry about this.**

chessBOTneeder:
[...]

PR #253, opened 2021-01-31
Call task_done() after processing an event on the control queue

branransom:
This will indicate that all items in the queue have been processed, so
that queue.join() no longer blocks the main thread. This is described here
in the Python docs: https://docs.python.org/3/library/queue.html

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -136,6 +136,8 @@ def start(li, user_profile, engine_factory, config):
                         logger.info("    Skip missing {}".format(chlng))
                     queued_processes -= 1
 
+            control_queue.task_done()
+
     logger.info("Terminated")
     control_stream.terminate()
     control_stream.join()


branransom:
This should fix this issue: https://github.com/ShailChoksi/lichess-bot/issues/241

I was having the same problem described in that issue when running locally on my computer and in an AWS EC2 instance. Often, my bot would crash every couple hours and I would have to restart it. Since adjusting my local copy of the lichess-bot code to include the call to task_done(), I have been able to run my bot through the lichess-bot bridge for 5 days straight.

As this is my first PR into this project, I'd also like to say thank you for sharing this with everybody. I have mentioned your project in the README of my [chess-ai](https://github.com/branransom/chess-ai) project, as it has made wiring up my chess bot to lichess very easy!

ShailChoksi (approved):

PR #258, opened 2021-02-08
Send go commands properly

hyperchessbot:
There were several problems with sending custom go commands to the engine.

- Go commands config was not even sent to engine init ( only uci options ).

- An attempt was only made to read those setting in normal search, but this is never called ( only search with ponder option is ever called ).

- In config go command options were misaligned, making for invalid yml.

All those problems are fixed in this PR by sending the go command options to engine init, by adding an option to search with ponder, which ignores thinking times if any of the possible go command options is set, and sends the custom go command instead, and by aligning `config.yml.default` properly.

### DIFF
--- a/config.yml.default
+++ b/config.yml.default
@@ -27,10 +27,10 @@ engine:                      # engine settings
     Ponder: true
     Threads: 2               # max CPU threads the engine can use
     Hash: 256                # max memory (in megabytes) the engine can allocate
-#   go_commands:             # additional options to pass to the UCI go command
-#     nodes: 1               # Search so many nodes only.
-#     depth: 5               # Search depth ply only.
-#     movetime: 1000         # Integer. Search exactly movetime milliseconds.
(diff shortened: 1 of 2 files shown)


MarkZH:
Are you sure about the config misalignment? The go_commands are only relevant to UCI bots, so they should be grouped under uci_options. The change to search_with_ponder() is correct, though.

MarkZH:
You're right. I'm working on upgrading lichess-bot to use the latest version of pychess and had to fix the same problem.  Here's what I did, which is a bit hacky in comparison:

```python
def __init__(self, board, commands, options, silence_stderr=False):
    self.go_commands = options.get("go_commands", {})
    if self.go_commands:
        del options["go_commands"]
    self.engine = chess.engine.SimpleEngine.popen_uci(commands, stderr=subprocess.DEVNULL if silence_stderr else None)
    self.engine.configure(options)
    # etc.
```

PR #259, opened 2021-02-09
Simplify updating board

MarkZH:
Since every game update contains every move of the game, rebuild the
board from scratch instead of carefully parsing the message for just the
latest move. The function update_board() is merged into setup_board()
and setup_board() is called with every gameState message.

Furthermore, since the board variable contains all of the information
about the current state of the game, use it instead of unparsed text
data. This simplifies is_engine_move() and any function needed the
number of moves in the game.


MarkZH:
I would be very surprised if setting up a new board from scratch even took a millisecond. Network latency is still orders of magnitude larger than that, so I don't think this change is consequential for performance. I could be wrong.

PR #264, opened 2021-02-28
Update config.yml.default

vigneshnandi-hub:
i updated to connect it with my bot


TheYoBots:
Why have you added your personal api access token (which is not supposed to be shared with anyone) and the engine you are using?

lichess-bot is supposed to support all engines running under either uci or xboard protocol. Please do close this PR and make sure you delete that api access token and make a new one and never share it.

vigneshnandi-hub:
Thanks for your help
But I dont use this token.

TheYoBots:
Then this PR will only work for the user who has that token. Also lichess-bot gives users the choice on which engine to use. Hence this PR will not be necessary or added.

ShailChoksi:
Declining due to the comments above. This is too specific to one configuration for a generalized tool.

Issue #263, opened 2021-02-26
No config.yml file


i4D4:
I cannot find my config.yml file, but every tutorial I've seen seems to have one from the start.

MarkZH:
Make a copy of the provided `config.yml.default` and rename it `config.yml`. Then, edit it to suit your bot.

vibranium21:
use "save as" in notepad plus plus

PR #265, opened 2021-02-28
Rename config.yml.default file to config.yml

TheYoBots:
Skip unnecessary step of copying the config.yml.default file to config.yml file to provide ease for users. 

Related to #263

I have edited the readme.MD file and renamed the config.yml.default file as I found the step of copying the config.yml.default file to config.yml file unnecessary.


MarkZH:
I think it's useful to have a config.yml.default file. If future code changes require modifications to the configuration file, the user's configuration file won't be modified when upgrading to the latest lichess-bot. I wouldn't want to inflict merge conflicts on users who just want to run a bot. If changes are required, we should strive to make them backwards compatible with older configuration file formats.

TheYoBots:
Thanks (@) MarkZH for your review. I hadn’t thought about that when committing these changes. I’ll close this PR in that case.

Thanks again!

PR #267, opened 2021-02-28
Fix broken link

TheYoBots:
The link under lichess api was directed to the old bot api url. I’ve changed it to use https://lichess.org/api#tag/Bot instead of https://lichess.org/api#tag/Chess-Bot as the latter redirects to https://lichess.org/api

Sorry I have made 3 different PR’s at the same time, but I have done so because I feel that all 3 are totally different and unrelated to each other.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # lichess-bot
-A bridge between [Lichess API](https://lichess.org/api#tag/Chess-Bot) and bots.
+A bridge between [Lichess API](https://lichess.org/api#tag/Bot) and bots.
 
 
 ## How to Install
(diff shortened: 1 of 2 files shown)


AttackingOrDefending:
This link is also on conversation.py. https://github.com/ShailChoksi/lichess-bot/blob/master/conversation.py#L25

TheYoBots:
Thanks (@) AttackingOrDefending, sorry I missed that out.

I’ve committed the change.

TheYoBots:
Also recently lichess stopped allowing bots to post links in players or spectators chat room. So should I just remove the ‘!howto’ command or break the link so that it can be sent or something else?

TheYoBots:
Lichess doesn’t allow bots to send links in players/spectators chat room. Hence if a player/spectator sends the ‘!howto’ command the bot gives no response. To make sure that this doesn’t occur I have replaced the link sent in the ‘!howto’ with ‘Check out Lichess Bot API’.

ShailChoksi (approved):

PR #260, opened 2021-02-13
Upgrade python chess library to latest version (1.4.0)

MarkZH:
This commit updates lichess-bot to use the latest version of the python
chess library. The main result is a large decrease in the amount of code
in engine_wrapper.py due to so many functions being incorporated into
the chess library.

One annoying difference is that the new library rejects the 'usermove'
feature from Xboard engines, which is rare among chess engine
interfaces.

I hope this can be tested by a variety of engines to make sure it is on
par with the current lichess-bot. This would close #240.

PR #270, opened 2021-03-04
Refactor lichess-bot.py

MarkZH:
This PR starts with the branch from which PR #260 derives and refactors the `play_game()` function to be shorter and easier to read. New functions are created that greatly simplify the code in the `elif u_type == "gameState":` branch of the main loop of the `play_game()` function. The entire `while` loop in that function now fits on one screen, and the logical structure of how moves are chosen (whether by book, pondering, or searching) is now clear.

Other resulting enhancements and corrections: (1) The engine can now ponder after the first move; (2) The time adjustments from PR #246 now include the fake thinking time. Other than these, there should be no functional differences.

The first new commit after #260 is 116a7e4d39ac777d6267ef9797b7460b17e0050a on February 27. If desired, I can create a pull request with all of these commits squashed into one (or just the commits needed after PR #260 is applied). I think it may also be possible to create an equivalent pull request that does not require #260.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # lichess-bot
-A bridge between [Lichess API](https://lichess.org/api#tag/Chess-Bot) and bots.
+A bridge between [Lichess API](https://lichess.org/api#tag/Bot) and bots.
 
 
 ## How to Install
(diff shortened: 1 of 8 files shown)


ShailChoksi:
@MarkZH Thanks a lot for this refactor. it was badly needed. Can you fix the merge conflicts and then we can merge this in

MarkZH:
@ShailChoksi I think that should do it.

PR #272, opened 2021-03-05
correcting README.md

OIVAS7572:
Please accept this pull request.
For so many people didn't work these commands.
I found the original commands.
Please watch and accept this.


TheYoBots:
Is thi only an issue in Mac/Linux or also in Windows? If it’s an issue in windows too, then please change that too :)

OIVAS7572:
I didn't find issue on Windows
If you found any issue solve that and propose

PR #274, opened 2021-03-07
Addition of git actions status badge

TheYoBots:
I’ve added the git actions status badge to this repo, which gives the status of the latest master build not only in ‘actions’ , but also clearly seen in the README.md docs. In case you would like to see how it looks after the change, do check [here](https://github.com/TheYoBots/lichess-bot/tree/theyobots-patch-3#lichess-bot).


TheYoBots:
Sorry, was a mistake.

TheYoBots:
I’ve changed the workflow name to ‘Python Build’ as that would suit better in the README.md docs than ‘Python application’. Also if you check the link I gave to see how it looks, it will show ‘no status’ as those changes are only a part of my repo for now. Once it is merged, that would change.

PR #277, opened 2021-03-08
Print move

AttackingOrDefending:
It makes the bot print the move that the game is now. This also makes the print_stats output clearer.

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -189,7 +189,7 @@ def play_game(li, game_id, control_queue, engine_factory, user_profile, config, 
                 if not is_game_over(game) and is_engine_move(game, board):
                     start_time = time.perf_counter_ns()
                     fake_thinking(config, board, game)
-                    print("move: {}".format(len(board.move_stack) // 2 + 1))
+                    print_move_number(board)
 
                     best_move, ponder_move = get_book_move(board, polyglot_cfg), None
                     if best_move is None:
@@ -335,6 +335,10 @@ def fake_thinking(config, board, game):
         accel = 1 - max(0, min(100, len(board.move_stack) - 20)) / 150
         sleep = min(5, delay * accel)
         time.sleep(sleep)
+
+
+def print_move_number(board):
+    logger.info("\nmove: {}".format(len(board.move_stack) // 2 + 1))
 
 
 def setup_board(game):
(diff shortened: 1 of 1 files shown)


MarkZH:
Doesn't this just print the move number? If you want the move as well, maybe something like this could work:
```python
if board.move_stack:
    temp_board = board.copy()
    temp_board.pop()
    print(temp_board.variation_san([board.move_stack[-1]])
```
https://python-chess.readthedocs.io/en/latest/core.html#chess.Board.variation_san

If you put this above the `if not is_game_over(game) and is_engine_move(game, board):` line, then both sides moves will be printed.

MarkZH:
I see. In that case, it would be better to put this code as the first line of the `Engine.print_stats()` method [here](https://github.com/ShailChoksi/lichess-bot/blob/936c516e5327d01eb7fb16efd4453513a294303a/engine_wrapper.py#L54). That way, if the program is run with `-v`, the numbering stays near the stats. Plus, you don't need separate cases for when it's black or white to move. `int(len(board.move_stack)/2) + 1` (or more compactly, `len(board.move_stack) // 2 + 1`) works for both cases.

AttackingOrDefending:
If I put the move as the first line of print_stats, It won't get printed when the bot got the move from the opening book. Also, print_stats is run after the engine thinks. I will use your version of printing a move. I can put it lichess-bot.py in line 192 or in the first line of search in engine_wrapper.py and a seperate case if it gets the move from the opening book
(`print(f"move: {len(board.move_stack) // 2 + 1}")`) in line 285 of lichess-bot.py. Which do you think is better?

MarkZH:
If this is for debugging, then run lichess-bot like this: `python lichess-bot.py -v`. The `-v` option  causes lichess-bot to print every command sent to the engine and everything your engine sends out while choosing a move and pondering. I think that would be a lot more useful than just printing the move number.

AttackingOrDefending:
I use this primarily so the print_stats are clearer but it also helps to understand if my engine has crashed when it is playing a game. It isn't useful for debugging.

MarkZH:
OK. My final suggestions:

1. Create a function called something like `def print_move_label(board)`, put the line in that function, and call the function in the loop. I'd like to keep each line of `play_game()` function as simple as possible.
2. Use `logger.info()` instead of `print()`
3. Add line breaks between moves, something like `logger.info(("\nmove: {}".format(len(board.move_stack) // 2 + 1))`

ShailChoksi (approved):

PR #281, opened 2021-03-11
Replace print() with logging

MarkZH:
Log all messages to the user with the logging module. Before, if the
user specifies a logfile via -l, the print() statements would be missed.

Also, log the usage of the choose_first_move() function similarly
to the choose_move() function.


ShailChoksi:
Can you resolve the merge conflict?

MarkZH:
Done.

PR #282, opened 2021-03-11
Notify engines of the final endgame state

MarkZH:
To Xboard engines: Send the final move (if any) and the result of the game.

To UCI engines: Send the final state of the board.

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -32,6 +32,21 @@ def remove_managed_options(config):
     return {name: value for (name, value) in config.items() if not is_managed(name)}
 
 
+class Termination:
+    MATE = 'mate'
+    TIMEOUT = 'outoftime'
+    RESIGN = 'resign'
+    ABORT = 'aborted'
+    DRAW = 'draw'
+
+
+class Game_Ending:
+    WHITE_WINS = '1-0'
+    BLACK_WINS = '0-1'
+    DRAW = '1/2-1/2'
+    INCOMPLETE = '*'
+
+
 class EngineWrapper:
     def __init__(self, commands, options, stderr):
         pass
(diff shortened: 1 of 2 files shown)


TheYoBots:
Just a question, Is it python chess that supports sending the final endgame state or have you added that support for lichess-bot, but not through python chess? And if it’s python chess that supports this could you send me the docs or the lines of the code in which it is mentioned. 

Thanks!

MarkZH:
A bit of both. The calls to `self.engine.protocol._position()` in the UCI engine wrapper and `self.engine.protocol._new()` in the Xboard wrapper are calls into the python-chess library. The first sends the final board position to UCI engines, and the second sends the final moves (if any) to Xboard engines. You can find the code for `_position()` [here](https://github.com/niklasf/python-chess/blob/c5739f13b64708e371e2be29a950b14096267a5b/chess/engine.py#L1456) and `_new()` [here](https://github.com/niklasf/python-chess/blob/c5739f13b64708e371e2be29a950b14096267a5b/chess/engine.py#L1959). Neither function handles reporting the winner or game-ending reason. I call these methods because the library figures out which moves need to be sent.

The UCI protocol doesn't support sending endgame information at all. Sending the last position line is something I find useful so my engine can create a complete PGN record after a game finishes. The Xboard protocol does have commands for sending the result of a game (`result 1-0 {Black resigned}`, for example), but this is not implemented in python-chess. So, I've implemented it in the Xboard wrapper `report_game_result()` by extracting the information from the last JSON message from lichess.

As I interpret it, the python-chess library is very much designed in a way similar to the UCI protocol. That is, the primary purpose of communicating with engines is to analyze board positions for the benefit of human users. Supporting the playing of full games is of secondary concern. That's why every command sent to a UCI engine before the `go` command conveys the full state of the game: it is not assumed that consecutive board positions are from the same game. Since engines are only tools for analysis, there is no need to tell the engine that the game has ended.

The Xboard protocol was designed with engines playing full games in mind. It is assumed that the engines remembers the state of the game and is only told about updates between moves. The engine is free to think and ponder as it sees fit without being commanded by the program sending commands (see the comments in the Critique and Contra sections of [this chessprogramming article on UCI](https://www.chessprogramming.org/UCI)). Because python-chess uses a communication style that is very similar to UCI, I might have to move Xboard clock-handling code to inside the lichess-bot Xboard engine wrapper once a future version of python-chess is released. See [this discussion](https://github.com/niklasf/python-chess/issues/742) for philosophical details.

TheYoBots:
Thanks a lot for the information! Really helped!

engine_wrapper.py
```diff
@@ -137,6 +143,49 @@ def search_with_ponder(self, board, wtime, btime, winc, binc, ponder):
...
         return self.search(board, time_limit, ponder)
 
+    def report_game_result(self, game, board):
+        # Send final moves, if any, to engine
+        self.engine.protocol._new(board, None, {})
+
+        winner = game.state.get('winner')
+        termination = game.state.get('status')
+
+        if winner == 'white':
```
ShailChoksi, line 153: can you move the magic strings to a variable/class? e.g. 

```
class Termination():
  MATE = "mate";
  TIMEOUT = "outoftime";
  RESIGNATION = "resign";
  ...
```

MarkZH:
Done. Let me know if this is what you had in mind.

PR #285, opened 2021-03-12
some problems in config.yml.default

OIVAS7572:
i got this error :
"chess.engine.EngineError: engine does not support option egtpath (available options: Debug Log File, Contempt, Analysis Contempt, Threads, Hash, Clear Hash, Ponder, MultiPV, Skill Level, Move Overhead, Slow Mover, nodestime, UCI_Chess960, UCI_AnalyseMode, UCI_LimitStrength, UCI_Elo, UCI_ShowWDL, SyzygyPath, SyzygyProbeDepth, Syzygy50MoveRule, SyzygyProbeLimit, Use NNUE, EvalFile
"
Check my request
If there is no problem add this

### DIFF
--- a/config.yml.default
+++ b/config.yml.default
@@ -33,11 +33,7 @@ engine:                      # engine settings
 # xboard_options:            # arbitrary xboard options passed to the engine
 #   cores: "4"
 #   memory: "4096"
-#   egtpath:                 # dir containing egtb, relative to this project
-#     gaviota: "Gaviota path"
-#     nalimov: "Nalimov Path"
-#     scorpio: "Scorpio Path"
-#     syzygy: "Syzygy Path"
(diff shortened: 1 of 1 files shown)


TheYoBots:
If you have added Stockfish engine, then it doesn’t run under xboard protocol and those options are only functional for xboard protocol. If you would like to add syzygy to your engine, then after line 28 in config.yml.default add a new line and set “SyzygyPath: ‘your path to syzygy’”.

OIVAS7572:
if you removed all lines strat with "#".
you can see that “SyzygyPath: ‘your path to syzygy’” this will come under uci option.
lines started with '#' are comments. python program don't read that comments

TheYoBots:
But this removes xboard support in config.yml.default file completely.

Issue #280, opened 2021-03-09
Unexpected engine output Error


TheYoBots:
I’ve been receiving an ‘Unexpected Engine Output’ errors, but am not sure on how to solve this. These are the logs.

This is the line of the error
```
2021-03-09 12:48:38,836: [33m<UciProtocol (pid=404)>: Unexpected engine output: 'Unknown command: ponderhitgo ponder wtime 13690 btime 2270 winc 0 binc 0'[0m
```

These are the lines before the error line

```
2021-03-09 12:48:38,643: [94mSearching for wtime 13803 btime 2270[0m
     depth: 245
     nps: 8605720
     nodes: 430286
     score: PovScore(Mate(+3), WHITE)
2021-03-09 12:48:38,756: [94mPondering for wtime 13690 btime 2270[0m
```
(Have removed my user info in logs)

After the error line, there are no logs and the bot simply crashes. In this case it was a mate in 3 and the bot just stopped playing after these logs. 

Do note that I have updated to the latest master build (https://github.com/ShailChoksi/lichess-bot/commit/936c516e5327d01eb7fb16efd4453513a294303a).

Any help would be appreciated. Thanks in advance.

MarkZH:
That's very strange. The commands `ponderhit` and `go ponder ...` should have been sent to your engine, not received from it. Even weirder is that the line break between the two commands was lost. Can you post your engine's communication logs showing what your engine sent and received leading up to the crash?

TheYoBots:
I’ve just used this client for testing and I received these logs and my bot crashed. Luckily since it was a mate in 3 I could manually play the moves till the mate.

Not sure if this is what you want, but these are the logs from the start of the game. (Do note that this is a different game as for the previous game, the logs are gone as I have closed my terminal.)
 
 My bot is YoBot_v2 here. This is how the logs are at the start of the game.

```
2021-03-09 19:00:42,161: [94m    Accept Bullet rated challenge from BOT YoBot_v2(2595)[0m
2021-03-09 19:00:42,248: [31mGiving up api_post(...) after 1 tries (requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://lichess.org/api/challenge/NwCyOWNG/accept)[0m
2021-03-09 19:00:42,248: [94m    Skip missing Bullet rated challenge from BOT YoBot_v2(2595)[0m
2021-03-09 19:00:48,749: [94m--- Process Used. Total Queued: 0. Total Used: 1[0m
[...]

PR #287, opened 2021-03-13
Fix UCI pondering issue

MarkZH:
The meaning of the ponder parameter of python-chess engine methods that
start searches has changed since the upgrade. Before, ponder=True would mean
that the current search was a pondering search. Now, ponder=True means
the engine should start pondering after choosing a move. Starting
pondering is handled by the python-chess library, so all pondering code
has been removed.

UCI engines will not receive a ponderhit command when their ponder
move is played by the opponent. They will only receive a stop command
followed by the next go command.

This should fix the issue discussed in #280.

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -36,20 +36,17 @@ class EngineWrapper:
     def set_time_control(self, game):
         pass
 
-    def first_search(self, board, movetime):
-        return self.search(board, chess.engine.Limit(time=movetime // 1000), False)
+    def first_search(self, board, movetime, ponder):
+        return self.search(board, chess.engine.Limit(time=movetime // 1000), ponder)
 
-    def search_with_ponder(self, board, wtime, btime, winc, binc, ponder=False):
+    def search_with_ponder(self, board, wtime, btime, winc, binc, ponder):
         pass
 
     def search(self, board, time_limit, ponder):
         result = self.engine.play(board, time_limit, info=chess.engine.INFO_ALL, ponder=ponder)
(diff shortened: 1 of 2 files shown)


ShailChoksi (approved):

amanjpro:
Not sure why this is merged. This clearly violates UCI protocol. My engine plays considerably worst when pondering due to this implementation. It takes time management control away from my engine

MarkZH:
@amanjpro Sorry about the regression. I've put in a bug report with the chess library to start work on correct pondering. Until then, I would recommend using the lichess-bot from [this commit](https://github.com/ShailChoksi/lichess-bot/tree/2ba454e6144c232fc75011f7859a0a1cc2f27587). This should restore the original pondering functionality. Be sure to run `pip install -r requirements.txt` after downloading from that commit.

@ShailChoksi Perhaps it would be a good idea to make a release at the above linked commit to direct people towards a fully UCI-compliant lichess-bot.

amanjpro:
Thank you Mark :) I really appreciate your comment and help

MarkZH:
@amanjpro With the latest updates, UCI engines should now have standard pondering behavior. Can you confirm this? You'll need to make sure to install the latest dependencies from `requirements.txt`.

amanjpro:
> @amanjpro With the latest updates, UCI engines should now have standard pondering behavior. Can you confirm this? You'll need to make sure to install the latest dependencies from `requirements.txt`.

Thank you, I'll give it a try

amanjpro:
I confirm it works as it should

Issue #131, opened 2018-06-29
Xboard versus UCI - a discussion


drbeco:
Dear friends,

I propose in this issue a discussion on how Xboard protocol is to be implemented, and how they differ from UCI.

I see that lichess-bot tries to "mimic" UCI logic using Xboard commands. After I tried really hard to update my engine to this strange process, I was myself distorting Xboard protocol to adapt to this new world.

I have here a UML diagram with one of the first source codes of Xadreco document. Please, lets see the difference:

![image](https://user-images.githubusercontent.com/5805111/42069090-1f423f1a-7b26-11e8-9f95-7dc3887e9776.png)

As we can see, the main "loop" for playing is between the engine states "`thinking`" and "`pondering`". As a state machine, they need a trigger to change, and that trigger is the `usermove` and the engine `move`.

When it is used "`setboard`" instead of "`new`" we go to a different state, and from there other functions may work, but also, others much more important when playing may not! (Like cache history of moves, opening book, etc.!)

That said, I'll still try to keep up with the current implementation, but I urge you guys to try to improve how Xboard protocol is handled in lichess-bot.

Thanks a lot. Keep up the good work!
Beco

ddugovic:
Since many old chess engines do use the [Chess-Engine Communication Protocol v2](http://hgm.nubati.net/CECP.html), this seems like a useful (although difficult) ask.

drbeco:
Some references cited at issue #128 should be here for fast access, so here they go:



Refs:

* http://wbec-ridderkerk.nl/html/UCIProtocol.html
* https://www.gnu.org/software/xboard/engine-intf.html
* http://hgm.nubati.net/CECP.html

careless25:
What is the minimum required here?
I am assuming you want the Observing -> Waiting -> Thinking -> Pondering loop implemented correctly? (and don't use the force and playother commands)

MarkZH:
I've been writing code and submitting pull requests to make lichess-bot more friendly to Xboard engines. I'm wondering if anyone can comment on whether lichess-bot is better now in terms of the Xboard protocol. If pull requests #193, #195, and #217 are accepted, I think we'll have a fully workable protocol.

drbeco:
Great news! Thanks to all envolved.

PR #288, opened 2021-03-13
Allow Xboard engines to ponder

MarkZH:
Now that pondering is handled by the python-chess library, Xboard
engines can now ponder on their opponent's time if desired.

The configuration uci_ponder is kept for backward compatibility.


MarkZH:
With this update, Xboard and UCI engines should be on equal footing, which would allow bug #131 to finally be closed.

Issue #234, opened 2020-11-15
"From Position" breaks after castling


imBigo:
The bot tries to play as the opposite side when it is playing "From Position" and usually happens to me if it is completely losing. Evaluation will go from -2000 to 2000 and it will get stuck trying to play from the winning side. I am using Lc0.
```
Traceback (most recent call last):
  File "D:\LichessBot\logging_pool.py", line 16, in __call__
    result = self.__callable(*args, **kwargs)
  File "C:\Users\user\AppData\Local\Programs\Python\Python39\lib\site-packages\backoff\_sync.py", line 99, in retry
    ret = target(*args, **kwargs)
  File "D:\LichessBot\lichess-bot.py", line 201, in play_game
    li.make_move(game.id, best_move)
  File "D:\LichessBot\lichess.py", line 74, in make_move
    return self.api_post(ENDPOINTS["move"].format(game_id, move))
  File "C:\Users\user\AppData\Local\Programs\Python\Python39\lib\site-packages\backoff\_sync.py", line 99, in retry
    ret = target(*args, **kwargs)
  File "D:\LichessBot\lichess.py", line 64, in api_post
    response.raise_for_status()
  File "C:\Users\user\AppData\Local\Programs\Python\Python39\lib\site-packages\requests\models.py", line 941, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://lichess.org/api/bot/game/hcLy9S2c/move/b1a3
```

This is the game where this problem occurred: 
https://lichess.org/hcLy9S2c/black

imBigo:
I have actually found out that the problem lies in castling. Whenever the bot or the other player castles, the bot tries to play a move as the opposite side and will crash.

imBigo:
I have found a fix for this issue for those using Leela. Modify the file **lichess-bot.py** so that you trick Leela into thinking the game is a chess960 game. This will allow Leela to send the correct form of castling. (Changes are in bold)
`def setup_board(game):
    if game.variant_name.lower() == "chess960":
        board = chess.Board(game.initial_fen, chess960=True)
    elif game.variant_name == "From Position":
        board = chess.Board(game.initial_fen, **chess960=True**)
    else:
        VariantBoard = find_variant(game.variant_name)
        board = VariantBoard()
    moves = game.state["moves"].split()
    for move in moves:
        board = update_board(board, move)

    return board`

[...]

PR #289, opened 2021-03-15
From Position castling fix

imBigo:
This fix is to address the From Position game mode that will break whenever a player castles. Lichess handles From Position the same as chess960 so it makes sense to tell the engine the same.
See [issue #234](https://github.com/ShailChoksi/lichess-bot/issues/234) for more info.


imBigo:
The issue seems to have been fixed with latest python-chess library so this is no longer needed.

PR #293, opened 2021-03-17
Regression fix: Disable pondering after first move

MarkZH:
Because the first move on lichess uses a different clock than the rest
of the game, engines cannot ponder after the first move because the
python-chess library sets the clock for pondering to be the same as the
move. After black's first move, it will think it has 10 seconds to
ponder no matter how short the time control and no matter how little
time white takes for its second move.

This was a mistake on my part introduced in the refactoring.


ShailChoksi:
can you fix the merge conflicts?

MarkZH:
Done.

ShailChoksi (approved):

PR #295, opened 2021-03-19
Add small move delay to avoid rate limiting

jazzzooo:
This might be an edge case, but my bot plays moves instantly, and when it plays multiple players in hyperbullet, keyboard users or other bots that move instantly, it would sometimes run into an 429 error. The problem is that after running into this error, the bot does not resume play on that game, and will consequently lose on time. My solution might not be sophisticated but it gets the job done, and the delay is negligible. 

After adding the delay, my bot was able to exploit this bug against a few other instant-moving bots, to make them lose on time. Here are some examples:
https://lichess.org/tXHK71bx/black
https://lichess.org/4qBo7GZ0/black
https://lichess.org/MOBAl4vY

In the first game, 26 moves were played within 0.3 seconds, at which point the opponent bot stopped making moves, presumably due to the same issue as I had. This shows that it's a real issue affecting real bots.

I believe that the behavior stems from is_final() in lichess.py, which considers 429 errors to be final, and will therefore not retry after receiving them. But I believe this is the intended behavior, so this shouldn't be changed? Perhaps another solution would be to only delay moves if the previous move was played very recently. This way, it would not affect slow games at all.

Here is the error:

```2021-03-19 15:42:29,502: Giving up api_post(...) after 1 tries (requests.exceptions.HTTPError: 429 Client Error: Too Many Requests for url: https://lichess.org/api/bot/game/hrhPYVIn/move/f1a6)
2021-03-19 15:42:29,547: Giving up api_get(...) after 1 tries (requests.exceptions.HTTPError: 429 Client Error: Too Many Requests for url: https://lichess.org/api/account/playing)
2021-03-19 15:42:29,547: Giving up play_game(...) after 1 tries (requests.exceptions.HTTPError: 429 Client Error: Too Many Requests for url: https://lichess.org/api/account/playing)
Traceback (most recent call last):
  File "/home/justas/documents/lichess-bot/lichess-bot.py", line 193, in play_game
    li.make_move(game.id, best_move)
  File "/home/justas/documents/lichess-bot/lichess.py", line 68, in make_move
    return self.api_post(ENDPOINTS["move"].format(game_id, move))
  File "/usr/lib/python3.9/site-packages/backoff/_sync.py", line 94, in retry
    ret = target(*args, **kwargs)
  File "/home/justas/documents/lichess-bot/lichess.py", line 58, in api_post
    response.raise_for_status()
  File "/usr/lib/python3.9/site-packages/requests/models.py", line 943, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 429 Client Error: Too Many Requests for url: https://lichess.org/api/bot/game/hrhPYVIn/move/f1a6

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/justas/documents/lichess-bot/logging_pool.py", line 17, in __call__
    result = self.__callable(*args, **kwargs)
  File "/usr/lib/python3.9/site-packages/backoff/_sync.py", line 94, in retry
    ret = target(*args, **kwargs)
  File "/home/justas/documents/lichess-bot/lichess-bot.py", line 208, in play_game
    if game.id not in (ongoing_game["gameId"] for ongoing_game in li.get_ongoing_games()):
  File "/home/justas/documents/lichess-bot/lichess.py", line 97, in get_ongoing_games
    ongoing_games = self.api_get(ENDPOINTS["playing"])["nowPlaying"]
  File "/usr/lib/python3.9/site-packages/backoff/_sync.py", line 94, in retry
    ret = target(*args, **kwargs)
  File "/home/justas/documents/lichess-bot/lichess.py", line 47, in api_get
    response.raise_for_status()
  File "/usr/lib/python3.9/site-packages/requests/models.py", line 943, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 429 Client Error: Too Many Requests for url: https://lichess.org/api/account/playing
```

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -160,6 +160,7 @@ def play_game(li, game_id, control_queue, engine_factory, user_profile, config, 
     is_uci = engine_cfg["protocol"] == "uci"
     is_uci_ponder = is_uci and engine_cfg.get("uci_ponder", False)
     move_overhead = config.get("move_overhead", 1000)
+    delay_seconds = config.get("rate_limiting_delay", 0)/1000
     polyglot_cfg = engine_cfg.get("polyglot", {})
 
     first_move = True
@@ -190,7 +191,7 @@ def play_game(li, game_id, control_queue, engine_factory, user_profile, config, 
                         else:
                             best_move = choose_move(engine, board, game, is_uci_ponder, start_time, move_overhead)
                     li.make_move(game.id, best_move)
-                    time.sleep(0.06)
+                    time.sleep(delay_seconds)
 
                 wb = 'w' if board.turn == chess.WHITE else 'b'
                 game.ping(config.get("abort_time", 20), (upd[f"{wb}time"] + upd[f"{wb}inc"]) / 1000 + 60)


TheYoBots:
This won’t cause too many timeouts right? Because if that is the case people would rather hit limits once in x many games that time out in game after x many moves.

Justazzz:
0.06 seconds per move would be 4.8 seconds for an 80-move game, that is considerable in bullet or hyperbullet. This is a tradeoff that the maintainers need to decide on. For me, 5 seconds at most is not a big cost, compared to losing games just when my bot is most active.

MarkZH:
I am neutral on whether this is a good change. What would make this idea better is to make the delay configurable. Let the bot users decide if this is a problem and how they want to solve it.

To `config.yml.default`, add a line under `fake_think_time` like
```
rate_limiting_delay: 0 # Time (in ms) to delay after sending a move to prevent "Too Many Requests" errors
```
To `lichess-bot.py`, add a line near the top of `start_game()` to get the value:
```
delay_seconds = config.get("rate_limiting_delay", 0)/1000
```
Then, use that value for `time.sleep()`
```
time.sleep(delay_seconds)
```

Justazzz:
Done

icecream17:
I have a suggestion
Some bots don't play instantly, so maybe instead of `pause(ms)`, you `if timetaken < ms: pause(timetaken - ms)`

OIVAS7572:
This delay is only needed for fast moves that rate limiting time by lichess.It don't need for every move.

OIVAS7572:
My suggestion, give 0.06 seconds to engine to think.It is better than delaying moves.I know this is hard.If possible try this.

MarkZH:
The suggestion from @icecream17 is better. If an engine thinks for less than the delay time, call `time.sleep()` with the difference. Otherwise, if the engine takes a longer time, skip the sleep.

icecream17:
Wait, the setting says "time in ms", but the code still divides it by 1000

icecream17:
Wait, you also have to account for your opponent too. If your opponent takes 12 seconds to make a move, you don't have to delay

MarkZH:
I've realized something. Since the `time.sleep()` happens after the move is submitted, no extra logic is necessary . The delay happens during the opponent's move. If the opponent takes longer than the delay time to think, the no sleeping occurs during this bot's move. If the opponent takes less time, the bot will finish whatever sleep is left before starting to think.

ShailChoksi (approved):

PR #302, opened 2021-03-30
Add the engine's PV to stats

imBigo:
This PR adds the engine's PV in beautiful regular notation (SAN) to the printed stats. A feature I feel is very useful.

Here is an exemple of what you can expect to be printed:
```ponderpv: 50...d4 51. Ba6 d3 52. Nc3 f3 53. Ke1 Kd4 54. Nb5+ Ke3 55. Nc3 f2+ 56. Kf1 Kf3```

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -48,16 +48,17 @@ class EngineWrapper:
     def search(self, board, time_limit, ponder):
         result = self.engine.play(board, time_limit, info=chess.engine.INFO_ALL, ponder=ponder)
         self.last_move_info = result.info
-        self.print_stats()
+        self.print_stats(board)
         return result.move
 
-    def print_stats(self):
-        for line in self.get_stats():
+    def print_stats(self, board):
+        for line in self.get_stats(board):
             logger.info(f"    {line}")
 
-    def get_stats(self):
+    def get_stats(self, board):
         info = self.last_move_info
-        stats = ["depth", "nps", "nodes", "score"]
+        info["ponderpv"] = board.variation_san(info["pv"])
+        stats = ["ponderpv", "depth", "nps", "nodes", "score"]
         return [f"{stat}: {info[stat]}" for stat in stats if stat in info]
 
     def get_opponent_info(self, game):

### CODE REVIEW

**Comment by imBigo** [comment]
Currently this PR seems to break the conversation part of the engine. This needs to be fixed but I'm not sure how to and would need some help.

**Comment by AttackingOrDefending** [comment]
I think it breaks the conversation because it is longer than the character limit. I don't think it is useful because the opponents will be able to see what move the bot thinks is the best and play it.

**Comment by amanjpro** [comment]
> I think it breaks the conversation because it is longer than the character limit. I don't think it is useful because the opponents will be able to see what move the bot thinks is the best and play it.

What if we print it to the stdout, and not send it to the other opponent? After all this is a debugging feature more or less

**Comment by AttackingOrDefending** [comment]
It breaks because https://github.com/ShailChoksi/lichess-bot/blob/master/conversation.py#L32 the arguement board should be passed. The character limit will be another problem.

**Comment by AttackingOrDefending** [comment]
If you want only print it to the stdout, you should seperate print_stats and get_stats. print_stats will have to use `info = self.last_move_info` instead of `self.get_stats()`.

**Comment by ShailChoksi** [comment]
Is this still a breaking change?

**Comment by imBigo** [comment]
Yes, I have not worked on it yet to fix this. I wish I could change this to a draft

**Comment by AttackingOrDefending** [comment]
You can use something like
```python
if "pv" in info:
    info["pv"] = list(map(lambda move: move.uci(), info["pv"]))
```
which doesn't need the board but the output will be
`pv: ['c8d7', 'b1d2', 'd8c7', 'd2c4', 'b7b5', 'c4b6', 'a8d8', 'c2c4', 'f8d6', 'g2g3', 'd6c5', 'b6d5', 'f6d5', 'c4d5', 'c6d4', 'e3d4', 'c7b6', 'd4c5', 'b6c5', 'a1c1', 'c5d6', 'f1h3', 'e8g8', 'd5e6', 'd7e6', 'h3e6', 'f7e6']`

AttackingOrDefending:
This is longer than the character limit. I would suggest adding a parameter to get_stats, so the function will return a different output if it is for the chat or for the cmd.
```python
def get_stats(self, for_chat):
    info = self.last_move_info.copy()
    if "pv" in info:
        info["pv"] = list(map(lambda move: move.uci(), info["pv"]))
    if for_chat:
        stats = ["pv", "depth", "nps", "nodes", "score"]
    else:
        stats = ["depth", "nps", "nodes", "score"]
    return [f"{stat}: {info[stat]}" for stat in stats if stat in info]
```
If you want to use variation_san you should pass the board from https://github.com/ShailChoksi/lichess-bot/blob/master/lichess-bot.py#L212 with try except
```python
try:
    conversation.react(ChatLine(upd), game, board)
except:
    pass
```
and then add board to the functions in conversation.py.

MarkZH:
Implemented by #340. Thanks for the idea.

PR #304, opened 2021-04-06
Enable strategies written directly in Python as an alt to engines. (duplicate)

icecream17:
Same as #187 but updated. 

Notable difference between this pr and #187 is that instead of a strategy function, you implement the whole class.
Which enables more functionality but takes more setup.

Adds docs folder (only 1 file)

Breaking change: Previously, if the protocol was not "xboard" it was assumed to be "uci".
But now if someone put something like "yci" it throws


docs/creating-a-custom-bot.md
```diff
@@ -0,0 +1,27 @@
+# Creating a custom bot
```
TheYoBots, line 1: Would be nice if this file is in the main README.md file. But make the change at own will.

TheYoBots (changes requested):

TheYoBots:
Also maybe linking something that shows a bunch of strategies written in python would allow further usage of this ‘protocol’ and it won’t be less extensively used [referring to https://github.com/ShailChoksi/lichess-bot/pull/187#issuecomment-792349236]

engine_wrapper.py
```diff
@@ -148,3 +156,8 @@ def get_opponent_info(self, game):
             self.engine.protocol.send_line(f"rating {game.me.rating} {game.opponent.rating}")
         if game.opponent.title == "BOT":
             self.engine.protocol.send_line("computer")
+
+def getHomemadeEngine():
+    raise NotImplementedError(
+        "    You haven't changed the getHomemadeEngine function yet!\n"
+        "    See docs/creating-a-custom-bot.md")
```
TheYoBots, line 163: I think after the request I made, you had forgotten to change this line.

TheYoBots (changes requested):

TheYoBots:
Does an engine have to be added for this to function? It would be nice if the engine was disabled all together, though I’m not sure of any way to do this.

icecream17:
If there's not a bug, a class only has to implement `search` and extend `MinimalEngine`:

```py
# strategies.py
class ExampleEngine(MinimalEngine):
    pass

class RandomMove(ExampleEngine):
    def search(self, board, *args):
        return random.choice(list(board.legal_moves))
```

MarkZH:
> If there's not a bug, a class only has to implement `search` and extend `MinimalEngine`:
> 
> ```python
> # strategies.py
> class ExampleEngine(MinimalEngine):
>     pass
> 
> class RandomMove(ExampleEngine):
>     def search(self, board, *args):
>         return random.choice(list(board.legal_moves))
> ```
> 
> Although `MinimalEngine` itself sets `self.engine = self` because of some methods in `EngineWrapper`

`name()` needs to be overridden, too.

ShailChoksi (approved):

Issue #173, opened 2019-03-26
Don't keep connection open for correspondence games


koedem:
When dealing with correspondence games it is not necessary to keep the connection open, instead one can open a connection, poll for new moves, then close it again and wait for e.g. 10 minutes.
The problem with the current way is when playing e.g. 10 correspondence games the client opens 10 connections to lichess which gets blocked as potential DDoS attack.

MarkZH:
This was implemented in PR #316.

PR #316, opened 2021-05-02
Improve Handling of Correspondence Games

nkoppel:
This PR improves handling of correspondence games by adding the configuration field "correspondence_time", which sets the time to think on moves in correspondence games, and by disconnecting from games 5 minutes after it makes a move and checking for opponent moves every 10 minutes, effectively implementing #173.


MarkZH:
I like the way this works. I'll try it out later, but I have some critiques first:


The variable names `first_state` or `first_gamestate` give no indication of what it controls. Instead, use something like

    correspondance_wait_time = 0

then, after making a move

    elif is_correspondence:
        best_move = choose_move_time(engine, board, correspondence_time, is_uci_ponder)
        correspondance_wait_time = 300

then

    game.ping(config.get("abort_time", 20), (upd[f"{wb}time"] + upd[f"{wb}inc"]) / 1000 + 60, correspondance_wait_time)
    correspondance_wait_time = 0

Plus, this better follows your description of the desired behavior: if it is not the bot's turn to move, disconnect immediately. After a bot move, wait 5 minutes for an opponent's move.

The initial `correspondance_wait_time = 0` can be moved after the line

    elif u_type == "gameState":

to keep it close to where it is used.

Along with this, `correspondence_time` should be renamed `correspondence_move_time`.

---

`engine.stop()` must come before `engine.quit()`. The quit command causes the engine to shut down immediately. It will not receive any commands afterwards.

---

In this block of code in lichess-bot.py (line 237):

    if is_correspondence:
        logger.info("--- {} Disconnecting from correspondence game".format(game.url()))
        control_queue.put_nowait({"type": "local_game_done"})
        corrospondence_wait(control_queue, game_id)

Shouldn't `correspondance_wait` be called asynchronously so the disconnection happens immediately?


----

In model.py in the Game class:

    def is_my_move(self):
        return len(self.state.get("moves").split()) % 2 == (0 if self.is_white else 1)

returns the wrong answer in From Position games where black plays the first move. The member `self.white_starts` needs to be considered.

Alternatively, you can move that logic out to lichess-bot.py using this:

    if is_correspondence and not is_engine_move(game, board) and game.should_disconnect_now():
        break

---

I need to think about how pondering is handled here. Is the idea here to allow pondering for 5 minutes just in case the opponent makes a move within that time?

nkoppel:
Thanks for the feedback!

I've implemented most of your suggestions in the most recent commit, with a few changes:

I've renamed your correspondence_wait_time variable to correspondence_disconnect_time, as I feel that correspondence_wait_time implies some connection with the correspondence_wait function.
I've kept the initialization for correspondence_disconnect_time outside of the outer loop so that the bot does not disconnect from the game immediately after moving and receiving a gameState where it is not the bot's move.

Also, I've redone the system to periodically check in on correspondence games so that it doesn't use any extra threads in the pool. It checks in on games when they haven't been checked in the current 10 minute interval and when either another 10 minute interval starts or a game ends or is disconnected from.

The idea with the 5 minute wait time is to allow opponents to play more synchronously with the bot and to not have to wait 10 minutes between every move. As for pondering, I think it would be best to have a separate "correspondence_ponder" option to let users decide if they want to have computation potentially wasted by searching for five minutes before killing the engine.

MarkZH:
The check-in system is much improved now.

---

I don't understand the purpose of the empty strings in the `correspondence_queue`. Can you explain what's going on there?

---

In the block that starts with

    if event["type"] == "correspondence_ping" or (event["type"] == "local_game_done" and not wait_for_correspondence_ping):

decrementing the `queued_processes` variable seems wrong. This will happen every ten minutes (`correspondence_ping`) regardless of whether the number of processes has changed.

nkoppel:
The empty strings in correspondence_queue signals to the program that it has checked in on all correspondence games in the last 10 minutes, so it should set wait_for_correspondence_ping to wait until the next 10 minute interval before checking in on more games.

---

> decrementing the queued_process variable seems wrong.

I agree. After thinking about it, I realized starting new processes without regard for queued processes could cause the program to run out of processes if it gets a correspondence_ping between accepting a challenge and a game starting. I'll fix this momentarily.

MarkZH:
At the moment, lichess-bot will disconnect immediately if the bot does not make a midgame thinking move. If the bot makes a book move or the first move of the game, the `correspondence_disconnect_time` will still be zero. I think the line 225 `correspondence_disconnect_time = correspondence_config.get("disconnect_time", 300)` needs to be moved before line 219 `best_move = get_book_move(board, polyglot_cfg)`.

---

     correspondence_config = config.get("correspondence", {})

This is not quite right. If the user has the following in their `config.yml`:

    correspondence:
        # move_time: 60            # time in seconds to search in correspondence games
        # checkin_period: 600      # how often to check for opponent moves in correspondence games that have been disconnected from
        # disconnect_time: 300     # time before disconnecting from a correspondence game

Then `config.get("correspondence", {})` will return `None`. You need to use `config.get("correspondence", {}) or {}`.

---

Finally, a comment on aesthetics, I would put a blank line above line 135:

    if event["type"] == "correspondence_ping" or (event["type"] == "local_game_done" and not wait_for_correspondence_ping):

to tell readers that this if statement is separate from the if-elif structure above it.

MarkZH:
I've started using this branch to play correspondence and it works quite well.

One thing I've noticed is that the time between checking correspondence games is twice as long as in my `config.yml`. I think this is because the empty string is at the front of the `correspondence_queue` about half the time.

At the start of the program, the empty string is at the front of the queue. The first time the `correspondence_ping` is encountered, the empty string will abort the queue iteration. Games will only be checked on the next `correspondence_ping` since the empty string is now at the back. However, since the games are started asynchronously, the empty string will be added to the queue before any `play_game()` can add their IDs to the queue. So, on the next `correspondence_ping`, no correspondence games will be checked because the empty string will be found first.

Here's what I propose:

1. The `correspondence_queue` should start empty.
2. Upon entering the block controlled by `if event["type"] == "correspondence_ping" or ... :`, put the empty string at the end. This will signal the end of the current crop of correspondence games.
3. Remove the empty string addition on line 143.

Personally, I like `None` as a sentinel value. But, that's a matter of taste.

---

Speaking of

    if event["type"] == "correspondence_ping" or (event["type"] == "local_game_done" and not wait_for_correspondence_ping):

What situation is handled by the right side of the `or`?

nkoppel:
The purpose of 
```python
event["type"] == "local_game_done" and not wait_for_correspondence_ping
```
is to let the program check in on another correspondence game every time a worker thread gets freed and while there are games that haven't yet been checked in on. It essentially ensures that all correspondence games are checked at the same rate even if there are more games than threads.

My most recent commit should fix the issue you mentioned, but in a slightly different way than you described. It will only add an empty string when it has encountered and removed one, to prevent more than one empty string from ever being in the queue and slowing down the processing of the queue.

MarkZH:
I see how that works now.

This pull request looks complete. I'll keep running it and looking it over, but I would be comfortable merging this.

ShailChoksi:
Can you resolve the merge conflicts? Thank you!

nkoppel:
Ok, I've resolved the merge conflicts and things seem to be working, but you all should check things over.

MarkZH:
There will be at least one more conflict once #293 gets merged.

ShailChoksi:
Once more - thanks :)

nkoppel:
Done!

ShailChoksi:
Can you add a check that correspondence is enabled under the `variants` section before accepting challenges that are correspondence?

MarkZH:
@ShailChoksi This is already implemented. [This line in start() in lichess-bot.py](https://github.com/ShailChoksi/lichess-bot/blob/da69083ced83fc6fe6881ba1060aee4b64e9f5c3/lichess-bot.py#L123) calls the method [Challenge.is_supported()](https://github.com/ShailChoksi/lichess-bot/blob/da69083ced83fc6fe6881ba1060aee4b64e9f5c3/model.py#L33) which checks if a game challenge is supported by the `config.yml`. I've tested this by commenting out the `- Correspondence` line in my own `config.yml` and correspondence games are declined.

MarkZH:
@OIVAS7572 You can have an unlimited number of simultaneous correspondence games active at the same time. However, the bot will only be to think about one game at a time if concurrency is 1. After choosing a move and waiting for a time (`disconnect_time` in `config.yml` in the `correspondence` section), lichess-bot will disconnect from that game and move on to the next game. Disconnected games are not counted towards the number of concurrent games. However, if the bot is currently connected to a game (either real-time or correspondence), it will not be able to start a new game or connect to a another correspondence game until it disconnects.

If the bot starts a real-time (non-correspondence) game in the interval between checking for correspondence updates, then checking on the correspondence games will be blocked until the real-time game is over.

OIVAS7572:
@MarkZH 
Thanks for clearing the doubt

PR #320, opened 2021-05-09
Workflows for every OS and python version

AttackingOrDefending:
It tests the code for every OS and python version that lichess-bot supports.


TheYoBots:
Nice, will close #266

ShailChoksi (approved):

PR #321, opened 2021-05-09
Test updates by playing a game

AttackingOrDefending:
This PR downloads SF and plays one game to test the update. It will only work if someone puts their token https://github.com/AttackingOrDefending/lichess-bot/blob/lichess-test/tests.py#L88. I was thinking that ShailChoksi could make a new bot and put its token here so every PR will be tested by playing a game. Maybe I should reduce the time control (e.g. 0+1) so it doesn't use a lot of resources on Github. It fails one test because there is no token to play the game. The PR also aborts every game that it can in the testing bot. It aborts games is because if a bot crashes before playing a move the game will not be aborted because sometimes it starts 3 games, although it still plays only 1.


TheYoBots:
Can `logger.info()` be used instead of `print()` (https://github.com/ShailChoksi/lichess-bot/commit/132ac8b7da12b9a1d856b342cb2b9f1184332153)

MarkZH:
It's a good idea to rename the file `test_nothing.py` since it now tests more than nothing.

tests.py
```diff
@@ -0,0 +1,68 @@
+import pytest
+import zipfile
+import requests
+from shutil import copyfile
+copyfile('lichess-bot.py', 'lichess_bot.py')
```
MarkZH, line 5: Alternatively, you could do this:

    import importlib
    lichess_bot = importlib.import_module("lichess-bot")

This way, there's no extra file left over after the tests.

AttackingOrDefending:
TheYoBots made a new bot and provided the api token (https://github.com/AttackingOrDefending/lichess-bot/pull/1). Shall I use his token or will ShailChoksi make a new one (because they are the owner of lichess-bot)?

AttackingOrDefending:
pytest doesn't seem to end if there is a bug unless I use assert so I lowered the AI level and if SF doesn't win the AI it will be assumed that it is a bug (because lichess-bot will restart and will lose on time because of backoff). I don't know how to get the response with json (https://github.com/AttackingOrDefending/lichess-bot/blob/lichess-test/tests.py#L48) using https://lichess.org/api#operation/gamePgn. Does anyone know how I can get the output in json?

.github/workflows/python-build-rest.yml
```diff
@@ -0,0 +1,37 @@
+# This workflow will install Python dependencies, run tests and lint
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Python Build
```
TheYoBots, line 4: Could you change the name of either this workflow or the other? Maybe name this Python Build Rest and leave the other one as Python Build or leave this as Python Build and name the other workflow lichess-bot Build?

TheYoBots (changes requested):

TheYoBots:
- ‘Code Errors’ isn’t honestly a suitable workflow name IMO. It doesn’t convey much. Maybe ‘Python Tests’ would make more sense as that workflow tests the python code.
- ‘Tests’ isn’t a great name too again for the same reason. Maybe ‘lichess-bot Build’ or ‘Python Run’?
- If you are changing the name of the workflows, do make sure to change the status badge too in the README.md file

AttackingOrDefending:
They can cheat but this is a BOT account. They can also post in a team chat but I think this can be stopped with the permissions of the token. They can play rated games and resign but this will get them banned and if we also get banned, we won't be able to play rated games, which will solve the problem. I can only see 2 problems:

1. Spamming the game chat (spectator and between the 2 players)
2. Automatically resigning every game so we can't test them

If the 2nd option happens we can make a new file for tests from github and have this file, so we can test this on our own computers (with our own token), so we would still have automated tests.

AttackingOrDefending:
I think I can change the code so it gets the token from environment secrets, so nobody would know the token apart from ShailChoksi. Github prevents secrets from being printed in the logs, but I can think of 2 ways for someone to get the token:

1. Print it with some changes (e.g. instaed of 'y4y278h' print 'y4y2-78h')
2. Send it to a website that stores it, so you can later get it.

The advantage is that we can see who made a PR that printed the token.
Shall I implement this change?

AttackingOrDefending:
I have already renamed the files in .github/workflows a few times but I don't really like the names. Does anyone have any good names for the files? I am think of renaming them from 'Python Build' to 'Build' and from 'Lichess Bot Build' to 'Test' (or 'Tests').

AttackingOrDefending:
I made a simple environment that tries to emulate lichess, so there is no need for a token (https://github.com/AttackingOrDefending/lichess-bot/tree/test-offline).

Pros:
- No token needed
- Can run offline

Cons:
- Doesn't test the functions in lichess.py because it replaces them
- It hasn't implemented every function in lichess.py (e.g. chat)

Note: If chat is implemented, we can add the script to send a chat command (e.g. eval) to the engine, so we can test the engine's response to commands. This can be done also for other functions.

Should I make a new PR with the new branch and close this one?

AttackingOrDefending:
Closing since #404 was merged.

PR #323, opened 2021-05-10
XBoard support for go commands

TheYoBots:
Still not sure how well this would work. Accepting any reviews/feedback.


MarkZH:
I think the `uci_` and `xboard_` prefixes are redundant, since these options are already grouped under `uci_options` and `xboard_options`. In fact, if these prefixes are deleted, then both `UCIEngine.search_with_ponder()` and `XBoardEngine.search_with_ponder()` will be identical, which would allow the method to be defined once in the `EngineWrapper` class.

MarkZH:
It's fine for any of the `go_commands` to be `None`. There are a few options:

1. Listing only `depth` in the `config.yml.default` under `xboard_options`.
2. Put a small note that `movetime` and `nodes` are invalid for XBoard engines.
3. Remove these two options from the `XBoardEngine` `self.go_commands` in the `__init__()` method.

I think 1. and/or 2. are fine.

MarkZH:
Just to have a concrete answer, if it were me, I would write

    xboard_options:
      go_commands:
        depth: 5 # Limit search to this depth
        # The go commands movetime and nodes will cause bad time management for XBoard engines.

TheYoBots:
Would that work even after leaving the `engine_wrapper.py` file unchanged or I’ll have to do it up in the `engine_wrapper.py` file too?

MarkZH:
Do you mean unchanged from how it is now in this branch or unchanged from lichess-bot master?

TheYoBots:
Unchanged from how it is in this branch.

MarkZH:
Yes. The `engine_wrapper.py` as it is now in the branch will work. Any `go_command` that is not listed will yield `None` in `search_with_ponder()`. The statement `nodes=cmds.get("nodes")` in line 79 would yield `nodes=None`, which the python-chess library handles by skipping the relevant engine commands.

TheYoBots:
Thanks for your help Mark! I’ll leave this as ‘ready for review’ for the author now. Sorry for the bunch of commits above, my terminal crashed and I had to use an online terminal (goorm) which resulted in a bunch of weird stuff while resolving conflicts. But that’s all resolved now.

MarkZH:
I think this PR works. I'll try it out with my bot tomorrow.

In the meantime, some style changes:

Please indent the "invalid" comment in `config.yml.default` to match `depth: 5`:

    #   go_commands:             # Additional options to pass to the XBoard go command.
    #     depth: 5               # Search depth ply only.
    #     Do note that the go commands 'movetime' and 'nodes' are invalid and may cause bad time management for XBoard engines.

Also, please delete the indentations added to some of the blank lines of `engine_wrapper.py` (line 151 and below).

ShailChoksi (approved):

PR #330, opened 2021-05-16
Add slight whitespacing when printing 'logger' logs

TheYoBots:
When printing 'logger' logs, in some OS/Cloud Hosts a weird `[94m` is seen before the logs are print (for example, `2021-05-15 21:10:47,966: [94mmove: 5`). To prevent any confusion and joining of `[94m` and the print statement, I have added a slight white spacing.

Here’s a before/after comparison:

**Before**

```
 2021-05-1521:10:47,966: [94m
     .   _/|
     .  // o\
     .  || ._)  lichess-bot 1.2.0
     .  //__\
     .  )___(   Play on Lichess with a bot
     [0m
 2021-05-15 21:10:48,039: [94mWelcome YoBot_v2![0m
 2021-05-15 21:10:48,039: [94mYou're now connected to https://lichess.org/ and awaiting challenges.[0m
```

**After**

```
 2021-05-1521:10:47,966: [94m
     .   _/|
     .  // o\
     .  || ._)  lichess-bot 1.2.0
     .  //__\
     .  )___(   Play on Lichess with a bot
     [0m
 2021-05-15 21:10:48,039: [94m      Welcome YoBot_v2![0m
 2021-05-15 21:10:48,039: [94m      You're now connected to https://lichess.org/ and awaiting challenges.[0m
```

If you’d like a few other examples do comment. This is mainly applicable when printing move number which is changed from `2021-05-15 21:10:47,966: [94mmove: 5` to `2021-05-15 21:10:47,966: [94m      move: 5`. The two `m`‘s together made it slightly confusing.

### DIFF
--- a/ColorLogger.py
+++ b/ColorLogger.py
@@ -117,7 +117,7 @@ def add_coloring_to_emit_ansi(fn):
             # #'\x1b[35m' # pink
         else:
             color = '\x1b[0m'  # normal
-        args[1].msg = color + args[1].msg + '\x1b[0m'  # normal
+        args[1].msg = f'{color}    {args[1].msg}\x1b[0m'  # normal
         # print "after"
         return fn(*args)
 
--- a/ColorLogger.py
+++ b/ColorLogger.py
@@ -117,7 +117,7 @@ def add_coloring_to_emit_ansi(fn):
             # #'\x1b[35m' # pink
         else:
             color = '\x1b[0m'  # normal
-        args[1].msg = f'{color}    {args[1].msg}\x1b[0m'  # normal
+        args[1].msg = f'{color}  {args[1].msg}\x1b[0m'  # normal
         # print "after"
         return fn(*args)
 
--- a/config.py
+++ b/config.py
@@ -11,7 +11,7 @@ def load_config(config_file):
         try:
             CONFIG = yaml.safe_load(stream)
         except Exception as e:
-            logger.error("There appears to be a syntax problem with your config.yml")
+            logger.error("    There appears to be a syntax problem with your config.yml")
             raise e
 
         # [section, type, error message]
--- a/config.py
+++ b/config.py
@@ -11,7 +11,7 @@ def load_config(config_file):
         try:
             CONFIG = yaml.safe_load(stream)
         except Exception as e:
-            logger.error("    There appears to be a syntax problem with your config.yml")
+            logger.error("There appears to be a syntax problem with your config.yml")
             raise e
 
         # [section, type, error message]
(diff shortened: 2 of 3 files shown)


TheYoBots:
Well, all I can say is that this is **not** a breaking change and fails for some reason (I’m not sure what the reason is). **I guess** it should be fine once it is merged.

TheYoBots:
Thanks to AttackingOrDefending!
> The error code is 500, so it is a server error.

https://github.com/ShailChoksi/lichess-bot/pull/328#issuecomment-841783579

MarkZH:
Those characters are for coloring text in terminals. You can see what the values mean in ColorLogger.py.

TheYoBots:
> Those characters are for coloring text in terminals. You can see what the values mean in ColorLogger.py. Another possibility is to bypass enable_color_logging() on line 430 of lichess-bot.py with another command-line option like --nocolor.

It can be a solution, but won’t colored text be better? What I have done allows both colored text and no confusion by just simply adding a slight white spacing.

MarkZH:
Sure. Colored text is useful. I was thinking that coloring text could be disabled if the log is being written to a file (whether through the `--logfile` option or redirecting the output), which is where the extra characters appear.

In any case, a better place to put the whitespsce is [here](https://github.com/ShailChoksi/lichess-bot/blob/f6813e48655ea36791d6905a4e7491aec39e5a71/ColorLogger.py#L120). That way, all logging messages are modified automatically, even if new messages are added. This only affects non-Windows systems, so someone needs to check what a log file looks like on Windows to see if the spacing is necessary there.

MarkZH:
On Windows, the extra characters don't appear. So, the only change needed is the line I linked to above:

    args[1].msg = color +'    ' + args[1].msg + '\x1b[0m'  # normal

Or, if you want to get fancy with f-strings

    args[1].msg = f'{color}    {args[1].msg}\x1b[0m'  # normal

TheYoBots:
Thanks for the suggestion, I’ll revert all changes and do it only in the ColorLogger file soon.

MarkZH:
Two cosmetic comments: (1) A lot of logging statements already have indentation, so I would reduce the new spacing to 1 or 2 spaces. (2) The change from single to double quotes in `lichess-bot.py` seems unnecessary.

TheYoBots:
> (1) A lot of logging statements already have indentation, so I would reduce the new spacing to 1 or 2 spaces.

So would it be possible to to change this directly in the ColorLogger.py file? Or so I do that in all logger print statements?

> (2) The change from single to double quotes in lichess-bot.py seems unnecessary.

That was just to make it uniform with all other quotes in all other files. I could revert that, though IMO it isn’t necessary.

MarkZH:
Just in ColorLogger.py. I was only concerned about adding too much more indentation.

ShailChoksi (approved):

PR #340, opened 2021-05-21
Add the engine's PV to stats (#302 duplicate)

AttackingOrDefending:
Updated version of https://github.com/ShailChoksi/lichess-bot/pull/302.
It now adjustst how many moves the pv will show depending on how many characters.

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -83,16 +83,24 @@ class EngineWrapper:
     def search(self, board, time_limit, ponder):
         result = self.engine.play(board, time_limit, info=chess.engine.INFO_ALL, ponder=ponder)
         self.last_move_info = result.info
-        self.print_stats()
+        self.print_stats(board)
         return result.move
 
-    def print_stats(self):
-        for line in self.get_stats():
+    def print_stats(self, board):
+        for line in self.get_stats(board):
             logger.info(f"{line}")
 
-    def get_stats(self):
-        info = self.last_move_info
-        stats = ["depth", "nps", "nodes", "score"]
+    def get_stats(self, board, for_chat=False):
+        info = self.last_move_info.copy()
+        if for_chat:
+            stats = ["depth", "nps", "nodes", "score", "ponderpv"]
+            bot_stats = [f"{stat}: {info[stat]}" for stat in stats if stat in info]
+            bot_stats = ", ".join(bot_stats)
+            pv_moves = ((140 - len(bot_stats) - 12) // 15) * 2
+            info["ponderpv"] = board.variation_san(info["pv"][:pv_moves])
+        else:
+            stats = ["depth", "nps", "nodes", "score", "ponderpv"]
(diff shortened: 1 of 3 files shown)


AttackingOrDefending:
It fails the tests because board is undefined but this error is handled by the try except so it doesn't cause any problems.

engine_wrapper.py
```diff
@@ -83,16 +83,24 @@ def search_with_ponder(self, board, wtime, btime, winc, binc, ponder):
...
-    def get_stats(self):
-        info = self.last_move_info
-        stats = ["depth", "nps", "nodes", "score"]
+    def get_stats(self, board, for_chat=False):
+        info = self.last_move_info.copy()
+        if for_chat:
+            stats = ["depth", "nps", "nodes", "score", "ponderpv"]
+            bot_stats = [f"{stat}: {info[stat]}" for stat in stats if stat in info]
+            bot_stats = ", ".join(bot_stats)
+            pv_moves = ((140 - len(bot_stats) - 12) // 15) * 2
```
MarkZH, line 99: How does this calculation work? What do 140, 12, 15, and 2 represent?

AttackingOrDefending: 140 is the chat character limit in lichess. 12 is the length of ', ponderpv: '. 15 is an estimation on how many characters, both players moves are going to be (e.g. '1. e4 e5' (8 characters) or '153. Nfxd3+ Nfxd3+' (for position 8/8/8/2n1k3/5n2/3b4/1N3N2/4K3 w - - 0 153) (18 characters)). Most of the time the length will be less than 15 but to be safe, I put a high number. 2 is there because '1. e4 e5' has 2 moves in it, so we know how many moves we should get. I use // 15 and then * 2 instead of // 7 becuase '1. ' also takes up characters.

MarkZH: I see. That makes sense.

Maybe this alternate way could work, too:

1. In this method, return all stats including the full `ponderpv` (just use the `else` branch).
2. In the `command()` method of the `Conversation` class, if the result of `join` is longer than 140 characters, truncate to 137 characters and add  `...`.

This should get you the maximum number of moves while staying under 140 characters. This also negates the need for the `for_chat` parameter.

AttackingOrDefending: I changed it now so the output is always less than 140 characters (there was a very small chance that the output would be more than 140 characters in the previous version). If I used the output and truncated it to 137 characters, it would split moves (e.g. '1. e4 e5 2. Nf...'). I prefer it with the for_chat parameter because someone may want to print more stats to their cmd (e.g. hashful, tbhits, seldepth), but they don't want them to be printed in the chat. When someone plays with a bot they can write !eval in the spectator room and get a move to play from the bot (if the bot doesn't ponder they will be able to see the move while if it does they can guess it from the following moves).

ShailChoksi: Can you name these magic numbers accordingly into constants?

MarkZH (approved):

Issue #339, opened 2021-05-20
No priority in correspondence games


OIVAS7572:
Problem is happening when in these conditions :
* One day correspondence games.
* 50+ correspondence games at a time.

When bot playing large number of correspondence games at a time, Bot is not giving any priority to low time correspondence games. These leads to Timeout in a large number of correspondence games. It will lead to get message about restrictions from Lichess.
 Can anybody solve this problem. Please take it serious problem

MarkZH:
What settings do you have in the `correspondence` and `concurrency` sections of your `config.yml`?

OIVAS7572:
I found this problem after restarting bot.

concurrency: 1
correspondence:  
move_time: 60      
checkin_period: 1200  (also tryed 2 hour and 5 mins)
disconnect_time: 5 (also tryed 10sec ) 
ponder: false

nkoppel:
I can confirm that there is not system for prioritizing daily games that are low on time. Even so, it seems strange that daily games are timing out even with 50+, as the end of every game should the bot to check all correspondence games and to reject all challenges until it is has checked every game.

@OIVAS7572 Could you send over the output of the bot when it is failing to move in daily games?

OIVAS7572:
@nkoppel
I didn't found any error in logs.
Did you need verbose logs.
Or small logs

OIVAS7572:
2021-05-21T02:06:43.570248+00:00 app[worker.1]: 2021-05-21 02:06:43,570: [94m--- Process Used. Total Queued: 0. Total Used: 1[0m
2021-05-21T02:06:44.040475+00:00 app[worker.1]: 2021-05-21 02:06:44,040: [94m+++ https://lichess.org/cVxiNpTe/black Correspondence vs BOT Official_bot(1395?)[0m
2021-05-21T02:06:50.683185+00:00 app[worker.1]: 2021-05-21 02:06:50,683: [94m--- Disconnecting from https://lichess.org/cVxiNpTe/black[0m
2021-05-21T02:06:50.684963+00:00 app[worker.1]: 2021-05-21 02:06:50,684: [94m+++ Process Free. Total Queued: 0. Total Used: 0[0m
2021-05-21T02:06:50.685197+00:00 app[worker.1]: 2021-05-21 02:06:50,685: [94m--- Process Used. Total Queued: 0. Total Used: 1[0m
[...]

PR #341, opened 2021-05-21
Put Active Correspondence Games into Queue on Startup

nkoppel:
This pr fixes the issue in #339 which caused correspondence games that could not be given a thread at startup to be forgotten by the bot and for the bot to never move in them. It does so by pushing game ids the main thread receives through a gameStart event and can't start a process for into the game queue. This behavior works with normal games as well, because games started from the queue are started in exactly the same way as from a gameStart event.

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -112,7 +112,7 @@ def start(li, user_profile, engine_factory, config, logging_level, log_filename)
     correspondence_pinger.start()
     correspondence_queue = manager.Queue()
     correspondence_queue.put("")
-    game_start_queue =  manager.Queue()
+    startup_correspondence_games = [game["gameId"] for game in li.get_ongoing_games() if game["perf"] == 'correspondence']
     wait_for_correspondence_ping = False
 
     busy_processes = 0
@@ -161,16 +161,15 @@ def start(li, user_profile, engine_factory, config, logging_level, log_filename)
                         pass
             elif event["type"] == "gameStart":
                 game_id = event["game"]["id"]
-                if busy_processes >= max_games:
-                    game_start_queue.put(game_id)
+                if game_id in startup_correspondence_games:
+                    correspondence_queue.put(game_id)
+                    startup_correspondence_games.remove(game_id)
                 else:
-                    # skip correspondence games if this gameStart is from the initial connection to lichess
-                    skip_correspondence = queued_processes == 0
                     if queued_processes > 0:
                         queued_processes -= 1
                     busy_processes += 1
                     logger.info("--- Process Used. Total Queued: {}. Total Used: {}".format(queued_processes, busy_processes))
-                    pool.apply_async(play_game, [li, game_id, control_queue, engine_factory, user_profile, config, challenge_queue, correspondence_queue, logging_queue, game_logging_configurer, logging_level, skip_correspondence])
+                    pool.apply_async(play_game, [li, game_id, control_queue, engine_factory, user_profile, config, challenge_queue, correspondence_queue, logging_queue, game_logging_configurer, logging_level])
 
             if event["type"] == "correspondence_ping" or (event["type"] == "local_game_done" and not wait_for_correspondence_ping):
                 if event["type"] == "correspondence_ping" and wait_for_correspondence_ping:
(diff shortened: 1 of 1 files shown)


MarkZH:
I think the correspondence games need to be treated separately from real time games. If there are both types of games active on startup and correspondence games happen to take up all the processes first, then the real time games will be delayed from starting, possibly resulting in time out losses. Maybe if `queued_processes == 0` (indicating lichess-bot has been restarted during games) correspondence games are queued and real time games are started.

nkoppel:
The changes from a056 to 225e implement prioritizing real-time games over correspondence games using the following strategy:

1. At startup, start all games we can, but skip all correspondence games and place them in the correspondence queue
2. Place all games we can't start into "game_start_queue"
3. While there are games in the game start queue, start them instead of starting correspondence games, also skipping correspondence games
4. Once the game start queue is empty, function as normal

This should work as intended unless the "local_game_done" event from the first game is handled before the second "gameStart" event, which would cause the program to wait on all active games for "checkin_period", but I do not think this will ever happen.

---

I've also removed this line
```python
logger.debug("Something went wrong. Game is starting and we don't have a queued process")
```
because having no queued process no longer means that anything went wrong.

lichess-bot.py
```diff
@@ -110,9 +110,11 @@ def start(li, user_profile, engine_factory, config, logging_level, log_filename)
     correspondence_checkin_period = correspondence_cfg.get("checkin_period", 600)
     correspondence_pinger = multiprocessing.Process(target=do_correspondence_ping, args=[control_queue, correspondence_checkin_period])
     correspondence_pinger.start()
-    correspondence_queue = manager.Queue()
-    correspondence_queue.put("")
+    correspodence_queue = manager.Queue()
+    correspodence_queue.put("")
```
MarkZH, line 114: Spelling: correspondence_queue (also in play() function)

MarkZH:
> but I do not think this will ever happen.

You just had to tempt fate, didn't you? ;)

This change looks correct. I think your last comment about the "shouldn't happen" situation is correct as I would imagine all of the `gameStart` messages would come at once on startup. If it actually does turn out to be a problem, you might be able to make use of `Lichess.get_ongoing_games()` in `lichess.py` to get a list of ongoing games all at once before entering the main loop.

Now that I think about it, what happens if a challenge is received before getting through all the ongoing games? I think it's the same as `local_game_done`: all of the `gameStart` messages should be front-loaded in the queue at startup so nothing else can get in front of them. No message can jump in front of any other message since they all get placed in the `control_queue` upon receipt.

So, yeah. Still looks correct.

---

Another possible idea: you can get a [list of all ongoing correspondence games](https://lichess.org/api#operation/apiAccountPlaying) at startup with (adapted from [this line](https://github.com/ShailChoksi/lichess-bot/blob/6c41ae28e824b1e5009e55bca999c77585b00355/lichess-bot.py#L290)):
```python
correspondence_games = [game["gameId"] for game in li.get_ongoing_games() if game["perf"] == 'correspondence']
```

Then, in the `gameStart` branch:
```python
elif event["type"] == "gameStart":
    if game_id in correspondence_games:
        correspondence_queue.put(game_id)
        correspondence_games.remove(game_id)
    else:
        # Do the normal thing
```
This way, starting correspondence games can be avoided until all current games have been processed, avoiding delays for real time games. This might negate the need for a `game_queue` since all real time games could be started immediately.

nkoppel:
Ok, I like that idea a lot, as it will simplify the pr quite a bit. And while some real-time games could be forgotten about if users reduce the number of threads below the number of active games, some games are likely to time out anyways if they do that, it was like that way before, and it is the result of user error.

MarkZH:
I don't think any real-time games will be forgotten. There isn't `busy_processes < max_games` check in the `gameStart` branch. All games that were active before the restart will be active afterwards, even if the `concurrency` is lowered before the restart.

Anyway, I like the latest version of this PR. Very simple and easy to understand.

MarkZH:
This is a note for my own education and a rebuttal to the first paragraph of my previous post. The `gameStart` branch does not check if `(busy_processes + queued_processes) < max_games`. But, the process pool is created with
```python
with logging_pool.LoggingPool(max_games + 1) as pool:
```
This means that `pool.apply_async(play_game, ...)` will not start the game if `max_games + 1` games have already been started. All subsequent games will be stalled until enough other games are completed. So, starting 50 games all at once with `concurrency: 1` means that 49 of them will be stalled, so any real time games will likely time out while the correspondence games make their way through the pool.

AttackingOrDefending:
Can you resolve the conflicts?

MarkZH (approved):
This will be merged once the conflicts are resolved.

MarkZH:
@nkoppel Thanks for all the great work. These changes were completed and merged in PR #424.

PR #344, opened 2021-05-27
Homemade engine improvements

AttackingOrDefending:
Now, if someone wants to use homemade engines, they can just change the name in config.yml and not change engine_wrapper.py. Also, if someone used a homemade engine, they would have to change the name in config.yml because they would get an error
`Exception: The engine ./engines/engine_name file does not exist.`

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,25 +94,15 @@ Steps to create a custom bot
 
 1. Do all the steps in the [How to Install](#how-to-install)
 2. In the `config.yml`, change the engine protocol to `homemade`
-3. Create a class in some file that extends `EngineWrapper` (in `engine_wrapper.py`)
-    - Or extend `MinimalEngine` (in `strategies.py`),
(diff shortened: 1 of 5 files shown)


MarkZH:
I think the instructions in the *Creating a custom bot* section in the README file need updating, too.

MarkZH:
I think the README could use a short introduction explaining what a homemade bot is. I've written a short paragraph below that could go at the beginning of the section. Plus, I replaced the word "custom" with "homemade" for consistency.

```md
## Creating a homemade bot

As an alternative to creating an entire chess engine and implementing one of the communiciation protocols (UCI or XBoard), a bot can also be created by writing a single class with a single method. The `search()` method in this new class takes the current board and the game clock as arguments and should return a move based on whatever criteria the coder desires.

Steps to create a homemade bot:
```

ShailChoksi (approved):

PR #353, opened 2021-06-16
Inherit `search_with_ponder` from `EngineWrapper`

icecream17:
Nowadays `search_with_ponder` is outdated.

____
The inherited `search_with_ponder` requires `go_commands` be defined, so set that too


icecream17:
The Fix Attribute error pr is now combined with this one. It might make sense to have UCIEnginge and XBoardEngine call `super` if you want

MarkZH (approved):

AttackingOrDefending (approved):

MarkZH (approved):

PR #354, opened 2021-06-16
Fix AttributeError

icecream17:
UCI Engines don't have the property `go_commands`, so this gets an attribute error:

2021-06-15 22:12:42,151: move: 2
2021-06-15 22:12:42,151: Searching for wtime 231030 btime 240000
2021-06-15 22:12:42,152: Backing off play_game(...) for 4.1s (AttributeError: 'Thousandatom' object has no attribute 'go_commands')
2021-06-15 22:12:46,631: +++ https://lichess.org/9scg1Sa3/white Blitz vs AI level 1
2021-06-15 22:12:46,633: --- https://lichess.org/9scg1Sa3/white Game over

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -1,4 +1,5 @@
 import os
+import chess
 import chess.engine
 import backoff
 import subprocess
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -1,5 +1,4 @@
 import os
-import chess
 import chess.engine
 import backoff
 import subprocess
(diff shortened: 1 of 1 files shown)


MarkZH:
Do you mean `homemade` engines don't have a `go_command` member?

I think a better place for this is in the `MinimalEngine.__init__()` method. Just set `self.go_commands = {}`.

Wouldn't these changes be better placed in PR #353? That PR is what makes this PR necessary, so they should be merged at the same time.

icecream17:
Oh yeah, didn't realize that

PR #356, opened 2021-06-22
Update lichess.py

TG-KRISH:
Does it work?


MarkZH:
What is the reason for these changes?

TG-KRISH:
I just try this out

PR #361, opened 2021-06-26
(small but good)

TG-KRISH:
The code gets neater, I ran this but I have an issue with stockfish on my laptop 
so I just came with this
the code is small but in my repository, it works well
[See here](https://github.com/PYTH0N-B0T/Lichess-B0T)
but still not finished I thought about it and worked hard

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -1,47 +1,51 @@
 import argparse
 import chess
-from chess.variant import find_variant
+from chess import engine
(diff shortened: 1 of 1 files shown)


ShailChoksi:
This PR removes a lot of existing functionality (e.g. correspondence games) and has a lot of hardcoded values.

TG-KRISH:
Ok!

PR #363, opened 2021-06-26
Wrong directory doesnt actually display directory

eansengchang:


icecream17 (approved):
I would use an f-string, but I know nothing about what's pythonic so take this with a grain of salt.

MarkZH:
I like f-strings since they are usually more compact and there's no need to figure out which brackets go with which `.format()` argument. All new code I write uses f-strings, so they are slowly taking over the code base.  I've tried making a branch in order to convert all `.format()` strings to f-strings, but it seemed too easy to mess up with almost no benefit.

AttackingOrDefending (approved):

MarkZH (approved):

icecream17 (approved):

PR #375, opened 2021-07-26
Bots can send greetings at start/end of game

### LINKED ISSUE #373: welcoing messages
when the bots play the game at start  there shoyuld be welcoming messages like hello good luck
and at end of game good game!

MarkZH:
Another configuration is added to config.yml.default that
gives the user the option of having their bot send a chat
message before its first move and after the game ends.

Fixes #373


eduherminio:
Does this require any extra permissions in the token to send chat messages, or that's just the case for regular messages?

AttackingOrDefending:
No it doesn't. lichess-bot was already sending messages when responding to !eval, !name, etc.

If the strings are empty for welcome message, does it raise 200 error? I think that it will continue working but it will print it to the console.

Also, is there a way to insert the players name? I think you can use `hello = eval('f"' + hello.replace('"', '\"') + '"')` so the string will be tuned into an f-string and then replace all the parameters in the string. With this, I think that if someone's config.yml has `hello: "Hi @{game.opponent.name}"` it will be turned to `Hi @AttackingOrDefending`, because hello would be `Hi @{game.opponent.name}`, after `'f"' + hello.replace('"', '\"') + '"'` it would turn to `f"Hi @{game.opponent.name}"` and by using eval it would become `Hi @AttackingOrDefending`. The `.replace('"', '\"')` is there so eval knows that it should see it as a character and not as an end of string.

I think that having an empty string (so `hello: # Message to send to chat at the start of a game` will become `hello: ""  # Message to send to chat at the start of a game`) or a default message (e.g. `Hi`) will be better because it will make it easier for people to understand where they should write the message.

MarkZH:
@eduherminio I've tried this out with my own bot and sending messages works.

@AttackingOrDefending If the string is empty, then nothing happens (see the `send_reply()` method in `conversation.py`).

I think it should be possible to add parameters to the string since `str.format()` can take keyword parameters. So, if the `hello` string is `"Hello, {opponent}! Good luck!"`, then the code in `lichess-bot.py` could be

```
conversation.send_reply("player", hello.format(opponent=game.opponent.name))
```

I'll try this out later.

MarkZH:
Now the bot can send messages that include the bot's name and the opponent's name. Other substitution field suggestions welcome.

AttackingOrDefending:
I think sending the engine name will also be good. If we can't get the engine name the same way we do it for !name in conversation.py then we can do something like [here](https://github.com/CSSLab/lichess-bot/blob/master/lichess-bot.py#L164).

Also, can someone put many hello messages (because there is a character limit)? We can probably do something like what we did with the multiple opening books.

MarkZH:
I think I want to discourage complicated chat messages. I want to keep the code and the configuration file simple. Also, there is already the `!name` command to give detailed engine information. I was hesitant to add substitution keywords because it invites more complication. I think the most common use for this feature will be to say "gg" at the end of a game.

AttackingOrDefending:
I think we should add a default hello message telling people to use !commands (or !help) to see all the commands because we expect people to know that they should type !commands but there is no way for them to know that apart from looking at the lichess-bot repo.

MarkZH:
Something like this?

`Hi! I'm {me}. Good luck! Type !help for a list of commands I can respond to.`

barca-cloud:
ok! let em try to tell this to ovasis

barca-cloud:
btw where can we do this thing actually i have to write so can you send it if you can pls

barca-cloud (approved):
very good

TG-KRISH (approved):

ShailChoksi (approved):

TG-KRISH (approved):
well done!

OIVAS7572 (approved):

TG-KRISH:
@OIVAS7572 can i know what is your lichess account?

OIVAS7572:
@TG-KRISH 
My bot account banned by Lichess . It was FOCAL7572 and Nxqe4.

TG-KRISH:
> @TG-KRISH
> My bot account banned by Lichess . It was FOCAL7572 and Nxqe4.
> My human account is SAVIO-PRINCE.

hey I need to ask something to you so will you message **[ME](Https://lichess.org/inbox/Enticingseal)**

TG-KRISH (commented):
Man, itz working

Issue #314, opened 2021-04-30
Syzygy/Gaviota Tablebases probing


TheYoBots:
It would great if somehow it was possible to make use of python-chess’s feature to probe syzygy/gaviota tablebases (like how opening book was probed). This can get really handy when Engines support syzygy and gaviota tablebases can be used or vice versa and even support syzygy and gaviota probing for engines that don’t support it.

https://python-chess.readthedocs.io/en/latest/gaviota.html
https://python-chess.readthedocs.io/en/latest/syzygy.html

MarkZH:
@TheYoBots Can you test out PR #533 that implements these tablebase lookups?

TheYoBots:
Sure I'll have a look at it and comment.

PR #382, opened 2021-08-04
Add ChessDB opening book and online endgame TBs

AttackingOrDefending:
A lot of engines have been using the ChessDB opening book (e.g. [BrainLearn](https://github.com/amchess/BrainLearn)) and I recently saw a chess engine ([weiss](https://github.com/TerjeKir/weiss)) use the online TBs. This will be like what was mentioned in #314 but without having to download the files.

We can also get EGTBs from chessdb.cn and not only from lichess but I am not sure if it is needed because lichess also has atomic and antichess but I added code also for chessdb because it may be faster for someone to get the move from chessdb instead of lichess.

I am not sure if the best place for them was in the engines section in config.yml but I put them there because the polyglot opening book section is also there. chessdb.cn has a 100000 rate limit per day (1.15 per second), so I don't think that anyone is going to hit that limit unless they are playing 0.5+0 for a whole day.

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -438,7 +438,7 @@ def get_online_egtb_move(li, board, game, online_egtb_cfg):
                     move = random_move["uci"]
                     wdl = random_move["wdl"] * -1
                     dtz = random_move["dtz"] * -1
-                    dtm = data["moves"][0]["dtm"]
+                    dtm = random_move["dtm"]
                     if dtm:
                         dtm *= -1
                 if wdl is not None:
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -14,6 +14,7 @@ import signal
 import time
 import backoff
 import sys
+import random
 from config import load_config
 from conversation import Conversation, ChatLine
 from functools import partial
@@ -241,6 +242,8 @@ def play_game(li, game_id, control_queue, engine_factory, user_profile, config, 
     move_overhead = config.get("move_overhead", 1000)
     delay_seconds = config.get("rate_limiting_delay", 0)/1000
     polyglot_cfg = engine_cfg.get("polyglot", {})
+    chessdb_cfg = engine_cfg.get("chessdb_book", {})
+    online_egtb_cfg = engine_cfg.get("online_egtb", {})
 
     first_move = True
     correspondence_disconnect_time = 0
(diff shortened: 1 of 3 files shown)


AttackingOrDefending:
Added also lichess cloud eval. Should I make a new function, so that it will only take two lines in play_game?
```python
def get_online_move(li, board, game, online_cfg):
    online_egtb_cfg = online_cfg.get("online_egtb", {})
    chessdb_cfg = online_cfg.get("chessdb_book", {})
    lichess_cloud_cfg = online_cfg.get("lichess_cloud_analysis", {})
    best_move = get_online_egtb_move(li, board, game, online_egtb_cfg)
    if best_move is None:
        best_move = get_chessdb_move(li, board, game, chessdb_cfg)
    if best_move is None:
        best_move = get_lichess_cloud_move(li, board, game, lichess_cloud_cfg)
    return best_move
```
and play_game will change to:
```python
if best_move is None:
    best_move = get_online_move(li, board, game, online_cfg)
```
I also had to change api_get because get_lichess_cloud_move will raise an error (doesn't crash but it prints it to the console) if it can't find a move for the position because of `response.raise_for_status()`.

lichess.py
```diff
@@ -41,10 +41,11 @@ def is_final(exception):
                           max_time=60,
                           interval=0.1,
                           giveup=is_final)
-    def api_get(self, path):
+    def api_get(self, path, raise_for_status=True):
```
eduherminio, line 44: Unrelated to this PR, but would it be worth adding `timeout` as an additional, optional param. as well?

lichess-bot.py
```diff
@@ -362,6 +372,159 @@ def get_book_move(board, polyglot_cfg):
...
+            if data["status"] == "ok":
+                move = data["move"]
+                logger.info("Got move {} from chessdb.cn".format(move))
+
+        elif quality == "all":
+            data = li.api_get(f"https://www.chessdb.cn/cdb.php?action=query&board={board.fen()}&json=1")
+            if data["status"] == "ok":
+                move = data["move"]
+                logger.info("Got move {} from chessdb.cn".format(move))
+    except:
```
MarkZH, line 404: Every `except:` should be changed to `except Exception:` to prevent `ctrl-C` from being ignored.

MarkZH:
My only concern would be to check with each of the online services whether they are able to handle a sudden influx of queries. This PR makes it easy to turn on these queries with a couple of changes to the config.yml file. I would ask the owners of these databases whether they are prepared for however many bot users there are on lichess to all take advantage of this. What seems like reasonable limits can easily grow unwieldy with distributed automation.

AttackingOrDefending:
chessdb.cn can probably handle it since it is already used by a lot of engines and it has a very high rate limit. lichess can also probably handle it since a lot of people already use these APIs from their browser. Most of the time, there are less than 200 bots online, most of them aren't playing a game and a lot of them won't use it because it doesn't fit their bot (e.g. it won't make sense for a bot that plays random moves to use it).

MarkZH:
I think the `get_online_move()` function is a good idea.

ShailChoksi:
@MarkZH can you resolve the merge conflict?

MarkZH:
@ShailChoksi This isn't my PR.

cc @AttackingOrDefending

PR #327, opened 2021-05-13
Allow XBoard bots of offer/accept draws and resign

MarkZH:
I'm submitting this PR as a draft since I'm not confident this works adequately yet. Critiques and comments are welcome.

These changes allow XBoard engines to resign games (by printing "resign" to stdout) and to offer/accept draws (by printing "offer draw" to stdout). There are no similar functions in UCI protocol.

Two problems so far:

1. The python-chess library will only pick up on "offer draw" if the bot prints it while thinking about a move. Otherwise, it is ignored. So, if the bot responds immediatlely to the "draw" command on line 262 before one of the choose_move functions can be called, it will be ignored.

2. When accepting a draw, the lichess API returns with

```
https://lichess.org:443 "POST /api/bot/game/za8D4j9z/move/c1f4?offeringDraw=1 HTTP/1.1" 400 43
Giving up api_post(...) after 1 tries (requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://lichess.org/api/bot/game/za8D4j9z/move/c1f4?offeringDraw=1)
```

The draw acceptance is properly accepted by lichess and the game actually ends with a draw. I'm not sure if the error message is
important.

PR #385, opened 2021-08-19
PlayResult warning

icecream17:
Somewhere in the code (I forgot where) doesn't work if an instance of `chess.engine.PlayResult` _isn't_ returned

There's probably a better solution for this


MarkZH:
This should be merged at the same time as #327, right? If so, this PR looks correct.

ShailChoksi:
@icecream17 Can you resolve the conflict

PR #391, opened 2021-09-23
Fix bug in draw offers due to lichess API change

MarkZH:
While fixing the merge conflicts in PR #375, I discovered a bug in #327. Apparently, there's been a change to the lichess API where draw offers (`bdraw` and `wdraw` fields in game state) are not sent on the first move. This PR supplies the fix.

The same commit is also present in PR #375, but I thought it was important to fix this bug even if the greeting PR had to be delayed.


MarkZH:
lichess-bot is unusable until this pull request is merged. It will close #393.

PR #396, opened 2021-10-06
Mention repo link in !howto command

SriMethan:


TheYoBots:
Lichess doesn't allow bots to send even github.com/ShailChoksi/lichess-bot, and if you break the link for it to be sent, it would be like cheating lichess. There is a reason lichess doesn't permit bots to send links in chat, we must respect that.

PR #404, opened 2021-12-15
Test updates by playing a game offline

AttackingOrDefending:
Like #321 but offline. I had mentioned it [here](https://github.com/ShailChoksi/lichess-bot/pull/321#issuecomment-932723186) but I decided to make a new PR because it looks clearer and it is more easily accessed.
It is a simple environment that tries to emulate lichess, so there is no need for a token.

Pros:
- No token needed
- Can run offline

Cons:
- Doesn't test the functions in lichess.py because it replaces them
- It hasn't implemented every function in lichess.py (e.g. chat)

Note: If chat is implemented, we can add the script to send a chat command (e.g. eval) to the engine, so we can test the engine's response to commands. This can be done also for other functions.


test.py
```diff
@@ -0,0 +1,146 @@
...
+import chess.engine
+import threading
+import random
+from shutil import copyfile
+import importlib
+copyfile('./test/lichess.py', 'lichess.py')
+lichess_bot = importlib.import_module("lichess-bot")
+
+
+def test_nothing():
```
MarkZH, line 17: I think it's safe to delete this "test."

MarkZH:
I have some suggestions.

- Add `python-timeout` to `requirements.txt`
- Rename `test.py` to `test_bot.py` or similar. Running `pytest` with no arguments only collects files of the form `test_*.py`.

The following changes may not matter when tests are run on Github servers, but tests can also be useful for developers to run locally. To that end, here are some more suggestions:

- Shorten the time of the game played. When I run the test, it takes more than three minutes. Something like 5 seconds with 0.1 second increment should be plenty for a purely local game.
- In `test/lichess.py`, delete the functions and methods for online communicate (and other code that becomes dead as a result). Deleting all of those will make it easier to maintain and easier to understand since all of the code will be offline.
- Clean up the files that get created by the tests. The easiest way is to create a temporary directory, run the tests there, then leave the test area and delete it.

MarkZH:
I'm trying to test out these changes, but I'm having trouble installing `pathlib` from requirements. I get the following error message. Any ideas?

```
Collecting pathlib==1.0.1
  Using cached pathlib-1.0.1.tar.gz (49 kB)
  Preparing metadata (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [1 lines of output]
      ERROR: Can not execute `setup.py` since setuptools is not available in the build environment.
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
```

MarkZH:
Also, I see you preserve the original `lichess.py`
```python
shutil.copyfile('lichess.py', 'correct_lichess.py')
shutil.copyfile('./test/lichess.py', 'lichess.py')
```
But I don't see where the original is put back. Maybe something like this would work:
```python
if __name__ == '__main__':
    try:
        test_bot()
    finally:
        shutil.move('correct_lichess.py', 'lichess.py')
```

AttackingOrDefending:
Can you test both 2367d32 and 6d5d0bf?

MarkZH:
I can't test this PR because of [the error above](https://github.com/ShailChoksi/lichess-bot/pull/404#issuecomment-1025353363). Even when I delete `pathlib` from `requirements.txt`, it still tries to install it and it fails. 
I'm assuming this is because pathlib is a dependency of something else on the list. I've tried on my linux and windows machines and get the same error.

MarkZH:
I can't test this PR as is because the installation of `pathlib` always fails with the error in [my previous comment](https://github.com/ShailChoksi/lichess-bot/pull/404#issuecomment-1025353363). `pathlib` is a dependency of `wrapt-timeout-decorator` so I can't just delete it from the list. I've tried on both Windows and Linux and the same error occurs.

I was able to get the test running by:

1. Deleting `pathlib` and `wrapt-timeout-decorator` from `requirements.txt`
2. Deleting the corresponding `import` line from `test_bot.py` on line 7.
3. Replacing `@timeout` with `@pytest.mark.timeout` in line 50 of `test_bot.py`

Both time when I ran the test, it failed due to running more than 150 seconds.

MarkZH:
Then again, since the github tests are running fine, I must be doing something wrong.

MarkZH:
I was running Python 3.10.2. Downgrading to 3.9.9 allowed me to install everything and successfully run the tests. So, I think this PR is incompatible with Python 3.10. Try adding 3.10 to the workflows in `.github` and see what happens.

AttackingOrDefending:
pytest.mark.timeout only works at the global level, not inside a function, which is why I added wrapt-event-timeout. I removed it now and fixed a bug for Linux in test-bot.py. In GitHub, Linux passes as well, can you test it?

MarkZH:
Pychess ([homepage](https://pychess.github.io/), [source](https://github.com/pychess/pychess)) can run in Xboard or UCI mode.

The tests seem to be working now with one exception. The function `pytest_sessionfinish()` does not run. That function needs to be placed in its own file called `conftest.py` in order for pytest to run it.

TheYoBots:
Can the `test_bot.py` file also be added to the `test/` dir? That way I feel it would be more organized and all test files will be seperated from the main files.

AttackingOrDefending:
@MarkZH Testing SF on linux fails when I add `CONFIG['engine']['working_dir'] = './TEMP/'`. In windows it works fine. I added it for Sjeng because it makes some files when it plays, so it is not needed for SF so I removed it in 576cc2d and it works fine. I don't know if it has anything to do with your PR or if I am doing something wrong but I can't test it because I am using windows.

MarkZH:
The latest changes work on both my Linux and Windows computers--including cleanup. My one suggestion is to change the `if __name__ == '__main__':` part so that it doesn't run any tests. Instead, it should just print a message saying that this file should only be run with pytest. This is because the cleanup code doesn't run without pytest. Something like
```python
if __name__ == '__main__':
    sys.exit(f'The script {os.path.basename(__file__)} should only be run by pytest.')
```
This should be placed near the top of the file above the `shutil.copyfile` statements on lines 15 and 16. This way, no files are changed or downloaded, and the user won't be left with a broken lichess-bot.

MarkZH:
Anything else before I approve and merge?

TheYoBots:
> Anything else before I approve and merge?

I really think the `test_bot.py` file should also be in the `test_bot/` directory.

AttackingOrDefending:
I moved conftest to test_bot/ but I wasn't successful in moving test_bot.py to test_bot/. I tried replacing the importlib line with:
```python
loader = importlib.machinery.SourceFileLoader('lichess-bot', os.path.join(os.path.dirname(os.path.dirname(__file__)), 'lichess-bot.py'))
spec = importlib.util.spec_from_loader(loader.name, loader, origin=os.path.dirname(os.path.dirname(__file__)))
lichess_bot = importlib.util.module_from_spec(spec)
spec.loader.exec_module(lichess_bot)
```
but when loading lichess-bot it can't file engine_wrapper. Probably it expects engine_wrapper to be in test_bot/. I can use copyfile to copy all the files to test_bot/ and then delete them, but I personally prefer the way it is now than copying all the files to test_bot/.

MarkZH:
I think only two things need to change to put all the tests in one folder:

- Move `test_bot.py` to `test_bot/`
- Put an empty `__init__.py` in `test_bot/`

No changes to the files necessary. This allows pytest to find all the tests and run. I am getting weird timeout failures with the stockfish test on Linux. Still investigating.

MarkZH:
Annoyingly, the timeouts are intermittent. The tests pass on Windows and most of the time on Linux.

MarkZH:
OK. This looks good. @AttackingOrDefending, @TheYoBots, anything else?

AttackingOrDefending:
It is ready from me. I run the checks 7 times on github and all of them passed.

requirements.txt
```diff
@@ -6,3 +6,6 @@ PyYAML>=5.4.1
 requests>=2.25.1
 urllib3==1.26.8
 backoff==1.11.1
+
```
TheYoBots, line 9: Can a comment be added here saying something like `# for tests`?

MarkZH (approved):

MarkZH:
With that last commit, Linux tests pass every time. Go ahead and squash merge. If there are more changes that are just cosmetic, feel free to merge after those.

PR #410, opened 2022-01-09
Fix Intro message

TheYoBots:
The intro message never gets sent in a game, but the message at the end of the game gets sent. This commit fixes that by placing the line to send the intro message (`conversation.send_message("player", hello)`) at the start of the game.

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -250,7 +250,7 @@ def play_game(li, game_id, control_queue, engine_factory, user_profile, config, 
     get_greeting = lambda greeting: str(greeting_cfg.get(greeting, "") or "").format_map(keyword_map)
     hello = get_greeting("hello")
     goodbye = get_greeting("goodbye")
-    
+
     first_move = True
     correspondence_disconnect_time = 0
     while not terminated:
@@ -269,6 +269,8 @@ def play_game(li, game_id, control_queue, engine_factory, user_profile, config, 
                 game.state = upd
                 board = setup_board(game)
                 if not is_game_over(game) and is_engine_move(game, board):
+                    if len(board.move_stack) < 2:
+                        conversation.send_message("player", hello)
                     start_time = time.perf_counter_ns()
                     fake_thinking(config, board, game)
                     print_move_number(board)
@@ -282,7 +284,6 @@ def play_game(li, game_id, control_queue, engine_factory, user_profile, config, 
                         draw_offered = check_for_draw_offer(game)
 
                         if len(board.move_stack) < 2:
-                            conversation.send_message("player", hello)
                             best_move = choose_first_move(engine, board, draw_offered)
                         elif is_correspondence:
                             best_move = choose_move_time(engine, board, correspondence_move_time, can_ponder, draw_offered)


MarkZH:
I just tried a game, and the opening greeting message does get sent during the first move. I put the code line where it was so that the message would only be sent once per game. With this change, the greeting will be sent after every disconnection. During correspondence games, this would result in the greeting being sent on every move.

TheYoBots:
I tried to run lichess bot with no changes and at the start of each game a welcome message wasn't sent. I tested it for a few other games and it still didn't work. The code changes I made in this PR fixes that.

MarkZH:
I see. Your bot uses an opening book or the online databases, right? That would cause the sending of the greeting to be skipped. This change is correct.

Please add `if len(board.move_stack) < 2:` above the greeting line so that the greeting is only sent once per game.

TheYoBots:
> I see. Your bot uses an opening book or the online databases, right? That would cause the sending of the greeting to be skipped. This change is correct.
> 
> A better place to move the line is under `if not is_game_over(game) and is_engine_move(game, board):` (line 271) with `if len(board.move_stack) < 2:` so that the greeting is only sent once per game.

Yes my Bot does use an opening book and that causes the greeting to be skipped. Without an opening book the greeting message is sent.

Keeping the line right under `if not is_game_over(game) and is_engine_move(game, board):` will send the opening message every move and if placed with `if len(board.move_stack) < 2:` then it is in the exact same place it was before this PR (i.e, the error still occurs).

Placing it right after [`if first_move:` in line 258](https://github.com/ShailChoksi/lichess-bot/blob/3d0582fc4ffea2a62df633066a5b60e09556070f/lichess-bot.py#L258) should fix this and send the message only once every game.

MarkZH:
This new change using `first_move` will send the greeting every time the bot disconnects and reconnects (network errors or every move of a correspondence game). My suggestion for the new placement would be something like this starting at line 271:

```python
if not is_game_over(game) and is_engine_move(game, board):
    if len(board.move_stack) < 2:
        conversation.send_message("player", hello)
    start_time = time.perf_counter_ns()
    fake_thinking(config, board, game)
    # etc.
```
This will definitely print the message and only once per game.

TheYoBots:
> This new change using `first_move` will send the greeting every time the bot disconnects and reconnects (network errors or every move of a correspondence game). My suggestion for the new placement would be something like this starting at line 271:
> 
> ```python
> if not is_game_over(game) and is_engine_move(game, board):
>     if len(board.move_stack) < 2:
>         conversation.send_message("player", hello)
>     start_time = time.perf_counter_ns()
>     fake_thinking(config, board, game)
>     # etc.
> ```
> 
> This will definitely print the message and only once per game.

A message is sent once every move. https://lichess.org/ZQLNeqGb

MarkZH:
Now I'm confused. I just tried the quoted code and the message is only sent once. https://lichess.org/WJIoBtM4Daex

TheYoBots:
> Now I'm confused. I just tried the quoted code and the message is only sent once in my bot's games. https://lichess.org/WJIoBtM4Daex

I can't see anything in the chat room because the code only allows the bot to send a message in the players room and only players who played the game can see that. So please do add the spectators chat room (Only for tests to show others).

Sorry I had tested this instead of what you had sent.
```py
                if not is_game_over(game) and is_engine_move(game, board):
                    conversation.send_message("player", hello)
                    start_time = time.perf_counter_ns()
                    fake_thinking(config, board, game)
                    print_move_number(board)
                    correspondence_disconnect_time = correspondence_cfg.get("disconnect_time", 300)
```
Which should explain the error in the game I sent.

MarkZH:
> I can't see anything in the chat room because the code only allows the bot to send a message in the players room and only players who played the game can see that. So please do add the spectators chat room (Only for tests to show others).

Thanks. I didn't realize that the messages weren't visible.

> Works https://lichess.org/JJ68kMyl
> 
> Thanks!

Great! I like it. One last cosmetic suggestion to clean up the diff: the newline at the end of the last line in the file got deleted. That's why there's a ⛔-looking symbol at the end of the commit diff.

TheYoBots:
> One last cosmetic suggestion to clean up the diff: the newline at the end of the last line in the file got deleted. That's why there's a ⛔-looking symbol at the end of the commit diff.

I don't think that effects how the file is run right?

MarkZH:
Right. There's no functional effect. It's just cosmetic.

MarkZH (approved):

AttackingOrDefending (approved):

PR #413, opened 2022-01-11
Resign or offer/accept draw based on score

AttackingOrDefending:
The bot can resign if the score falls very low or accept/offer draw if the score stays close to 0 for a few moves.

resign_enabled / offer_draw_enabled: Enable resigning or offering/accepting draws
resign_score: If the score is less than or equal to this value, the bot resigns (in cp)
offer_draw_score: If the absolute value of the score is less than or equal to this value, the bot offers/accepts draw (in cp) (if offer_draw_score=10, draw values will be from -10 to 10 cp)
resign_moves / offer_draw_moves: How many moves in a row the score has to be below the value, to resign or offer/accept draw
offer_draw_pieces: Maximum number of pieces the board should have offer/accept draw

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -13,6 +13,7 @@ def create_engine(config):
     engine_path = os.path.join(cfg["dir"], cfg["name"])
     engine_type = cfg.get("protocol")
     engine_options = cfg.get("engine_options")
+    draw_or_resign = cfg.get("draw_or_resign", {})
     commands = [engine_path]
     if engine_options:
         for k, v in engine_options.items():
@@ -30,7 +31,7 @@ def create_engine(config):
         raise ValueError(
             f"    Invalid engine type: {engine_type}. Expected xboard, uci, or homemade.")
     options = remove_managed_options(cfg.get(engine_type + "_options", {}) or {})
-    return Engine(commands, options, stderr)
+    return Engine(commands, options, stderr, draw_or_resign)
 
 
 def remove_managed_options(config):
@@ -56,8 +57,9 @@ class GameEnding:
 
 
 class EngineWrapper:
-    def __init__(self, commands, options, stderr):
-        pass
+    def __init__(self, commands, options, stderr, draw_or_resign):
+        self.scores = []
+        self.draw_or_resign = draw_or_resign
 
     def search_for(self, board, movetime, ponder, draw_offered):
         return self.search(board, chess.engine.Limit(time=movetime // 1000), ponder, draw_offered)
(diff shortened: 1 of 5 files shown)


MarkZH:
Some comments:

- After a draw offer, if a game's score stays under the draw score, then a draw offer will be sent on every move. This will greatly annoy human players and will probably result in disciplinary action from lichess. You need some way of limiting how often draw offers are sent in one game.
- `make_move()` in lichess.py does not need to be modified since resignations are already handled in [lichess-bot.py in lines  291 and 292](https://github.com/ShailChoksi/lichess-bot/blob/3d0582fc4ffea2a62df633066a5b60e09556070f/lichess-bot.py#L291). Adding draw and resign options to the chosen move result is enough.
- In `offer_draw_or_resign()`, split up the conditional expressions into variables to make them easier to interpret:

For example, instead of
```python
if len(scores) == len(list(filter(lambda score: abs(score.relative.score(mate_score=40000)) <= self.draw_or_resign.get('offer_draw_score', 10), scores))) and len([board.piece_type_at(sq) for sq in chess.SQUARES if board.piece_type_at(sq)]) <= self.draw_or_resign.get('offer_draw_pieces', 10):
                result.draw_offered = True
```
Try
```python
pieces_on_board = len([board.piece_type_at(sq) for sq in chess.SQUARES if board.piece_type_at(sq)])
scores_near_draw = lambda score: abs(score.relative.score(mate_score=40000)) <= self.draw_or_resign.get('offer_draw_score', 10)
if len(scores) == len(list(filter(scores_near_draw, scores))) and pieces_on_board <= self.draw_or_resign.get('offer_draw_pieces', 10):
    result.draw_offered = True
```

ShailChoksi:
@MarkZH / @AttackingOrDefending would one or both of you be interested in becoming a colloborator for this repo? I can't give this project enough time and it seems that you two have been putting in a lot of effort to improve this. I will have to confirm it with @ornicar but I think it will be for the better

AttackingOrDefending:
Lichess doesn't let you send a draw offer every move so it will not offer draw every move (I think it allows about once every 10 moves). We have to send the draw offer every move, because offering draws also works as accepting draws.

Thanks for the offer. I will be happy to become a collaborator.

MarkZH:
@ShailChoksi I would also be happy to be a collaborator.

ShailChoksi:
@MarkZH @AttackingOrDefending you both have collaborator access.

AttackingOrDefending:
This PR only works for moves played by the engine, so 2 moves in the offer_draw_moves parameter may not actually be the last 2 game moves. It is the last 2 engine moves.

MarkZH:
I think it would be good to mention in the documentation that these configurations are for bots that don't already handle draws and resignations on their own.

After that and fixing the merge conflicts, I'll test and merge.

MarkZH:
I think it would be good to mention in the documentation that these configurations are for bots that don't already handle draws and resignations on their own.

engine_wrapper.py
```diff
@@ -14,6 +14,7 @@ def create_engine(config):
     engine_path = os.path.join(cfg["dir"], cfg["name"])
     engine_type = cfg.get("protocol")
     engine_options = cfg.get("engine_options")
+    draw_or_resign = cfg.get("draw_or_resign", {})
```
MarkZH, line 17: This should be written this way: `draw_or_resign = cfg.get("draw_or_resign") or {}`. If the user's config.yml has the following:

    draw_or_resign:
      # resign_enabled: False
      # ... etc.

Then `cfg.get("draw_or_resign", {})` will return `None` since the key `draw_or_resign` is present. Since dict methods are called on `draw_or_resign` later, these will causes errors if that variable is not a dict.

MarkZH (approved):
Is this ready to merge?

AttackingOrDefending:
> Is this ready to merge?

Yes.

Torom:
Hi, I added the same function to my bot and encountered an issue. I don't see how you guys catch this case, so I'm sharing this here:

You add PovScores or floats to `self.scores` ([engine_wrapper.py#L110](https://github.com/AttackingOrDefending/lichess-bot/blob/986062bdba1389bd255176fe725110e829d92992/engine_wrapper.py#L110)), but then you access the values as if they were definitely PovScores ([engine_wrapper.py#L96](https://github.com/AttackingOrDefending/lichess-bot/blob/986062bdba1389bd255176fe725110e829d92992/engine_wrapper.py#L96)). I run in a `AttributeError: 'float' object has no attribute 'relative'` and I don't see how you catch this case. This happens when the engine has no evaluation for a move, this happens especially in the absolute endgame. You add a NaN and then treat it like a PovScore. Maybe I'm missing why this is not a problem.

Issue #232, opened 2020-11-11
Setting current directory of engine executable


cosinekitty:
This is an awesome project. I was able to get my own chess engine [Chenard](https://github.com/cosinekitty/chenard) up and running on linode.org in less than 10 minutes.

The xboard version of my engine reads and writes files from whatever the current directory is. When I run lichess.py from its directory (a clone of this repo), it apparently spawns Chenard with that same directory as the current directory. This causes Chenard's private files to get mixed in with lichess-bot files, which is a little messy.

Is there a way to specify an alternate directory to set as the current directory for the engine's spawned process?  If not, would you be open to me working on a pull request to do that?

It's no big deal because I could just add that feature to my own code via an environment variable telling Chenard where to read/write its private files to. I thought other engines might benefit from this too.

MarkZH:
What's worked for me is to create a script that changes to the desired directory before calling the actual chess program. So, in the file `lichess-bot/engines/my_chess_bot.sh`:

    cd /dir/for/writing/files/
    the_actual_chess_program --option1=value1 # etc.

Then, in `lichess-bot/config.yml`:

    # ...
    engine:
      dir: "./engines/"
      name: "my_chess_bot.sh
    # ...

MarkZH:
@cosinekitty I've created a pull request to implement this directly (#416). Please test if you can.

MarkZH:
Users can now specify a working directory for their engines (#416).

PR #416, opened 2022-01-16
Separate working directory

MarkZH:
This PR lets the user specify which directory the bot should run from. This is useful where the bot needs to read files or writes files and the user does not want them in the lichess directory.

If the new working directory configuration is blank or missing, the current working directory is used, which I believe is the current behavior. This change should be backward compatible with older config files.

This implements the request in issue #232.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,7 +269,7 @@ The examples just implement `search`.
 ## Tips & Tricks
 - You can specify a different config file with the `--config` argument.
 - Here's an example systemd service definition:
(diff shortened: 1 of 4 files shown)


AttackingOrDefending (approved):

MarkZH:
Since this is a new feature, I'll wait on merging this--both to see if I spot any other needed changes and the off chance that the initial issue reporter sees this PR.

PR #419, opened 2022-01-17
Fix ponderpv bug

### LINKED ISSUE #417: Bot is not responding to eval command if it is sent during the opponent's move.
On [both](https://lichess.org/@/AMemeBot/all) of my [bots](https://lichess.org/@/IUseStockfish), the eval command is not being recognized during my opponent's move after the recent bot changes from the last few days.

Steps to reproduce:
1. Start a game with any player.
2. Run the !eval command in spectator chat during the opponent's turn.
3. No response will be given.
4. Run the !eval command during your bot's turn.
5. You should get a response this time.

Intended behavior:
Bot should always respond to !eval, not only during your turn.

AttackingOrDefending:
ponderpv is now made immediately after the search and we only remove the moves when we want to send them in the chat.
closes #417

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -110,14 +110,14 @@ class EngineWrapper:
         self.scores.append(self.last_move_info.get("score", float('nan')))
         result = self.offer_draw_or_resign(result, board)
         self.last_move_info["ponderpv"] = board.variation_san(self.last_move_info.get("pv", []))
-        self.print_stats(board)
+        self.print_stats()
         return result
 
-    def print_stats(self, board):
-        for line in self.get_stats(board):
+    def print_stats(self):
+        for line in self.get_stats():
             logger.info(f"{line}")
 
-    def get_stats(self, board, for_chat=False):
+    def get_stats(self, for_chat=False):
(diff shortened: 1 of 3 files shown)


engine_wrapper.py
```diff
@@ -118,13 +119,11 @@ def print_stats(self, board):
 
     def get_stats(self, board, for_chat=False):
         info = self.last_move_info.copy()
-        if "pv" not in info:
-            info["pv"] = []
         if for_chat:
             stats = ["depth", "nps", "nodes", "score", "ponderpv"]
```
MarkZH, line 123: The `stats` variable is the same in both branches of `if for_chat:`. Put this line above the `if` statement and you can delete the `else` branch.

AttackingOrDefending: Done. It is not ready to be merged. I will change a few more things (e.g. board in get_stats is not used).

MarkZH (approved):

AttackingOrDefending:
It is ready. The problem was the board parameter. The board was updated with the move the bot played but the pv still contained that move, so python-chess thought that it was an illegal move.

PR #420, opened 2022-01-18
Fix bug with calling relative on float

### LINKED ISSUE #418: Bot is randomly stopping for a second in mate positions
image
Bot is stopping play for a second every so often in mate positions, unsure of the reason.

```2022-01-17 11:09:10,289:   move: 108
2022-01-17 11:09:10,291:   Searching for wtime 16090 btime 6009
2022-01-17 11:09:10,292:   Backing off play_game(...) for 0.3s (AttributeError: 'float' object has no attribute 'relative')

Related: My engine seems to be restarting often throughout the game for no apparent reason. I can attach logs if needed.

AttackingOrDefending:
`float('nan')` was replaced by `False` and a test for `score is not False` was added.
closes #418


MarkZH:
A better default value than `False` is `chess.engine.PovScore(chess.engine.Mate(1), board.turn)`. This way, the scores list always contains `PovScore`s and there no need to check for `False`. Plus, a mate in one is higher than any other chess score, so it won't count towards draws or resignations.

MarkZH (approved):

PR #422, opened 2022-01-21
Offer draw or resign based on online egtb score

### LINKED ISSUE #421: `offer_draw` doesn't work with `online_egtb`
Hi,
first of all, I really like the offer draw feature. I think you should release a new version and enable it by default so that as many as possible use the function.

However, it has no function when `online_egtb` is enabled. I think also egtb moves that are known to be draws should be counted as 0.00 moves, so that the bot can still offer and accept draws.

AttackingOrDefending:
`resign_moves`, `offer_draw_moves` and `offer_draw_pieces` are not taken into account when offering/accepting draw or resigning based on the online egtb score.
In lichess-bot.py line 293 I added `and len(board.move_stack) >= 2` because if the bot resigns on the first move, the game is instead aborted.
closes #421

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,8 +67,8 @@ Besides the above, there are many possible options within `config.yml` for confi
     - `resign_moves`: The evaluation has to be less than or equal to `resign_score` for `resign_moves` amount of moves for the bot to resign.
     - `offer_draw_enabled`: Whether the bot is allowed to offer/accept draw based on the evaluation.
     - `offer_draw_score`: The absolute value of the engine evaluation has to be less than or equal to `offer_draw_score` for the bot to offer/accept draw.
-    - `offer_draw_egtb_score`: The absolute value of the wdl score from the online egtb has to be less than or equal to `offer_draw_score` for the bot to offer/accept draw.
-    - `offer_draw_moves`: The absolute value of the evaluation has to be less than or equal to `offer_draw_score` for `offer_draw_egtb_score` amount of moves for the bot to offer/accept draw.
+    - `offer_draw_egtb_score`: The absolute value of the wdl score from the online egtb has to be less than or equal to `offer_draw_egtb_score` for the bot to offer/accept draw.
+    - `offer_draw_moves`: The absolute value of the evaluation has to be less than or equal to `offer_draw_score` for `offer_draw_moves` amount of moves for the bot to offer/accept draw.
     - `offer_draw_pieces`: The bot only offers/accepts draws if the position has less than or equal to `offer_draw_pieces` pieces.
 - `online_moves`: This section gives your bot access to various online resources for choosing moves like opening books and endgame tablebases. This can be a supplement or a replacement for chess databases stored on your computer. There are three sections that correspond to three different online databases:
     1. `chessdb_book`: Consults a [Chinese chess position database](https://www.chessdb.cn/), which also hosts a xiangqi database.
--- a/README.md
+++ b/README.md
@@ -67,7 +67,8 @@ Besides the above, there are many possible options within `config.yml` for confi
     - `resign_moves`: The evaluation has to be less than or equal to `resign_score` for `resign_moves` amount of moves for the bot to resign.
     - `offer_draw_enabled`: Whether the bot is allowed to offer/accept draw based on the evaluation.
     - `offer_draw_score`: The absolute value of the engine evaluation has to be less than or equal to `offer_draw_score` for the bot to offer/accept draw.
-    - `offer_draw_moves`: The absolute value of the evaluation has to be less than or equal to `offer_draw_score` for `offer_draw_moves` amount of moves for the bot to offer/accept draw.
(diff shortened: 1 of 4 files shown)


lichess-bot.py
```diff
@@ -289,7 +290,7 @@ def play_game(li, game_id, control_queue, engine_factory, user_profile, config,
                             best_move = choose_move_time(engine, board, correspondence_move_time, can_ponder, draw_offered)
                         else:
                             best_move = choose_move(engine, board, game, can_ponder, draw_offered, start_time, move_overhead)
-                    if best_move.resigned:
+                    if best_move.resigned and len(board.move_stack) >= 2:
```
MarkZH, line 293: What is the reason for this change?

AttackingOrDefending: I added `and len(board.move_stack) >= 2` because if the bot resigns on the first move, the game is instead aborted.

README.md
```diff
@@ -67,6 +67,7 @@ Besides the above, there are many possible options within `config.yml` for confi
     - `resign_moves`: The evaluation has to be less than or equal to `resign_score` for `resign_moves` amount of moves for the bot to resign.
     - `offer_draw_enabled`: Whether the bot is allowed to offer/accept draw based on the evaluation.
     - `offer_draw_score`: The absolute value of the engine evaluation has to be less than or equal to `offer_draw_score` for the bot to offer/accept draw.
+    - `offer_draw_egtb_score`: The absolute value of the wdl score from the online egtb has to be less than or equal to `offer_draw_egtb_score` for the bot to offer/accept draw.
```
MarkZH, line 70: Since wdl is only ever -2, -1, 0, 1, or 2, should this option be a true/false configuration? That is, if the egtb returns a 0 score, then offer a draw?

lichess-bot.py
```diff
@@ -532,24 +533,31 @@ def score_to_wdl(score):
...
+    offer_draw = False
+    resign = False
+    best_move, wdl = get_online_egtb_move(li, board, game, online_egtb_cfg)
     if best_move is None:
         best_move = get_chessdb_move(li, board, game, chessdb_cfg)
+    else:
+        if draw_or_resign_cfg.get('offer_draw_enabled', False) and abs(wdl) <= draw_or_resign_cfg.get('offer_draw_egtb_score', 0):
+            offer_draw = True
+        if draw_or_resign_cfg.get('resign_enabled', False) and wdl == -2:
+            resign = True
```
MarkZH, line 556: I would add a blank line after this line (before `if best_move is None:`, since it's not really a part of the next if-block structure.

lichess-bot.py
```diff
@@ -532,24 +533,31 @@ def score_to_wdl(score):
...
     if best_move is None:
         best_move = get_lichess_cloud_move(li, board, game, lichess_cloud_cfg)
     if best_move:
```
MarkZH, line 559: I would also add a blank line above `if best_move:` for the same reason. It's not a part of the same logic flow.

README.md
AttackingOrDefending: In the last commit it became true/false, but I am not sure if this is what you meant.

MarkZH: if you look at this function:

```python
def score_to_wdl(score):
    if score < -20000:
        return -2
    elif score < 0:
        return -1
    elif score == 0:
        return 0
    elif score <= 20000:
        return 1
    else:
        return 2
```

A wdl of -1 can mean anything from slight disadvantage to very-losing-but-the-route-to-checkmate-is-slightly-too-long-to-prove (and similarly for wdl of 1). This is too wide a range to be offering draws. If I were writing this, only wdl = 0 would result in a draw offer and the option would be called `offer_draw_for_egtb_zero`.

```python
if draw_or_resign_cfg.get('offer_draw_enabled', False) and draw_or_resign_cfg.get('offer_draw_for_egtb_zero', False) and wdl == 0:
    offer_draw = True
```
Or, so the line isn't so long
```python
can_offer_draw = draw_or_resign_cfg.get('offer_draw_enabled', False)
offer_egtb_draw = draw_or_resign_cfg.get('offer_draw_for_egtb_zero', False)
if can_offer_draw and offer_egtb_draw and wdl == 0:
    offer_draw = True
```
I am open to arguments that your way is better.

AttackingOrDefending: I added `offer_draw_for_egtb_zero` and `resign_for_egtb_minus_two`. Now the bot won't offer draws in positions with wdl of 1 or -1. A wdl of -1 means that the position is lost but the 50 move rule makes it a draw. A score of 20000 by the chessdb egtb isn't the same as a score of 20000 by a chess engine. The positions with wdl of 1 or -1 are converted to score by chessdb with `score = wdl * 20000 - wdl * dtz`, while in chess engines it means that the position is winning but I haven't found a mate yet. Chess engines will usually have a score of about 400 in positions with wdl of 1. If two bots have the online_egtb enabled or have the TB files on their computer, the bots will draw even though chessdb returns a score of 20000.

MarkZH (approved):

PR #435, opened 2022-02-07
Update config.yml.default

aceventurapetdetective:


MarkZH:
The change would cause everybody who uses lichess-bot to have an invalid config.yml file. The purpose of the config.yml.default file is so it can be updated with new features without breaking users' current configurations. Plus, every user has a different OAuth token; it would be bad to make everybody use the same one. If the token in this commit is the token your bot uses, I recommend going to the lichess website to delete it and generate a new one. Otherwise, everyone who sees this PR will be able to copy it and use your bot's lichess account for their own bots.

PR #437, opened 2022-02-07
Minor changes

AttackingOrDefending:
Changes made:
- A bug fix for test_bot
If SF lost, the test would still pass because `bool('0') = True`. The chances of SF losing to SF skill level 0 are very low.
- Add more debug statements for easier debuging
- Added `test_bot/__pycache__` to `.gitignore`


MarkZH:
I think a better way of handling the test win condition is to write `assert win == "1"` instead of reassignment.

Everything else is good.

MarkZH (approved):

PR #438, opened 2022-02-07
Formatting changes

AttackingOrDefending:
Changes made:
- Changed single quotes to double quotes
- Changed `.format` and `str + str` to f-strings (except for [here](https://github.com/AttackingOrDefending/lichess-bot/blob/minor-changes/engine_wrapper.py#L225) where I prefer `" {" + endgame_message + "}"` more than `f" {{{endgame_message}}}"` and when `str + str` is inside an f-string (e.g. `f"Text: {str1 + str2}"`))


lichess-bot.py
```diff
@@ -102,7 +101,7 @@ def game_logging_configurer(queue, level):
 def start(li, user_profile, engine_factory, config, logging_level, log_filename, one_game=False):
     challenge_config = config["challenge"]
     max_games = challenge_config.get("concurrency", 1)
-    logger.info("You're now connected to {} and awaiting challenges.".format(config["url"]))
+    logger.info(f'You\'re now connected to {config["url"]} and awaiting challenges.')
```
MarkZH, line 104: This change is strange. Why not use single quotes for `config['url']` and keep the outer double quotes? It gets rid of an escaping backslash.

lichess-bot.py
```diff
@@ -712,4 +711,4 @@ def intro():
         engine_factory = partial(engine_wrapper.create_engine, CONFIG)
         start(li, user_profile, engine_factory, CONFIG, logging_level, args.logfile)
     else:
-        logger.error("{} is not a bot account. Please upgrade it to a bot account!".format(user_profile["username"]))
+        logger.error(f'{user_profile["username"]} is not a bot account. Please upgrade it to a bot account!')
```
MarkZH, line 714: If you use `username` instead of `user_profile["username"]`, then you can use double quotes on the whole string.

MarkZH:
I approve. Do one more round of testing on your end and go ahead and merge when you're ready. My bots don't use local or online books so I can't run those lines. All other functions seem to run fine.

MarkZH (approved):

PR #444, opened 2022-02-19
Record games played by bot to local files

### LINKED ISSUE #175: Write bot move info (eval, pv, depth) into a pgn
The simplest way to do this would probably be to always write to some local pgn file, e.g. with the pgn name being the game id. (that way even through dis- and reconnecting it should (?) still write to the same file)
Then for now the bot maintainer can upload the pgn manually into some study or similar although later it might also be interesting to add an "!analysis" command that spectators or the opponent can use to have the bot automatically upload the pgn into a study after the game, provided there is a lichess api endpoint for that.

As for what should be written in there, I think the relevant infos are 1. pv 2. eval 3. depth.

MarkZH:
Add a configuration to have lichess-bot record a bot's games to a folder on the user's computer. The games are written in PGN format and include the bot's evaluation of the move--including the score and principle variation.

Depending on what characters are allowed in lichess account names, some more characters may need to be filtered out of file names. Right now, only question marks are removed (as from uncertain chess ratings).

Also, this may be too paranoid:
```python
try:
    print_pgn_game_record(config, game, board, engine, start_datetime)
except Exception:
    pass
```

Closes #175

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,10 @@ will precede the `go` command to start thinking with `sd 5`. The other `go_comma
     hello: Hi, {opponent}! I'm {me}. Good luck!
     goodbye: Good game!
 ```
+  - `pgn_directory`: Write a record of every game played in PGN format to files in this directory. Each bot move will be annotated with the bot's calculated score and principal variation. Each game will be written to a uniquely named file.
+```yml
+  pgn_directory: "game_records"
+```
 
 ## Lichess Upgrade to Bot Account
 **WARNING: This is irreversible. [Read more about upgrading to bot account](https://lichess.org/api#operation/botAccountUpgrade).**
--- a/README.md
+++ b/README.md
@@ -224,7 +224,7 @@ will precede the `go` command to start thinking with `sd 5`. The other `go_comma
     hello: Hi, {opponent}! I'm {me}. Good luck!
     goodbye: Good game!
 ```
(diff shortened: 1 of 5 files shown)


AttackingOrDefending:
Can you change single quotes to double quotes in test_bot.py, so it is like the rest of the code?

MarkZH:
Or, I could just get rid of all the invalid characters myself. The game ID will keep separate games in separate files.

Also, one feature I forgot to mention: if a game on lichess gets interruptedd (say, by network trouble or disconnecting during a correspondence game), the partial game will be written to the file. When the game is resumed, the same file will be completed with the rest of the game. However, stopping lichess-bot with Ctrl-C will stop any game information from being written, so if a game is in progress, Ctrl-C will lose the bot evaluation information, even if the game is completed later.

AttackingOrDefending:
`try: except:` for files can probably be replaced with `os.path.exists(path)`.
To add the engine score I think you can use https://python-chess.readthedocs.io/en/latest/pgn.html#chess.pgn.GameNode.set_eval.

MarkZH:
Hold off on merging for a bit. With these new methods, the score written to the PGN is always from white's perspective, even if the bot is playing black. I reported this as [an issue at the python-chess repository](https://github.com/niklasf/python-chess/issues/856) asking about this.

MarkZH:
OK. This is ready for review.

AttackingOrDefending (approved):

PR #447, opened 2022-02-19
Simplify dict get

MarkZH:
Replace the pattern `dict.get(key, default) or default` with `dict.get(key) or default`. This should be functionally equivalent and reduce the length of lines.

- If `key` does not appear in the `dict`, then `get()` returns `None` and the `default` is used.
- If `key` does appear and the result is `None`, `""`, `[]`, `{}`, or similar, then the default is used. This will happen if a configuration appears in a JSON file but no value is written.
- If `key` does appear and has a value, then that value will be used.

This pattern is primarily for replacing `None` values from `dict.get()` with empty values of the appropriate type.

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -32,7 +32,7 @@ def create_engine(config):
     else:
         raise ValueError(
             f"    Invalid engine type: {engine_type}. Expected xboard, uci, or homemade.")
-    options = remove_managed_options(cfg.get(f"{engine_type}_options", {}) or {})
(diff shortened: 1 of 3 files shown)


AttackingOrDefending:
There are also `dict.get(key, default) or default` in engine_wrapper.py in lines 69 and 180.

MarkZH:
The `dict.pop()` method raises an exception if the key is not found and there is no default provided. Replacing those lines would require `try: except:` blocks.

PR #452, opened 2022-03-06
Support changing engine for each variant

TheYoBots:
Well since there is a trend to create a duplicate PR here goes nothing. I'd honestly expect a duplicate to be created once a review is made for this PR so till then this will stay open.

This PR allows users to change engine for each variant. Under `name` all options to set engine for each variant is present. I've even allowed for using seperate working dirs for seperate variants. This could be useful when seperate variant engines have seperate files they use that will be stored in seperate working dirs (If this is a bit too much I could remove this). In this PR users will also be able to set different UCI options for different variants (not all variant engines use the same options). I'm not sure how xboard options work with lichess bot so setting different xboard options for different variants in not possible through this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,8 +49,8 @@ pip install -r requirements.txt
 ## Setup Engine
 Within the file `config.yml`:
 - Enter the directory containing the engine executable in the `engine: dir` field.
-- Enter the executable name in the `engine: name` field (In Windows you may need to type a name with ".exe", like "lczero.exe")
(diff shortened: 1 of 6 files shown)


MarkZH:
This feature turned out to be much more complicated than any of us thought. It either requires a user's config.yml file to become much larger to configure every different variant (like this PR), or it puts so many limitations on the engines--same directory, same command-line options, same UCI features, etc.--that the feature can't be used with different engines (like the other PRs). So, I'm closing all the related PRs and the issue that inspired them.

Thanks for the work you put in.

PR #454, opened 2022-03-09
Make PGN TimeControl header compliant

MarkZH:
According to the PGN spec, the time control is specified in seconds for
both the initial time and the increment.


TheYoBots:
Just wondering, instead of using python-chess can't lichess' api be used to record the pgn of a game to a local file?

AttackingOrDefending:
It can, but we wouldn't have the engine evaluation then. In #175, it was requested to save the engine pv, eval and depth, which wouldn't be possible by downloading the PGN from lichess.

MarkZH:
I could probably download the PGN from lichess and then add the commentary. That could simplify the lichess-bot code. I'll try that and, if it works, make another PR.

AttackingOrDefending (approved):

PR #465, opened 2022-04-02
Use default PGN visitor

MarkZH:
With the latest update to python-chess, a custom PGN visitor is no longer
necessary. The new default visitor joins comments in the same way as this
overriding method.


MarkZH:
This can be merged once the next version of python-chess is released.

AttackingOrDefending (approved):

PR #469, opened 2022-04-24
Update config.yml.

stockfish15skd:


AttackingOrDefending:
Every bot has a different OAuth token and adding `stockfish15exe` makes the bot crash. If this is your bot's token I would recommend deleting it and making a new one because anyone can copy the token and use your bot's account.

PR #479, opened 2022-05-30
Formatting Code

Riffecs:
The commit formats the source code according to current standards. Nothing is changed in terms of content. 
I hope you can merge the pull request.

### DIFF
--- a/ColorLogger.py
+++ b/ColorLogger.py
@@ -26,6 +26,7 @@ def add_coloring_to_emit_windows(fn):
     # add methods we need to the class
     def _out_handle(self):
         import ctypes
+
         return ctypes.windll.kernel32.GetStdHandle(self.STD_OUTPUT_HANDLE)
 
     # noinspection PyUnusedLocal
@@ -33,6 +34,7 @@ def add_coloring_to_emit_windows(fn):
 
     def _set_color(self, code):
         import ctypes
+
         # Constants from the Windows API
         self.STD_OUTPUT_HANDLE = -11
         hdl = ctypes.windll.kernel32.GetStdHandle(self.STD_OUTPUT_HANDLE)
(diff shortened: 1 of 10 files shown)


AttackingOrDefending:
While I do believe that some of the changes are good, I personally don't like breaking statements in a lot of lines even if the line is long.

> The commit formats the source code according to current standards.

I tried formatting with [Black](https://github.com/psf/black) and got similar results. Did you use a code formatter and what standards did you use (e.g. PEP 8)?

AttackingOrDefending:
I think changes like [here](https://github.com/ShailChoksi/lichess-bot/blob/master/lichess-bot.py#L261) or [here](https://github.com/ShailChoksi/lichess-bot/blob/master/lichess-bot.py#L504) are good but I don't like changes like [here](https://github.com/Riffecs/lichess-bot/blob/master/lichess-bot.py#L311-L322). Since it uses PEP-8, I will agree on merging it if MarkZH wants to. I would also suggest adding black to GitHub workflows, so it runs automatically and checks if a PR is formatted. The flake8 errors dropped from 67 to 35, which is good.

MarkZH:
I agree with @AttackingOrDefending that most of these changes don't make the code more readable (I find placing a comma after the last parameter or last list item to be annoying). However, some of the changes to overlong lines like [this one](https://github.com/ShailChoksi/lichess-bot/pull/479/files#diff-e326008cac463869bad7488e0e9b058e652abe5c59e980c0476f05bb86b10508L504-R687) in lichess-bot.py are great. I think we should take flake8 warnings about overlong lines more seriously and use this PR as a guide for how to fix them.

Riffecs:
I think I can close this pull request. Obviously there is no interest in it. If someone wants to merge the pull request they are welcome to do so.

PR #484, opened 2022-06-10
Multi engine

aceventurapetdetective4:


MarkZH:
This branch is for a specific niche purpose: playing with both leela and stockfish simultaneously. This is useless for general users who are using other bots or only one of these engines. These changes are also being applied to a very out-of-date version of lichess-bot. This will not be merged.

niqadiqa:
NIGGASEX

PR #488, opened 2022-07-01
Remove dependencies not required by lichess-bot

MarkZH:
The modules that have been deleted from requirements.txt are installed
as dependencies of other modules, in this case, by requests.

The lines importing urllib3 are deleted since requests should never throw
ProtocolError. Only requests exceptions should be thrown and caught
when contacting lichess.org.


MarkZH:
I left pytest as a requirement even though it is a dependency of pytest-timeout since there is lichess-bot code that explicitly calls pytest.

AttackingOrDefending (approved):

PR #490, opened 2022-07-02
When logging to a file, keep logging to console

MarkZH:
If the user opts to record lichess-bot logs to a file (-l), the program will keep printing messages to the console. Before, lichess-bot could only log to the console or to a file, but not both.

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,6 +270,11 @@ To output more information (including your engine's thinking output and debuggin
 python3 lichess-bot.py -v
 ```
 
+If you want to record the output to a log file, add the `-l` or `--logfile` along with a file name:
+```
+python3 lichess-bot.py --logfile log.txt
+```
+
 ## To Quit
 - Press `CTRL+C`.
 - It may take some time to quit.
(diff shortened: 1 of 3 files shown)


MarkZH:
This definitely needs checking with Linux to make sure repeated messages don't start showing up.

As it is now, the log file records at the same verbosity as the console output. Should the text log always be verbose?

AttackingOrDefending:
It doesn't work on my Linux VM. It doesn't print the engine stats (eval, depth, etc.) and the engine communication.
Before (file with -v -l):
old
After:
Terminal with -v -l:
vt
File with -v -l:
vf
Terminal with -l:
it
File with -l:
if

MarkZH:
OK. Let's try something similar to what worked before.

MarkZH:
Note to self: always run pytest before pushing.

AttackingOrDefending:
It works now. Is it ready to be merged?
Terminal:
nfv
Logs:
nfv2

MarkZH:
Yes, this is ready to merge.

AttackingOrDefending (approved):

PR #493, opened 2022-07-03
Add random time controls in matchmaking

Heiaha:
One change I think could be helpful for matchmaking is to allow users to give a list of time control settings to be chosen at random so that bots can create challenges for more than a single time control. This pull request includes a very simple change to add that functionality.


matchmaking.py
```diff
@@ -45,8 +45,8 @@ def choose_opponent(self):
         variant = self.matchmaking_cfg.get("challenge_variant") or "random"
         if variant == "random":
             variant = random.choice(self.variants)
-        base_time = self.matchmaking_cfg.get("challenge_initial_time", 60)
-        increment = self.matchmaking_cfg.get("challenge_increment", 2)
+        base_time = random.choice(self.matchmaking_cfg.get("challenge_initial_time") or [60])
+        increment = random.choice(self.matchmaking_cfg.get("challenge_increment") or [2])  # [0] evaluates to True
```
MarkZH, line 49: I don't understand the comment on this line.

    # [0] evaluates to True

Is this an explanation for the choice of `[2]` as a default? A warning against putting only `- 0` in the user's `config.yml` file?

Heiaha: No warning, just to make it clear that if the user puts 0 as the only option for the increment, it will not default the increment to 2. It will only default to 2 if no increment values are given.

MarkZH:
Good idea for a feature! Is this compatible with prior versions of config files? In other words, if a user specifies the game time and increment in a non-list format, does it result in an error in the program?

MarkZH:
This PR doesn't work with the current style of `config.yml`. If a user has this in their config file (like I do):
```yaml
matchmaking:
  challenge_initial_time: 120  # Initial time in seconds of the challenge.
  challenge_increment: 0      # Increment in seconds of the challenge.
```
then they will get the following error: `TypeError: object of type 'int' has no len()` We don't want users to have lichess-bot crash with errors if they upgrade.

On way to handle this can be found [here](https://github.com/ShailChoksi/lichess-bot/blob/86309a0d01444646dbc96b62953bbd91ee79133a/lichess-bot.py#L410)
```python
    if isinstance(books, str):
        books = [books]
```
Previously, a bot could only consult one opening book. When we changed to allow for multiple books, we used this to make sure users didn't get errors when they used the older style config file. In your case, you could do something like
```python
base_time_list = self.matchmaking_cfg.get("challenge_initial_time")
if not isinstance(base_time_list, list):
    base_time_list = [base_time_list or 60]
base_time = random.choice(base_time_list)
```
You'll notice that increment is 0 in my config file, so that will need special handling as well.
```python
increment_list = self.matchmaking_cfg.get("challenge_increment")
if not isinstance(increment_list, list):
    if increment_list is not None:
        increment_list = [increment_list]
    else:
        increment_list = [2]
base_increment = random.choice(increment_list)
```

mheinzel:
Hi @Heiaha!
Interestingly, I had a similar idea a few days ago: https://github.com/mheinzel/lichess-bot/commit/cba4141d28438efdd4e0d359d450a3f60714a40d

The approach chosen there gives a bit more flexibility, as it allows to randomly pick other settings as well (e.g. variant and mode) and not all combinations of settings need to be possible (e.g. one could pick exactly 3+2 and 1+0 without 3+0 and 1+2).

An example config could look like this (the `challenge_name` field just seemed useful for logging):

```yaml
matchmaking:
  allow_matchmaking: true
  challenge_timeout: 30

  challenge_variant: "random"
  challenge_mode: "random"
  challenge_initial_time: 180
  challenge_increment: 2
  opponent_min_rating: 1000
  opponent_max_rating: 2000
  challenges:
    # Keep all settings from the top level
    - challenge_name: "3+2"

    # Override time control
    - challenge_name: "10+5"
      challenge_initial_time: 600
      challenge_increment: 5

    - challenge_name: "1+0 easy only"
      challenge_initial_time: 60
      challenge_increment: 0
      opponent_max_rating: 1500

    - challenge_name: "Chess960 unrated"
      challenge_variant: "chess960"
      challenge_mode: "casual"
```

The way how overriding settings works also makes it easy to be backwards compatible. An old config file keeps working as before. It's treated as having a single `challenges` entry, which is empty.

What do you think about this design? Not sure if it's over-engineered, but maybe some of it is useful?

mheinzel:
As I said, my design is backwards-compatible, so users can keep their simple config files they already have if they are not interested in the advanced features.

However, I get that @Heiaha's design offers a good middle ground between complexity and power, so I can understand if you prefer that. And I can of course keep using my fork to solve my use case.

MarkZH:
@Heiaha Are you still working on this, or would you like us to finish it?

Heiaha:
@MarkZH Hi yes, sorry I have been away. I can finish it before the end of the week.

Heiaha:
Hi @MarkZH, the compatibility issue should be resolved. Let me know if there's anything else that looks off.

MarkZH:
Looks good and works well when I tried it with old parameters and lists. My one suggestion is to make a method to gather the repeated code. Something like
```python
def get_time(self, name, default):
    base_times = self.matchmaking_cfg.get(name, default)
    if isinstance(base_times, int):
        base_times = [base_times]
    return random.choice(base_times)
```
then, in `choose_opponent()`,
```python
    base_time = get_time("challenge_initial_time", 60)
    increment = get_time("challenge_increment", 2)
```

AttackingOrDefending:
Can you add random time controls also for correspondence chess?

Heiaha:
Sounds good. I've consolidated the time-getting into a method and added the ability to do random time controls for correspondence.

MarkZH:
I like how this turned out.

I've noticed that, if the user puts `0` as one of the choices for `challenge_days`, then it's possible for matchmaking to start both correspondence and real-time games since `if days:` [here](https://github.com/ShailChoksi/lichess-bot/blob/69620e64da352346f4013b5dc33db832836c36d6/matchmaking.py#L38) would be false. Furthermore, the more times `0` is listed, the less likely a correspondence game is. Should this be documented, or left as an undocumented feature? @Heiaha @AttackingOrDefending

Heiaha:
My first thought would be to add a check to config.py to check if 0 is in the options for ```challenge_days``` and either raise an exception or at least give the user a warning. How do we feel about this?

MarkZH:
The idea behind my calling this a "feature" is that it would allow matchmaking with both real-time and correspondence games. Right now, matchmaking can create correspondence games if `challenge_days` is a non-zero number or real-time games otherwise, but not both. Allowing `0` in `challenge_days` would allow for matchmaking with both kinds of games.

Actually, this idea is better handled in another PR. I'll do some testing of the current PR and then approve.

matchmaking.py
```diff
@@ -47,13 +47,23 @@ def create_challenge(self, username, base_time, increment, days, variant):
         except Exception:
             return None
 
+    def get_time(self, name, default=None):
+        match_time = self.matchmaking_cfg.get(name, default)
+        if match_time is None:
+            return
```
MarkZH, line 53: `return None` is better here, since this function is expected to return a value instead of just exiting. I know `return` and `return None` are equivalent, but the second reads better in this context.

config.yml.default
```diff
@@ -130,10 +130,16 @@ greeting:
 matchmaking:
   allow_matchmaking: false    # Set it to 'true' to challenge other bots.
   challenge_variant: "random" # If set to 'random', the bot will choose one variant from the variants enabled in 'challenge.variants'.
-  challenge_timeout: 30       # Create a challenge after being idle for 'challenge_timeout' minutes. The minimum is 1 minute.
-  challenge_initial_time: 60  # Initial time in seconds of the challenge.
-  challenge_increment: 3      # Increment in seconds of the challenge.
-# challenge_days: 2           # Days for correspondence challenge. If this option is enabled, a correspondence challenge will be created even if 'challenge_initial_time' is enabled.
+  challenge_timeout: 30       # Create a challenge after being idle for 'challenge_timeout' minutes.
```
MarkZH, line 133: Please put back the note that "The minimum is 1 minute."

README.md
```diff
@@ -235,9 +235,9 @@ will precede the `go` command to start thinking with `sd 5`. The other `go_comma
   - `allow_matchmaking`: Whether to challenge other bots.
   - `challenge_variant`: The variant for the challenges. If set to `random` a variant from the ones enabled in `challenge.variants` will be chosen at random.
   - `challenge_timeout`: The time (in minutes) the bot has to be idle before it creates a challenge.
-  - `challenge_initial_time`: The initial time (in seconds) for the challenges.
-  - `challenge_increment`: The increment (in seconds) for the challenges.
-  - `challenge_days`: The days for a correspondence challenge. If this option is enabled, a correspondence challenge will be created even if `challenge_initial_time` is enabled.
+  - `challenge_initial_time`: The initial time (in seconds and to be chosen at random) for the challenges.
+  - `challenge_increment`: The increment (in seconds and to be chosen at random) for the challenges.
+  - `challenge_days`: The days for a correspondence challenge (to be chosen at random). If this option is enabled, a correspondence challenge will be created even if `challenge_initial_time` is enabled.
```
MarkZH, line 240: The descriptions here should note that a list of times is expected. Otherwise, "chosen at random" doesn't make sense.

Heiaha:
Updated.

MarkZH (approved):

AttackingOrDefending (approved):

PR #500, opened 2022-07-18
Delete logging_pool.py

MarkZH:
The functionality of LoggingPool can be replaced with the error_callback
argument of Pool.apply_async(). This will also place traceback information
into the user's log file.


MarkZH:
Before:
before

After:
after

AttackingOrDefending (approved):

PR #503, opened 2022-07-20
Increase matchmaking wait time

AttackingOrDefending:
Increases matchmaking wait time to 60 seconds.

### DIFF
--- a/matchmaking.py
+++ b/matchmaking.py
@@ -14,18 +14,20 @@ class Matchmaking:
         self.last_challenge_created = time.time()
         self.last_game_ended = time.time()
         self.challenge_expire_time = 25  # The challenge expires 20 seconds after creating it.
+        self.challenge_timeout = max(self.matchmaking_cfg.get("challenge_timeout") or 30, 1)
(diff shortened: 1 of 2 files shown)


MarkZH:
Since the issue was closed, do you still want to merge this?

AttackingOrDefending:
Since redsalmon91 said that they had to increase the matchmaking time to fix the problem, I believe that it would be good to make some changes. I think we should either merge this or at least add
```python
self.challenge_timeout = max(self.matchmaking_cfg.get("challenge_timeout") or 30, 1)
```
to guarantee a minimum wait time.

matchmaking.py
```diff
@@ -14,18 +14,20 @@ def __init__(self, li, config, username):
         self.last_challenge_created = time.time()
         self.last_game_ended = time.time()
         self.challenge_expire_time = 25  # The challenge expires 20 seconds after creating it.
+        self.challenge_timeout = max(self.matchmaking_cfg.get("challenge_timeout") or 30, 1)
```
MarkZH, line 17: Convert `challenge_timeout` to seconds here so everything in `should_create_challenge()` is in seconds.

MarkZH (approved):

PR #514, opened 2022-07-30
Stop using online opening books after x out of book moves

AttackingOrDefending:
After `max_out_of_book_moves` of the online opening books (chessdb, lichess cloud analysis) not having a move, we will stop using them, because they are unlikely to have a move for future positions and these move requests take a lot of time (especially for bullet).
Also, changed lichess cloud analysis to show the score from the perspective of the user and added dtz score to the chessdb egtb.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ Besides the above, there are many possible options within `config.yml` for confi
     - `offer_draw_moves`: The absolute value of the evaluation has to be less than or equal to `offer_draw_score` for `offer_draw_moves` amount of moves for the bot to offer/accept draw.
     - `offer_draw_pieces`: The bot only offers/accepts draws if the position has less than or equal to `offer_draw_pieces` pieces.
 - `online_moves`: This section gives your bot access to various online resources for choosing moves like opening books and endgame tablebases. This can be a supplement or a replacement for chess databases stored on your computer. There are three sections that correspond to three different online databases:
+    - `max_out_of_book_moves`: Stop using online opening books after they don't have a move for `max_out_of_book_moves` positions. Doesn't apply for the online endgame tablebases.
     1. `chessdb_book`: Consults a [Chinese chess position database](https://www.chessdb.cn/), which also hosts a xiangqi database.
     2. `lichess_cloud_analysis`: Consults [Lichess' own position analysis database](https://lichess.org/api#operation/apiCloudEval).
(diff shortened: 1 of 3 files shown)


lichess-bot.py
```diff
@@ -32,6 +32,8 @@
 
 terminated = False
 
+out_of_online_opening_book_moves = 0
```
MarkZH, line 35: This needs to be a dictionary from game IDs to failed lookup counts. As it is now, the first game to fail to find enough moves will prevent all subsequent games from using the opening book. Simultaneous games will exhaust the opening book faster.

To keep the dictionary from growing arbitrarily large as games accumulate, you can watch for the `gameFinish` state in the `start()` function to remove entries from the dictionary.

README.md
```diff
@@ -81,6 +81,7 @@ Besides the above, there are many possible options within `config.yml` for confi
     - `offer_draw_moves`: The absolute value of the evaluation has to be less than or equal to `offer_draw_score` for `offer_draw_moves` amount of moves for the bot to offer/accept draw.
     - `offer_draw_pieces`: The bot only offers/accepts draws if the position has less than or equal to `offer_draw_pieces` pieces.
 - `online_moves`: This section gives your bot access to various online resources for choosing moves like opening books and endgame tablebases. This can be a supplement or a replacement for chess databases stored on your computer. There are three sections that correspond to three different online databases:
+    - `max_out_of_book_moves`: Stop using online opening books after they don't have a move for `max_out_of_book_moves` positions. Doesn't apply for the online endgame tablebases.
```
MarkZH, line 84: This line should be moved below the DB listings (since the previous sentence says that it's going to name them) and above the DB-specific configurations (above `- Configurations common to all`.

lichess-bot.py
MarkZH: Actually, a `defaultdict` with `default_factory=int` is even better to avoid the boilerplate `if game.id not in out_of_online_opening_book_moves:`.

AttackingOrDefending: I moved it inside `play_game` to also avoid `global`.

MarkZH: The problem here is that the counts are lost if `play_game()` exits in the middle of a game. This could happen in a correspondence game or because of connection issues.

If you want to avoid globals, put the dictionary version inside `start()` before the main loop and pass it into `play_game()`. Since dictionaries are mutable, you can modify them and the modifications will persist after the function call, so there's no need to return it.

Also, [Counter](https://docs.python.org/3/library/collections.html#collections.Counter) is probably the simplest data structure to use for this.

MarkZH (approved):

PR #518, opened 2022-07-31
Use separate method for online chess DB access

MarkZH:
Since the api_get() method uses backup to retry communications, any
problems with communicating with the online chess databases can
result in retrying for up to a minute. This would waste time that bots
could use to pick their own moves.

This commit replaces the Lichess.api_get() method with another that
does not have backoff, so it does not retry requests.

### DIFF
--- a/lichess.py
+++ b/lichess.py
@@ -61,12 +61,12 @@ class Lichess:
                           giveup=is_final,
                           backoff_log_level=logging.DEBUG,
                           giveup_log_level=logging.DEBUG)
-    def api_get(self, path, raise_for_status=True, get_raw_text=False, params=None):
+    def api_get(self, path, get_raw_text=False):
         logging.getLogger("backoff").setLevel(self.logging_level)
         url = urljoin(self.baseUrl, path)
-        response = self.session.get(url, timeout=2, params=params)
-        if rate_limit_check(response) or raise_for_status:
-            response.raise_for_status()
+        response = self.session.get(url, timeout=2)
+        rate_limit_check(response)
+        response.raise_for_status()
         response.encoding = "utf-8"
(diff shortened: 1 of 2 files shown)


lichess.py
```diff
@@ -153,3 +153,8 @@ def challenge(self, username, params):
     def cancel(self, challenge_id):
         return self.api_post(ENDPOINTS["cancel"].format(challenge_id),
                              raise_for_status=False)
+
+    def online_book_get(self, path, params=None):
+        response = self.session.get(path, timeout=2, params=params)
+        response.raise_for_status()
```
AttackingOrDefending, line 159: `response.raise_for_status()` will raise an error when lichess cloud eval doesn't have a move. This is why the `raise_for_status` parameter was added to `api_get`. I think in this case you can remove it entirely instead of adding a parameter.

AttackingOrDefending: `raise_for_status` and `params` can also be removed from `api_get`.

MarkZH: The `challenge()` and `cancel()` methods have `raise_for_status=False`, so I don't think that parameter can be removed from `api_get()`. Several methods have `params` like `make_move()`, so I don't think this can be removed either.

AttackingOrDefending: `challenge`, `cancel` and `make_move` use `api_post` and not `api_get`, so they can be removed from `api_get`.

MarkZH: I see. Parameters removed. `get_game_pgn()` used `params`, but that didn't actually do anything.

I have a feeling that the `data` and `payload` parameters of `api_post()` are redundant, but that's for another PR.

AttackingOrDefending (approved):

Issue #222, opened 2020-09-25
bot appears offline and doesn't accept challenges


vondele:
After having been online and accepting changes for a while, my bot (https://lichess.org/@/simpleEval) doesn't accept challenges and appears offline (empty dot, instead of green dot next to the name). (The webpage mentions the bot was active 1h ago, about 3h after the last game). This has been happening always. The verbose log doesn't indicate anything that strikes me as special:
```
2020-09-25 18:18:07,163: <PopenProcess at 0xb5aeab50 (pid=19875)> >> info depth 18 seldepth 19 multipv 1 score cp 599 nodes 233216 nps 656946 tbhits 0 time 355 pv b7b5 a2a3 d8b6 g1f2 b6b7 f1g1 e7g8 g1h1 h8h1 g2h1 b7c8 d2f1 e8f7 f1g3 g4f3 h1f3 a5a4 f3h5 f7e7
2020-09-25 18:18:07,443: <PopenProcess at 0xb5aeab50 (pid=19875)> >> info depth 19 seldepth 23 multipv 1 score cp 599 nodes 419584 nps 660762 tbhits 0 time 635 pv b7b5 a2a3 d8b6 g1f2 b6b7 f1g1 e7g8 g1h1 h8h1 g2h1 b7c8 d2f1 e8f7 f1g3 a5a4 d1d3 f7g7 d4e5 d6e5 h1g2 a4b3
2020-09-25 18:18:07,889: <PopenProcess at 0xb5aeab50 (pid=19875)> >> info depth 20 seldepth 22 multipv 1 score cp 599 nodes 717037 nps 663923 hashfull 20 tbhits 0 time 1080 pv b7b5 a2a3 d8b6 g1f2 b6b7 f1h1 h8h1 d1h1 e7g6 h1h6 b7g7 h6h1 c6c5 d4d5 a5a4 b3b4 g4f3 d2f3 g7f7 b4c5 d6c5
2020-09-25 18:18:08,866: Ignoring illegal move h5g4 on board rn1qk2r/1p2n3/2pp1p2/p3p3/4P1b1/1P1PPN2/PBPN2B1/3R1RK1 w kq - 0 18
2020-09-25 18:18:08,871: --- https://lichess.org/pU512AqJ/black Game over
2020-09-25 18:18:08,872: <PopenProcess at 0xb5aeab50 (pid=19875)> << stop
[...]

PR #525, opened 2022-08-03
Reset connection if bot is offline

### LINKED ISSUE #135: Lichess server doesn't reconnect when client IP changes
My internet provider changes my IP once a day which messes with the bot client. (I haven't observed problems when playing as a human though) I guess for that to not be a problem the client would need to send some keepalive signal and if it doesn't get returned start a new connection? Maybe worth noting that one can still send moves with the changed IP address but one doesn't get the response moves anymore. (even though by receiving the move, lichess should already learn the new IP)
Log:
"DEBUG:chess.engine:<PopenProcess at 0x7fe7c3118940 (pid=7960)> >> info depth 9 currmove f4g6 currmovenumber 22
DEBUG:chess.engine:<PopenProcess at 0x7fe7c3118940 (pid=7960)> >> info depth 9 score cp -194 nodes 140077506 nps 396496 time 353288 pv f1e1 c4c2 a3a7 d8b8 e1d1 c2h2 d1c1 b8b2 a2a4
DEBUG:chess.engine:<PopenProcess at 0x7fe7c3118940 (pid=7960)> >> bestmove f1e1
DEBUG:urllib3.connectionpool:Resetting dropped connection: lichess.org
DEBUG:urllib3.connectionpool:https://lichess.org:443 "POST /api/bot/game/tPh8e45K/move/f1e1 HTTP/1.1" 200 None
DEBUG:chess.engine:<PopenProcess at 0x7fe7c31136a0 (pid=9258)> >> info depth 8 currmove g8g7 currmovenumber 25
DEBUG:chess.engine:<PopenProcess at 0x7fe7c31136a0 (pid=9258)> >> info depth 8 currmove h7h6 currmovenumber 26
DEBUG:chess.engine:<PopenProcess at 0x7fe7c31136a0 (pid=9258)> >> info depth 8 currmove h7h5 currmovenumber 27
DEBUG:chess.engine:<PopenProcess at 0x7fe7c31136a0 (pid=9258)> >> info depth 8 currmove h8g7 currmovenumber 28
DEBUG:chess.engine:<PopenProcess at 0x7fe7c31136a0 (pid=9258)> >> info depth 8 currmove h8f6 currmovenumber 29
DEBUG:chess.engine:<PopenProcess at 0x7fe7c31136a0 (pid=9258)> >> info depth 8 currmove h8e5 currmovenumber 30
DEBUG:chess.engine:<PopenProcess at 0x7fe7c31136a0 (pid=9258)> >> info depth 8 score cp 10 nodes 233495743 nps 288036 time 810647 pv d7d6 c3d2 d8e6 e2e3 f4d5 d1c3 d5c3 d2c3
DEBUG:chess.engine:<PopenProcess at 0x7fe7c31136a0 (pid=9258)> >> bestmove d7d6
DEBUG:urllib3.connectionpool:Resetting dropped connection: lichess.org
DEBUG:urllib3.connectionpool:https://lichess.org:443 "POST /api/bot/game/gBPk40zz/move/d7d6 HTTP/1.1" 200 None
"

These are the corresponding games, as you can see in both games the move did still get played just no response on the moves played by SF: https://lichess.org/tPh8e45K#98 https://lichess.org/gBPk40zz/black#9

AttackingOrDefending:
Idea taken from https://github.com/ShailChoksi/lichess-bot/issues/222#issuecomment-1201663501.
Attempts to fix #135 and #222.
Checks every 5 minutes if the bot is offline and resets the connection. While matchmaking may be resetting the connection (https://github.com/ShailChoksi/lichess-bot/issues/222#issuecomment-1201088500), not all bots have matchmaking enabled.

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -121,7 +121,7 @@ def game_error_handler(error):
 def should_restart(li, username):
     online_bots = li.get_online_bots()
     bot = list(filter(lambda bot: bot["username"] == username, online_bots))
-    return bot and bot[0].get("online")
+    return not (bot and bot[0].get("online", True))
 
 
 def start(li, user_profile, config, logging_level, log_filename, one_game=False):
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -277,8 +277,9 @@ def start(li, user_profile, config, logging_level, log_filename, one_game=False)
                 logger.info("Challenging a random bot")
                 matchmaker.challenge()
 
-            if time.time() > last_check_online_time + 300:
+            if time.time() > last_check_online_time + 10:
                 if should_restart(li, user_profile["username"]):
+                    logger.info("Will reset connection with lichess")
                     li.reset_connection()
                 last_check_online_time = time.time()
 
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -120,8 +120,8 @@ def game_error_handler(error):
 
 def should_restart(li, username):
     online_bots = li.get_online_bots()
-    bot = list(filter(lambda bot: bot["username"] == username, online_bots))[0]
-    return bot.get("online")
+    bot = list(filter(lambda bot: bot["username"] == username, online_bots))
+    return bot and bot[0].get("online")
 
 
 def start(li, user_profile, config, logging_level, log_filename, one_game=False):
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -120,8 +120,8 @@ def game_error_handler(error):
 
 def should_restart(li, username):
     online_bots = li.get_online_bots()
-    is_bot_online = not list(filter(lambda bot: bot["username"] == username, online_bots))
-    return is_bot_online
+    is_bot_offline = not list(filter(lambda bot: bot["username"] == username, online_bots))
+    return is_bot_offline
 
 
 def start(li, user_profile, config, logging_level, log_filename, one_game=False):
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -118,12 +118,6 @@ def game_error_handler(error):
     logger.exception("Game ended due to error:", exc_info=error)
 
 
-def should_restart(li, username):
-    online_bots = li.get_online_bots()
-    is_bot_offline = not list(filter(lambda bot: bot["username"] == username, online_bots))
-    return is_bot_offline
-
-
 def start(li, user_profile, config, logging_level, log_filename, one_game=False):
     challenge_config = config["challenge"]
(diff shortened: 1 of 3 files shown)


lichess-bot.py
```diff
@@ -118,6 +118,12 @@ def game_error_handler(error):
     logger.exception("Game ended due to error:", exc_info=error)
 
 
+def should_restart(li, username):
+    online_bots = li.get_online_bots()
+    bot = list(filter(lambda bot: bot["username"] == username, online_bots))
+    return bot and bot[0].get("online")
```
MarkZH, line 124: I don't understand this test. I can understand checking that my bot is in the returned list. But, when I look at the my bot's JSON, there's no `"online"` field, so the result is False because `get()` returns `None`. But, if the data had `online: true`, that would cause a restart?

MarkZH: This PR does work for me, but I've never had trouble with losing the connection to lichess.

AttackingOrDefending: I fixed the check at line 124. According to the [lichess API docs](https://lichess.org/api#operation/apiBotOnline) there should be an "online" field. I remember in the past [lichess](https://lichess.org/player/bots) would show also a few bots that were offline, so maybe this was the reason that they added the online field. Probably they have fixed it, so it now only returns bots that are online, so they don't send the "online" value.

I also never had a problem with my bot losing connection to lichess, but I only use my bot for testing. This is just an attempt to fix the problem that was reported in the PRs mentioned above.

MarkZH:
Looks good, although I think checking every 5 minutes is too much. I think every hour would suffice.

lichess-bot.py
AttackingOrDefending: The second check with "online" was removed since the API stopped returning the "online" field and it now only returns bots that are online.

MarkZH:
Instead of downloading every online bot (which is rate-limited), you can use [the real-time user status API](https://lichess.org/api#operation/apiUsersStatus) to just query the bot's id (with practically no rate-limiting). Something like:
```python
ENDPOINTS = {
    # ...
    "user_online": "/api/users/status",
    # ...
}

class Lichess:
    # ...
    def bot_is_online(self, id):
        my_bot = self.api_get(ENDPOINTS["user_online"], params={"ids": id})
        return my_bot and my_bot[0].get("online")  # bot returns a list
```
and
```python
    if not li.bot_is_online(user_profile["id"]):
        logger.info("Will reset connection with lichess")
        li.reset_connection()
```

MarkZH (approved):

PR #527, opened 2022-08-04
Spectator chat

MarkZH:
Bot can send messages to spectators using their config.yml file.

### DIFF
--- a/conversation.py
+++ b/conversation.py
@@ -14,12 +14,12 @@ class Conversation:
     command_prefix = "!"
 
     def react(self, line, game):
-        if line.username != self.game.username:
-            logger.info(f'*** {self.game.url()} [{line.room}] {line.username}: {line.text.encode("utf-8")}')
-            if line.text[0] == self.command_prefix:
-                self.command(line, game, line.text[1:].lower())
+        logger.info(f'*** {self.game.url()} [{line.room}] {line.username}: {line.text.encode("utf-8")}')
+        if line.text[0] == self.command_prefix:
+            self.command(line, game, line.text[1:].lower())
 
     def command(self, line, game, cmd):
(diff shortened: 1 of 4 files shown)


conversation.py
```diff
@@ -14,9 +14,10 @@ def __init__(self, game, engine, xhr, version, challenge_queue):
     command_prefix = "!"
 
     def react(self, line, game):
-        logger.info(f'*** {self.game.url()} [{line.room}] {line.username}: {line.text.encode("utf-8")}')
-        if line.text[0] == self.command_prefix:
-            self.command(line, game, line.text[1:].lower())
+        if line.username != self.game.username:
```
AttackingOrDefending, line 17: Won't the bots be unable to respond to themselves with this line? I think that if a bot owner types !eval, their bot should be able to respond.

MarkZH:
Actually, I'm not sure about the last commit. Both the bot owner and the bot's opponent will be able to see the result of !eval.

AttackingOrDefending:
The chat is visible to both players and I believe that most people know that, so if the command is used by the bot owner, they understand that the opponent will be able to see the eval.

MarkZH:
OK. If there aren't any other suggestions, this is ready to merge.

mheinzel:
Small detail, but to me it sounds a bit weird to wish "good luck" to the spectators. Maybe something like "enjoy the show" or just leaving it out?

MarkZH:
You're right. That doesn't make sense.

AttackingOrDefending:
The greeting section of the config.yml is indented by 4 spaces instead of 2. It still works, but the rest of the file has 2 spaces.

AttackingOrDefending (approved):

Issue #494, opened 2022-07-05
Contributing a few features I implemented


mheinzel:
Hi!

Thanks for creating this project! It made it super easy to set up my bot and now it already played more than 2000 games on Lichess. :star_struck:

I also extended the code to tweak it to my needs and I think some features could be useful to others as well: https://github.com/ShailChoksi/lichess-bot/compare/master...mheinzel:lichess-bot:master
Before opening a bunch of PRs, I wanted to ask you whether you want them in the first place. At the moment, the changes are kind of intertwined and specific to my needs, but I'd be happy to open separate PRs for each feature you want, make changes you deem useful and also add some documentation. 

What I implemented:

1. **Randomly choosing between multiple matchmaking choices**
     As already explained in https://github.com/ShailChoksi/lichess-bot/pull/493#issuecomment-1175459973
2. **Explicitly handle hitting the rate limit for the challenge API (HTTP 429)**
  There was something about the code about not raising an exception no unsuccessful API calls, but then the response didn't contain any JSON, so it ended with a stack trace, requiring to restart the bot. This is now handled more explicitly (although there is still room for improvement I think) and receiving a 429 will lead to a 1h matchmaking pause, after which matchmaking resumes.
3. **Add challenge blocklist, allowing to exclude opponents from matchmaking**
  Some bots didn't accept my challenges for various reasons (e.g. they only play humans, (un)rated, specific time controls or variants), so the bot often needed 2-3 challenges to play a game. This quickly depleted the allowed 400 challenges/day. To avoid this, I checked the logs after a day or two and manually added some opponents to the blocklist. This helped a lot.
4. **Do not challenge bots with 0 games in this game type/variant**
  This does not make the previous feature obsolete, but in many cases where bots only play e.g. bullet and blitz, they won't have any rapid and classical games. Or an antichess bot had no regular chess games at all. Not challenging them in these cases reduced my blocklist quite a bit. Seems useful to me, but maybe there should be a way to turn it off in the config?
5. **Do not challenge bots that violated the Terms of Service**
  As I understand, you don't get rating from them anyways? I wanted to exclude them, but this could be turned into a config flag as well.
6. A few small changes to documentation and what is logged.

Which of these are you interested in?

AttackingOrDefending:
- Change 1: There is already a PR, so I don't think there is a reason to make another.
- Change 2: It is a good change.
- Change 3: I don't think many people will use this.
[...]

PR #529, opened 2022-08-04
Only challenge bots with previous games of this type

mheinzel:
The first PR coming out of #494.

This behaviour was already intended, to avoid making challenges that are very likely to be declined. But as suspected, the API returns bots with 0 games and a provisional rating of 1500 or 2000.

One question: The lambda keeps getting bigger and I'm planning to add another check for terms of service violation soon. Should I extract it into a local definition for readability?

```python
        def is_suitable_opponent(bot):
            perf = bot["perfs"].get(game_type, {})
            return (bot["username"] != self.username and
                not bot.get("disabled") and
                perf.get("games", 0) > 0 and
                min_rating <= perf.get("rating", 0) <= max_rating)

        online_bots = self.li.get_online_bots()
        online_bots = list(filter(is_suitable_opponent, online_bots))
```

### DIFF
--- a/matchmaking.py
+++ b/matchmaking.py
@@ -87,10 +87,10 @@ class Matchmaking:
 
         def is_suitable_opponent(bot):
             perf = bot["perfs"].get(game_type, {})
-            return (bot["username"] != self.username and
-                not bot.get("disabled") and
-                perf.get("games", 0) > 0 and
-                min_rating <= perf.get("rating", 0) <= max_rating)
+            return (bot["username"] != self.username
+                    and not bot.get("disabled")
+                    and perf.get("games", 0) > 0
+                    and min_rating <= perf.get("rating", 0) <= max_rating)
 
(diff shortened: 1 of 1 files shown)


AttackingOrDefending:
Yes, you can change the lambda to a def function as you did above.

AttackingOrDefending (approved):

MarkZH:
This works well, just two style comments on the `is_suitable_opponent()` function:

1. Break lines before the `and` conjuction, as in
```python
        return (bot["username"] != self.username
            and not bot.get("disabled")
            and perf.get("games", 0) > 0
... etc
```
2. Indent the lines after the return statement by another four spaces.

```python
        return (bot["username"] != self.username
                and not bot.get("disabled")
                and perf.get("games", 0) > 0
... etc
```

mheinzel:
Ah, yes. That's nicer. Updated!

MarkZH (approved):

Issue #535, opened 2022-08-08
Matchmaking opponent min/max rating is not ideal


fabriciodosanjossilva:
When using matchmaking, parameters "opponent_min_rating" and "opponent_max_rating" are not the ideal way of filtering opponents. This is because a bot have separate ratings for each time class (Bullet, Blitz, etc), which can be quite different. It might have a Bullet rating of 1800 and a Blitz rating of 1500. So, when playing Bullet games, opponents should have a rating around the bot's Bullet rating (in this case, 1800). The same for Blitz (in this case 1500). I suggest replacing these two parameters (min/max) with a new range parameter, which would consider defined time parameters. So, for this bot example, if I set this range parameter to 100 and set 15s+1s time controls (for a Bullet game), I would get opponents with rating ranging from 1700-1900. And if I set 3m+1s time controls (for a Blitz game), I would get opponents with rating ranging from 1400-1600.

MarkZH:
Hi, @fabriciodosanjossilva. Can you try out PR #536 and see if this is what you had in mind? You'll need to add a line to the `matchmaking` section of your config.yml: `opponent_rating_difference: 300` for a maximum difference of 300 in rating.

fabriciodosanjossilva:
@MarkZH, thanks for the prompt reply and code. I tested and it seems correct.
I setup a new config file using the new default file as a template. I did not uncomment opponent_min_rating and opponent_max_rating. Then I set opponent_rating_difference to a narrow range (50). Time controls were set to a Bullet game. As my bot got online, it challenged only bots within the correct rating (my bullet rating +/- 50). There were 8 challenges. I noted for one specific challenged bot, after it won a game, its rating increased and my bot's decreased, taking the difference to outside the range. It was challenged even though. It seems to be the issue you described earlier. Seems a minor issue, but this should get fixed. After that, I changed time controls for Blitz games. Challenged bots had then its rating within the range (50) of my bot's Blitz rating. It seems that everything is all right.

Thank you a lot. This feature is very useful to take bots to their deserved rating.

I recommend to set default value for opponent_rating_difference as 100. This would avoid challenging bots that you cannot beat or bots that are too easy to beat. The range should be one that is wide, but also one that wields taking or adding rating points.

AttackingOrDefending:
This feature was added in #536, so this issue can be closed.

fabriciodosanjossilva:
@MarkZH,

This is working amazingly well. Once again, thank you!

[...]

PR #536, opened 2022-08-09
Filter matchmaking based on relative ratings

MarkZH:
The user may specify a maximum rating difference when searching
for an opponent. Only those bots with a rating less than this difference
in the game to be played will be chosen.

Implements idea from #535.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,7 +269,7 @@ matchmaking:
      - 2
 # opponent_min_rating: 600
 # opponent_max_rating: 4000
-  opponent_rating_difference: 500
+  opponent_rating_difference: 100
   opponent_allow_tos_violation: true
   challenge_mode: "random"
 ```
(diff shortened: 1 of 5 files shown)


AttackingOrDefending:
I think the rating should be tracked but not after each game because the bot may get 429 if it playing many games at the same time. Everything more than a minute should be fine.

fabriciodosanjossilva:
@MarkZH, thanks for the prompt reply and code. I tested and it seems correct.
I setup a new config file using the new default file as a template. I did not uncomment opponent_min_rating and opponent_max_rating. Then I set opponent_rating_difference to a narrow range (50). Time controls were set to a Bullet game. As my bot got online, it challenged only bots within the correct rating (my bullet rating +/- 50). There were 8 challenges. I noted for one specific challenged bot, after it won a game, its rating increased and my bot's decreased, taking the difference to outside the range. It was challenged even though. It seems to be the issue you described earlier. Seems a minor issue, but this should get fixed. After that, I changed time controls for Blitz games. Challenged bots had then its rating within the range (50) of my bot's Blitz rating. It seems that everything is all right.

Thank you a lot. This feature is very useful to take bots to their deserved rating.

I recommend to set default value for opponent_rating_difference as 100, or at most 200. This would avoid challenging bots that you cannot beat or bots that are too easy to beat. The range should be one that is wide, but also one that takes or adds rating points.

AttackingOrDefending (approved):

Issue #547, opened 2022-08-20
Can hang indefinitely after one error from Lichess API


alex65536:
My installation of lichess-bot failed with the following error:

```
2022-08-15 06:36:15,636 __main__ INFO Challenging a random bot
2022-08-15 06:37:16,914 __main__ ERROR Quitting lichess-bot due to an error:
Traceback (most recent call last):
  File "/opt/sofcheck/lichess-bot/lichess-bot.py", line 923, in <module>
    start_lichess_bot()
  File "/opt/sofcheck/lichess-bot/lichess-bot.py", line 916, in start_lichess_bot
    start(li, user_profile, CONFIG, logging_level, args.logfile)
  File "/opt/sofcheck/lichess-bot/lichess-bot.py", line 271, in start
    matchmaker.challenge()
  File "/opt/sofcheck/lichess-bot/matchmaking.py", line 96, in challenge
    bot_username, base_time, increment, days, variant = self.choose_opponent()
  File "/opt/sofcheck/lichess-bot/matchmaking.py", line 88, in choose_opponent
    online_bots = self.li.get_online_bots()
  File "/opt/sofcheck/lichess-bot/lichess.py", line 142, in get_online_bots
    online_bots = self.api_get(ENDPOINTS["online_bots"], get_raw_text=True)
  File "/opt/sofcheck/venv/lib/python3.9/site-packages/backoff/_sync.py", line 105, in retry
    ret = target(*args, **kwargs)
  File "/opt/sofcheck/lichess-bot/lichess.py", line 69, in api_get
    response.raise_for_status()
  File "/opt/sofcheck/venv/lib/python3.9/site-packages/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 502 Server Error: Bad Gateway for url: https://lichess.org/api/bot/online
```

Still, some of the processes worked even after this error, and one of them was consuming 100% CPU.

Possibly, it would be better to retry this kind of temporary errors.

MarkZH:
@alex65536 Can you try the change in PR #549 and let us know if that works for your problem? Thanks.

alex65536:
I may try the proposed PR, but I don't really know how to reproduce it fast. The previous time the bot crashed after running for approximately two weeks.

If you know a simple way way to simulate a faulty Lichess API that yields 502 errors faster, it would be helpful to ensure if the proposed PR really helps.

[...]

PR #549, opened 2022-08-24
Restart on communication errors

MarkZH:
Since most communication errors are temporary, use backoff to restart
most of lichess-bot when there are uncaught problems with requests.

Giving up after an hour seems reasonable.

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -120,7 +120,7 @@ def game_error_handler(error):
     logger.exception("Game ended due to error:", exc_info=error)
 
 
-@backoff.on_exception(backoff.expo, RequestException, max_time=3600)
+@backoff.on_exception(backoff.constant, RequestException)
 def start(li, user_profile, config, logging_level, log_filename, one_game=False):
     logger.info(f"You're now connected to {config['url']} and awaiting challenges.")
     manager = multiprocessing.Manager()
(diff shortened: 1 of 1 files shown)


MarkZH:
@AttackingOrDefending  I'm starting to have doubts about this change. Would something simpler that just fixes the related issue be better?

```python
class Lichess:

    def get_online_bots(self):
        try:
            online_bots = self.api_get(ENDPOINTS["online_bots"], get_raw_text=True)
            online_bots = list(filter(bool, online_bots.split("\n")))
            return list(map(lambda bot: json.loads(bot), online_bots))
        except Exception:
            return []
```

AttackingOrDefending:
I believe that using backoff is better than adding a fix only for one function, because I would assume this error can happen with other requests also.
Why is the backoff in `start` and not in `api_get` and `api_post`?

MarkZH:
Both the methods `api_get()` and `api_post()` already have `backoff` decorations. The new `backoff` was added to `start()` for when `backoff` gives up (either from exceeding `max_time` or the status code being < 500) and lets the exception representing communication failure reach the caller. In the linked issue #547, the error came from the `backoff` function giving up on an `api_get()` call from `get_online_bots()`. The idea behind putting the new `backoff` on `start()` is to restart the communication streams with lichess (namely, `control_stream`).

AttackingOrDefending:
I now also have doubts. Probably the fix for `get_online_bots` is better.

PR #553, opened 2022-08-30
Move instantly when there is only one available move

### LINKED ISSUE #552: Automatically play when there is only one legal move
A great improvement in this wonderful repo could be to play automatically without loosing time asking the engine when there is only one legal move.

AttackingOrDefending:
closes #552


lichess-bot.py
```diff
@@ -406,7 +407,10 @@ def get_greeting(greeting):
                     fake_thinking(config, board, game)
                     print_move_number(board)
 
-                    best_move = get_book_move(board, polyglot_cfg)
+                    if instant_move and len(list(board.legal_moves)) == 1:
```
MarkZH, line 410: The following also works: `board.legal_moves.count()`

PR #558, opened 2022-09-04
Close background process and transport of engines

MarkZH:
Before this change, a Windows engine had a good chance of never closing
and endless messages of the form

<IocpProactor overlapped#=1 result#=0> is running after closing for 1788.2 seconds

would be printed to the DEBUG logging stream once a second.

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -221,6 +221,7 @@ class EngineWrapper:
 
     def quit(self):
         self.engine.quit()
+        self.engine.close()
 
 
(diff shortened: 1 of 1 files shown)


AttackingOrDefending (approved):

AttackingOrDefending:
Is this happening only on XBoard engines? I haven't seen my bot get this message.

MarkZH:
It only happens on Windows (both UCI and Xboard).



https://python-chess.readthedocs.io/en/latest/engine.html#chess.engine.SimpleEngine.close

PR #559, opened 2022-09-04
Do not challenge bots from blocked accounts

### LINKED ISSUE #537: Should not accept challenge from blocked users
As a challenge is received, linux-bot should decline if challenger is blocked. This information is available on challenger public info (https://lichess.org/api/user/{user}), on field "blocking".

MarkZH:
After choosing a bot to challenge, this PR checks if the target bot is an account that is blocked by the user. If it is, then no challenge is issued and the name of the bot is added to an internal block list to prevent that bot from being challenged again. A bot is also added to the internal block list if an error occurs while attempting to challenge it, usually because the target bot blocks the user's bot's account.

The test for blocking occurs only after a random selection is made to prevent rate limiting when fetching the public data necessary to check for blocked accounts. Unlike `tosViolation`, blocking information is not included in the data returned from `Lichess.get_online_bots()` and must be queried separately.

Fixes #537.


MarkZH:
Converted to draft since I'm starting to hit API rate limits.

MarkZH:
Now I'm being rate-limited on the master branch as well. I'll have to do further testing later.

MarkZH:
Looks like the rate-limiting was a temporary issue with my connection to lichess. The PR is ready for review.

AttackingOrDefending (approved):

PR #563, opened 2022-09-11
Delay challenging bot after challenge decline

### LINKED ISSUE #538: Matchmaking should avoid challenging user that consistently declines challenges
After challenging a user via Matchmaking, if challenged user declines the challenge, linux-bot should "remember" that for awhile, avoiding annoying the user and wasting time. I'm getting declines quite frequently from bots that: are not accepting challenges with the request time parameters; not accepting challenges form bots; or are not accepting challenges at all.

As an implementation suggestion, this can be done as a "no contact" time window, where no challenge is to be sent to same user, with the same time parameters, within this time window. The window should start low (say 5 minutes) and increase exponentially (doubling on every next decline). An upper limit may be set (say 24h). It would be great if this window (for each user) is persisted, surviving a lichess-bot restart.

It seems that public user data does not contain challenge restrictions. If this information were available, challenges that will be declined because of this public restrictions could be avoided upfront. I see that often users report their requirements by text on their profile. It's a shame that these restrictions are not available via API. This might turn into a Lichess API enhancement request. But for now, I think the best would be to implement the above described control.

MarkZH:
If an opponent declines a challenge, do not challenge them again for one hour. Increase the delay by another hour on every subsequent decline.

Will close #538.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,6 +260,7 @@ will precede the `go` command to start thinking with `sd 5`. The other `go_comma
   - `opponent_rating_difference`: The maximum difference between the bot's rating and the opponent bot's rating.
   - `opponent_allow_tos_violation`: Whether to challenge bots that violated Lichess Terms of Service. Note that even rated games against them will not affect ratings.
   - `challenge_mode`: Possible options are `casual`, `rated` and `random`.
+  - `delay_after_decline`: Whether to delay challenging a bot after that bot declines a challenge.
 
 If there are entries for both real-time (`challenge_initial_time` and/or `challenge_increment`) and correspondence games (`challenge_days`), the challenge will be a random choice between the two.
 
@@ -284,6 +285,7 @@ matchmaking:
   opponent_rating_difference: 100
   opponent_allow_tos_violation: true
   challenge_mode: "random"
+  delay_after_decline: false
 ```
 
 ## Lichess Upgrade to Bot Account
--- a/README.md
+++ b/README.md
@@ -260,10 +260,10 @@ will precede the `go` command to start thinking with `sd 5`. The other `go_comma
   - `opponent_rating_difference`: The maximum difference between the bot's rating and the opponent bot's rating.
   - `opponent_allow_tos_violation`: Whether to challenge bots that violated Lichess Terms of Service. Note that even rated games against them will not affect ratings.
   - `challenge_mode`: Possible options are `casual`, `rated` and `random`.
-  - `delay_after_decline`: Whether and how to delay challenging a bot after that bot declines a challenge. Options are "none", "coarse", and "fine".
-    - "none" does not delay challenging a bot that declined a challenge.
-    - "coarse" will delay challenging a bot to any type of game for a set time.
-    - "fine" will delay challenging a bot to the same kind of game that was declined for a set time.
+  - `delay_after_decline`: Whether and how to delay challenging a bot after that bot declines a challenge. Options are `none`, `coarse`, and `fine`.
+    - `none` does not delay challenging a bot that declined a challenge.
+    - `coarse` will delay challenging a bot to any type of game for a set time.
+    - `fine` will delay challenging a bot to the same kind of game that was declined for a set time.
 
 If there are entries for both real-time (`challenge_initial_time` and/or `challenge_increment`) and correspondence games (`challenge_days`), the challenge will be a random choice between the two.
 
--- a/README.md
+++ b/README.md
@@ -260,7 +260,10 @@ will precede the `go` command to start thinking with `sd 5`. The other `go_comma
   - `opponent_rating_difference`: The maximum difference between the bot's rating and the opponent bot's rating.
   - `opponent_allow_tos_violation`: Whether to challenge bots that violated Lichess Terms of Service. Note that even rated games against them will not affect ratings.
   - `challenge_mode`: Possible options are `casual`, `rated` and `random`.
-  - `delay_after_decline`: Whether to delay challenging a bot after that bot declines a challenge.
+  - `delay_after_decline`: Whether and how to delay challenging a bot after that bot declines a challenge. Options are "none", "coarse", and "fine".
+    - "none" does not delay challenging a bot that declined a challenge.
+    - "coarse" will delay challenging a bot to any type of game for a set time.
+    - "fine" will delay challenging a bot to the same kind of game that was declined for a set time.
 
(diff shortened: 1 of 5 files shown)


AttackingOrDefending:
I think that if there is no available bot we should challenge bots that have declined our challenges for the following reason: Let's say we challenge random bots with bullet and blitz time controls. The opponent may decline the challenge that we sent them because it is blitz, but if we had sent a bullet challenge they would have accepted it. It doesn't hurt attempting to challenge another bot first, since the opponent that declined our previous challenge may have `allow_bot` set to false, but if there is no available opponent I believe that we should take our chances and challenge them.

MarkZH:
To take an idea from the [issue](https://github.com/ShailChoksi/lichess-bot/issues/538), how about not adding a delay if the reason for the refusal is a time control mismatch?

AttackingOrDefending:
It will solve this problem but the same can be said about variant and mode (rated or casual), although time control is probably more common than the other two.

fabriciodosanjossilva:
@AttackingOrDefending , you are right. Considering my comment today on the original issue (https://github.com/ShailChoksi/lichess-bot/issues/538), the tuple to be suspended should also consider variant. Mode seems overwhelming, doesn't it?

AttackingOrDefending:
I think mode is still likely. The tuple in your idea should try to match any one of the characteristics, because at least in lichess-bot there is no way to decline specifically rated chess960 for example.

MarkZH:
Latest change: delay now considers the following parameters

 - bot name
 - game variant
 - time control speed (bullet, rapid, etc.)
 - rated/casual

If an opponent declines a challenge, lichess-bot will delay challenging that bot again with the same set of parameters.

fabriciodosanjossilva:
Seems promising. I'll test this.
Are you considering implementing strategy 2 as well? It would optimize/accelerate suspension when appropriate.

AttackingOrDefending:
Strategy 2 seems like an improvement but it's still not optimal, because we only send one reason we declined. We may challenge another bot for blitz antichess while they only accept bullet standard chess, but we will see that the challenge was declined only for time control, so strategy 2 won't solve all the problems.

I still think that we should keep the bots that declined us for variant, time control or mode reasons in a list (`set`) and only challenge them if there is no other available bot, like what I suggested in the first comment.

The problem with declining for variant reasons should have also been mostly fixed by #529.

Also, can you capitalize the first letter of the comments and end them with a period ([PEP8 on comments](https://peps.python.org/pep-0008/#comments)).

MarkZH:
Taking the idea from @AttackingOrDefending, if filtering opponents according to delay timers results in no available opponents, then a random online bot will be challenged (if it hasn't been blocked, if it has played at least one game of the variant, etc.).

As for @fabriciodosanjossilva's idea of parsing the reasons for declining a challenge, I don't see the point. Given that delays are determined by specific combinations of username, game variant, time control speed, and rating mode, challenging the same bot again but with a different time control will not be delayed. The only useful purpose I could see for analyzing the reason message is to see if it contains the message "I'm not accepting challenges from bots" or similar blanket rejections. A message like this could put the bot on the local block list so that it will not be challenged again. However, this message indicates a temporary state of affairs. I've seen opponents decline an offer with this message only to accept challenges a few hours later. This is unlike accounts that block the user's bot's account, which is a more permanent state of affairs.

fabriciodosanjossilva:
I agree. I have a feeling that we are discussing how to fix something that should not be fixed. The reason is that declining a challenge is part of the Lichess API, so the reason for a decline should be in the API. If present, clients (as lichess-bot) could respond appropriately. We are following a path of heuristics that will never be 100% correct.
What about requesting Lichess to include objective data in the decline event? Something parseable that lichess-bot could interpret and do precisely what should be done.

MarkZH:
I agree with keeping the latest change that chooses from all online bots in the case that the decline-delay filter results in no valid bots to choose from. I also do not want to parse the `declineReason` message.

I see two choices for this PR:

1. The current version (filter by bot name and game parameters): if a bot declines a challenge, do not challenge that bot with the specific combination of `(variant, time control, mode)` for an increasing amount of time.
2. A simpler version (filter by bot name): if a bot declines a challenge, do not challenge that bot to any games for an increasing amount of time.

Option 1 has the advantage of allowing for re-challenging a bot with different game settings immediately. Over time, a filter will be built up that will result in only challenging that bot with games that match its preferences. The disadvantage is that opponent bots with narrow game preferences or user bots with wide game preferences will result in a lot of declined challenges. This situation only marginally solves the original problem in issue #538.

Option 2 has the advantage of simplicity and a greatly reduced number of declined challenges. The disadvantage is a narrower field of bots to challenge, especially if the user's bot has a wide game parameter preference. As a compromise, perhaps the length of the delay increment could be shortened to something less than an hour.

MarkZH:
The current state of the PR is an implementation of option 1. I am increasingly leaning towards option 2.

fabriciodosanjossilva:
@MarkZH, feel free to decide. I'll go with you. My experience is still limited.

As a suggestion, could lichess-bot not enter "idle time" for "challenge_timeout" minutes when getting a decline? It could go right into searching a new opponent. Today this is not possible because there is no user suspension, so it could yield a challenge storm. When this feature is implemented it seems that it will be safe to do this.

AttackingOrDefending:
@MarkZH I also prefer option 2.

@fabriciodosanjossilva Are you talking about the 1 minute delay after we send each challenge? If so, this was added to not hit the API rate limits (error code 429).

fabriciodosanjossilva:
Please ignore my last suggestion. It was a mistake.

mheinzel:
For me, option 1 is something I'd use, but option 2 is not, as I'm making challenges from 1+0 to 20+10, so only few opponents support all of these time controls.
If we go with option 2, can we at least have a flag to turn it off in the config?

MarkZH:
Going with option 2 (delay based on username), but making it optional via the user's config.yml file. Default is no delay.

MarkZH:
Turns out I like both options, so I made it into a configuration.

AttackingOrDefending:
There are also bots that don't decline our challenges, so we cancel them. Should we also add them to the blacklist? This usually means that they have `concurrency` set to 0, so either the owner is present or they only want to challenge other bots and not be challenged.

MarkZH:
It could also mean that the bot is playing the maximum number of simultaneous games. The blacklist is for more permanent states such as the other bot blocking the user's bot account. Cancelling a challenge could be handled as if the challenge was declined, but I would be fine with doing nothing since the 25-second wait makes hitting API rate limits less likely.

AttackingOrDefending (approved):

PR #571, opened 2022-09-19
Simplify online DB functions

MarkZH:
1. Move large blocks of code to their own functions.
2. Create function for assigning score to each legal move on a board.
3. Create a piecewise function to replace if-elif-elif-else blocks that assign scores to ranges of input values.

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -628,8 +628,8 @@ def get_lichess_cloud_move(li, board, game, lichess_cloud_cfg):
     return move
 
 
-def range_scan(range_definition, last_value, position):
-    for border, value in range_definition:
+def piecewise_function(range_definitions, last_value, position):
+    for border, value in range_definitions:
         if position <= border:
             return value
     return last_value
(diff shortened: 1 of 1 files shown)


MarkZH:
Is `piecewise_function()` too obscure or complicated? Should I replace it with the `if-elif-else` blocks and just move them outside the functions to keep the flake8 complexity low enough?

AttackingOrDefending:
It is a little hard to understand on a first view, but I think a comment with an example will be good enough to explain it.
Something like:
```python
# piecewise_function([(-20001, 2), (-1, -1), (0, 0), (20000, 1)], 2, score)
#
# is equivalent to:
#
# if score < -20000:
#     return -2
# elif score < 0:
#     return -1
# elif score == 0:
#     return 0
# elif score <= 20000:
#     return 1
# else:
#     return 2
#
# Note: We use -20001 and not -20000, because we use <= and not <.
```
Only lines 980-990 should probably be replaced. The simple uses of `piecewise_function` are good enough in my opinion.

AttackingOrDefending (approved):

PR #572, opened 2022-09-21
Fix error when there is no ponderpv

AttackingOrDefending:
When moves were only played by online opening books, the bot will crash when responding to `!eval`.


MarkZH:
I merged to make sure the bug was fixed. One further change is to delete `"ponderpv" in stats` from that line since it's always true.

PR #575, opened 2022-09-24
Choose time control before choosing opponent

MarkZH:
If a user's config.yml file has both correspondence and real-time games specified, then the opponents will always be filtered based on correspondence game ratings. Then, there is a 50-50 chance that a real-time game challenge will be issued instead.

This change fixes this error by choosing which type of game to play before choosing an opponent.


MarkZH:
More changes incoming to fix flake8 issues.

MarkZH:
This is ready for review.

AttackingOrDefending (approved):

PR #580, opened 2022-10-12
Small fixes

MarkZH:
1. Delete two unused class members (one in `Game`, one in `Player`)
2. Move `move_attempted = True` to correct location in `play_game()`.


AttackingOrDefending:
I resigned while it was my bot's turn ([game](https://lichess.org/YAAXC2cP)) and the bot searched the same position continuously. My guess is that ths is happening because of the change in the location of `move_attempted`.

MarkZH:
Good catch. I think the latest commit should fix the problem by making sure that a new packet is retrieved from lichess for every main loop iteration.

AttackingOrDefending (approved):

PR #584, opened 2022-10-16
Restart bot if it is offline

AttackingOrDefending:
If the bot is offline, restart lichess-bot.

The logs provided by theoden8 (https://github.com/ShailChoksi/lichess-bot/issues/222#issuecomment-1280002750) show that the bot successfully restarted. While it is possible the `Resetting dropped connection` would have solved the problem by itself, it didn't for ianagbip1oti (https://github.com/ShailChoksi/lichess-bot/issues/222#issuecomment-1221612999). This makes me think that the change actually helped, so while this isn't conclusive evidence, it also doesn't seem to make the problem worse.

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -204,7 +204,7 @@ def lichess_bot_main(li,
                       logging_level]
 
     with multiprocessing.pool.Pool(max_games + 1) as pool:
-        while not terminated:
+        while not terminated and not restart:
             try:
                 event = control_queue.get()
                 if event.get("type") != "ping":
@@ -309,7 +309,6 @@ def lichess_bot_main(li,
                 if not li.is_online(user_profile["id"]):
                     logger.info("Will restart lichess-bot")
                     restart = True
-                    terminated = True
                 last_check_online_time.reset()
 
             control_queue.task_done()
@@ -641,7 +640,6 @@ if __name__ == "__main__":
     try:
         while restart:
             restart = False
-            terminated = False
             start_lichess_bot()
             time.sleep(10 if restart else 0)
     except Exception:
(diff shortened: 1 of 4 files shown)


lichess-bot.py
```diff
@@ -305,8 +307,9 @@ def log_proc_count(change, queued, used):
 
             if last_check_online_time.is_expired():
                 if not li.is_online(user_profile["id"]):
-                    logger.info("Will reset connection with lichess")
-                    li.reset_connection()
+                    logger.info("Will restart lichess-bot")
+                    restart = True
+                    terminated = True
```
MarkZH, line 312: Instead of setting `terminated = True` here, change the while-loop to `while not terminated and not restart:`. In the `while restart:` loop below, if the users pushes `Ctrl-C` during the 10-second sleep, it will be ignored since `terminated` will be set back to `False` once the sleep ends.

MarkZH:
I think the correct fix is to set the initial value of `restart` to `False` in lichess-bot.py on line 36. This change causes all tests to pass when I run them in Linux and Windows. It's correct because the function of `restart` is to stop and restart the program. The program doesn't need to be restarted when it first starts up.

MarkZH:
I don't understand your suggestion to add a `one_game` parameter to `start_lichess_bot()`. We don't call `start_lichess_bot()` from testing code since we create the `CONFIG` manually to pass to `start()`.

Since `one_game=True` is only used in testing, I don't see a problem with `lichess_bot.restart = False` in test_bot.py. Another option is to put this at the beginning of `start()`:
```python
    if one_game:
        global restart
        restart = False
```

MarkZH (approved):

PR #588, opened 2022-10-22
Suggest moves from EGTBs to the engine

AttackingOrDefending:
Suggests moves from EGTB to the engine. The moves that are suggested are all the moves that could have been chosen with `move_quality="good"`. This will fix the problem of EGTBs playing dumb moves (e.g. `8/8/8/8/1p6/1K1R4/8/2k5 w - - 0 1`), where the engine instead of mating will first pick up material. Getting the move from lichess will fix this problem for the most part because it also has DTM, but it doesn't have DTM for 6 or 7 men positions.

This can also be done with `lichess_cloud_analysis` but I don't know if this is useful.

Also had to move some code to fix the flake8 errors that the initial addition created.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,11 +88,12 @@ Besides the above, there are many possible options within `config.yml` for confi
     - Configurations common to all:
         - `enabled`: Whether to use the database at all.
         - `min_time`: The minimum time in seconds on the game clock necessary to allow the online database to be consulted.
-    - Configurations only in `chessdb_book` and `lichess_cloud_analysis`:
-        - `move_quality`: Choice of `"all"` (`chessdb_book` only), `"good"`, or `"best"`.
+        - `move_quality`: Choice of `"all"` (`chessdb_book` only), `"good"`, `"best"`, or `"suggest"` (`online_egtb` only).
             - `all`: Choose a random move from all legal moves.
             - `best`: Choose only the highest scoring move.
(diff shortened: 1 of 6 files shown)


MarkZH:
Xboard engines don't support suggestions to the engine. [The following lines are in chess.engine on line 2112](https://github.com/niklasf/python-chess/blob/c6e7e4961b32a7f4ad855598ce6d7103a76a7b32/chess/engine.py#L2112-L2114):
```python
    async def play(self, board: chess.Board, limit: Limit, *, game: object = None, info: Info = INFO_NONE, ponder: bool = False, draw_offered: bool = False, root_moves: Optional[Iterable[chess.Move]] = None, options: ConfigMapping = {}) -> PlayResult:
        if root_moves is not None:
            raise EngineError("play with root_moves, but xboard supports 'include' only in analysis mode")
```
Perhaps a check could be added to `load_config()` in config.py to make sure that the `"suggest"` value is never used with `protocol: xboard`.

MarkZH:
Using the square bracket operator in these checks causes errors if the sections are not in the user's config file. I've attached a patch with one way to fix this and cut down on some of the code repetition.

[config_xboard.txt](https://github.com/ShailChoksi/lichess-bot/files/9866869/config_xboard.txt)

MarkZH (approved):

PR #593, opened 2022-10-29
Decline bots that challenge too often

bec-ca:
This is a minimum prototype. Sorry this is very raw but I don't have much time right now and I would like to know if you think this direction makes sense. Thanks!

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -385,8 +385,8 @@ def enough_time_to_queue(event, config):
 def handle_challenge(event, li, challenge_queue, challenge_config, user_profile, matchmaker, recent_bot_challenges):
     chlng = model.Challenge(event["challenge"], user_profile)
 
-    time_window = challenge_config.get('recent_bot_challenge_age')
-    max_recent_challenges = challenge_config.get('max_recent_bot_challenges')
+    time_window = challenge_config.recent_bot_challenge_age
+    max_recent_challenges = challenge_config.max_recent_bot_challenges
 
     is_supported, decline_reason = chlng.is_supported(challenge_config)
 
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -219,7 +219,7 @@ def lichess_bot_main(li,
                 log_proc_count("Freed", active_games)
                 one_game_completed = True
             elif event["type"] == "challenge":
-                handle_challenge(event, li, challenge_queue, challenge_config, user_profile, matchmaker, recent_bot_challenges, config)
+                handle_challenge(event, li, challenge_queue, challenge_config, user_profile, matchmaker, recent_bot_challenges)
             elif event["type"] == "challengeDeclined":
                 matchmaker.declined_challenge(event)
             elif event["type"] == "gameStart":
@@ -387,7 +387,7 @@ def enough_time_to_queue(event, config):
     return not game["isMyTurn"] or game.get("secondsLeft", math.inf) > minimum_time
 
 
-def handle_challenge(event, li, challenge_queue, challenge_config, user_profile, matchmaker, recent_bot_challenges ):
+def handle_challenge(event, li, challenge_queue, challenge_config, user_profile, matchmaker, recent_bot_challenges):
     chlng = model.Challenge(event["challenge"], user_profile)
 
     time_window = challenge_config.get('recent_bot_challenge_age', None)
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -384,25 +384,14 @@ def enough_time_to_queue(event, config):
 
 def handle_challenge(event, li, challenge_queue, challenge_config, user_profile, matchmaker, recent_bot_challenges):
     chlng = model.Challenge(event["challenge"], user_profile)
-
-    time_window = challenge_config.recent_bot_challenge_age
-    max_recent_challenges = challenge_config.max_recent_bot_challenges
-
-    is_supported, decline_reason = chlng.is_supported(challenge_config)
-
-    if is_supported and chlng.challenger_is_bot and time_window is not None and max_recent_challenges is not None:
-        op = chlng.challenger_name
-        # Filter out old challenges
-        recent_bot_challenges[op] = [t for t in recent_bot_challenges[op] if not t.is_expired()]
-        if len(recent_bot_challenges[op]) >= max_recent_challenges:
-            is_supported = False
-            decline_reason = "later"
-        else:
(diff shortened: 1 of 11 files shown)


MarkZH:
This makes sense to me. Keep a list of recent challengers and the time that each challenge was received. If a bot makes too many challenges within a day, decline the challenge. Remove challenges that are more than a day old before counting challenges.

Here are some suggestions for improvements:

---

Instead of keeping all of the challenges in a list, use a dictionary that associates a bot's username with a list of challenge timestamps. Instead of the current line 374
```python
recent_challenges.append([now, chlng.challenger_name])
```
use
```python
recent_challenges[chlng.challenger_name].append(now)
```
Then, you can replace the `for`-loop on line 360 with a lookup:
```python
num_recent_challenges = len(recent_challenges[chlng.challenger_name])
```
 
Use a [`defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict) instead of a normal dictionary. This way, if you use `recent_challenges = defaultdict(list)`, you don't have to check if an entry for a username exists before calling `append()`.

---

Instead of doing time calculations yourself, use the `Timer` class in the file timer.py. You create a day-long timer with `timer = Timer(time_window)` and check if a day has passed by calling `timer.is_expired()`. With this, the append on line 374 becomes
```python
recent_challenges[chlng.challenger_name].append(Timer(time_window))
```

---

For getting rid of expired timers, you can use the built-in function [`filter()`](https://docs.python.org/3/library/functions.html#filter).
```python
recent_challenges[chlng.challenger_name] = list(filter(lambda t: t.is_expired(), recent_challenges[chlng.challenger_name]))
```

---

I don't think you need both the number of challenges and the time window in the configuration file. You could have a single setting called something like `max_daily_challenges` in the `challenge` section to let the user set the maximum number of challenges in a 24-hour period from any one bot.

mheinzel:
Just for clarification: The title talks about challenges from bots, but it doesn't seem the code distinguishes between bot or human opponents. Is this intended to apply to bot challenges only, or also humans?

Humans could theoretically also farm rating, but from my experience that's seems unlikely unless they found a specific weakness in your bot. They can also keep your bot from playing against others, but their time and patience is limited and I usually highly value any match my bot gets against a human opponent.

bec-ca:
Thanks for the feedback, I refactored the code to incorporate the feedback.

@MarkZH  Allowing to specify the time window gives the bot operator more control. E.g., limiting to 24 games per day is not the same as 1 game per hour. The former allows the bot to play 24 games in a row as fast as it possibly can, but then has to wait for the next day to play again. While the latter will require the bot to wait one hour between each game. I think I would like to set the span to less than a day for my bot. One could argue that doing what you suggested is good enough for most use cases, I'm fine with that as well. What do you think?

@mheinzel I realize the code wasn't clear. I only added to the list of recent challenges if the challenger was a bot, i moved that check earlier to make that clearer.

Thanks!

CodeBuddyUltra (approved):

MarkZH:
@bec-ca Write the time window settings how you want. Make sure the documentation in config.yml.default and README are clear.

MarkZH:
@bec-ca How are you doing with this PR? Do you need any help?

bec-ca:
Thanks for checking @MarkZH , I just have been too busy with other stuff so I didn't have time to look at this again, but I should have some time in the next couple of days if you don't mind waiting on me.

bec-ca:
It took me a while, but this PR should be ready for review.

lichess-bot.py
```diff
@@ -385,9 +387,24 @@ def enough_time_to_queue(event, config):
     return not game["isMyTurn"] or game.get("secondsLeft", math.inf) > minimum_time
 
 
-def handle_challenge(event, li, challenge_queue, challenge_config, user_profile, matchmaker):
+def handle_challenge(event, li, challenge_queue, challenge_config, user_profile, matchmaker, recent_bot_challenges):
     chlng = model.Challenge(event["challenge"], user_profile)
+
+    time_window = challenge_config.get('recent_bot_challenge_age', None)
```
MarkZH, line 393: The `None` argument here is redundant.

MarkZH (approved):

MarkZH:
I've been thinking about this more. Does it make more sense for the recent challenges check to be done in `Challenge.is_supported()`? This way, the `is_supported, reason` tuple is created with all the other checks, and the side effect of adding a time can be put with the other side effects [here](https://github.com/ShailChoksi/lichess-bot/blob/9058e94acf8e62936b7a2655b006c778ef9c734c/lichess-bot.py#L391).

AttackingOrDefending:
I agree that putting it in `Challenge.is_supported()` is better. Also, can you change `op` and `t` to something like `challenger` and `timer` respectively, so that it is easier to understand what these variables do.

MarkZH:
In case it isn't clear what @AttackingOrDefending and I are suggesting, here's one way to go about it:

In `Challenge.is_supported()` in model.py, put these lines just before the final return
```python
def is_supported(self, config):

    # ...

    if self.challenger_name in (config.get("block_list") or []):
                return False, "generic"

    time_window = config.get('recent_bot_challenge_age')
    max_recent_challenges = config.get('max_recent_bot_challenges')

    if chlng.challenger_is_bot and time_window is not None and max_recent_challenges is not None:
        op = chlng.challenger_name
        # Filter out old challenges
        recent_bot_challenges[op] = [t for t in recent_bot_challenges[op] if not t.is_expired()]
        if len(recent_bot_challenges[op]) >= max_recent_challenges:
            return False, "later"

    return True, None
```
And then, in `handle_challenges()` in lichess-bot.py
```python
    if is_supported:
        challenge_queue.append(chlng)
        sort_challenges(challenge_queue, challenge_config)
        time_window = config.get('recent_bot_challenge_age')
        recent_bot_challenges[op].append(Timer(time_window))
    elif chlng.id != matchmaker.challenge_id:
        li.decline_challenge(chlng.id, reason=decline_reason)
```

MarkZH:
One thing that will need to change with the merging of PR #608, is the retrieval of configuration values. For example, the line
```python
    time_window = config.get('recent_bot_challenge_age')
```
will need to change to
```python
    time_window = config.recent_bot_challenge_age
```
and similarly for other configuration parameters.

bec-ca:
Ok, this is update with @MarkZH and @AttackingOrDefending suggestions.Thank you for the view!

MarkZH (approved):

AttackingOrDefending (approved):

PR #594, opened 2022-10-29
Handle correspondence games with little time left

MarkZH:
During correspondence games, the time left on the clock is generally ignored since the latter is measured in days. However, if lichess-bot is started when a correspondence game clock is about to expire, the delay caused by queuing plus the move time in the user's config file can result in a bot losing on time. This is fixed in two parts:
- On startup: if it is the bot's move and the time on the bot's clock is less than ten times the sum of the `correspondence: checkin_period` and `correspondence: move_time`, then start the game immediately instead of queuing.
- In a game: if the time left on the bot's clock is less than the `correspondence: move_time` in the user's config file, use the game clock.

One possible problem with this solution is that it doesn't respect concurrency limits. It's possible to have many more correspondence games going simultaneously because they aren't all played simultaneously. With this PR, an arbitrary number of games may start up that is well beyond the user's concurrency limit.

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -136,7 +136,10 @@ class EngineWrapper:
             elif is_correspondence:
                 best_move = choose_move_time(self,
                                              board,
+                                             game,
                                              correspondence_move_time,
+                                             start_time,
+                                             move_overhead,
                                              can_ponder,
                                              draw_offered)
             else:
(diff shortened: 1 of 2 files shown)


MarkZH:
Previous concern fixed by putting correspondence games with little time left in a list. Games in this list are started on every iteration of the `lichess_bot_main()` loop when there are free processes in the thread pool.

AttackingOrDefending:
When someone plays a correspondence from position game, the bot will think that the game isn't correspondence, because in line 432 in lichess-bot.py `is_correspondence = game.perf_name == "Correspondence"` is false because `perf_name = "{perf?}"`. Changing it to `is_correspondence = game.speed == "correspondence"` fixes the problem. This bug wasn't introduced in this PR, but I discovered it while testing it.

AttackingOrDefending (approved):

PR #596, opened 2022-10-30
Allow to block users from challenges and matchmaking

alex65536:


alex65536:
The reason why this patch appeared is somehow similar to #593, but this one uses an explicit list to block players from challenges. I hope that both patches will be merged :)

MarkZH:
This can also be achieved by blocking players on the lichess website. Do you think it's common to want to block challenges from a player but still issue challenges to that same player (or vice versa)?

alex65536:
> This can also be achieved by blocking players on the lichess website.

Do I understand correctly that blocking users on the website prevents from both accepting the challenges from the user and sending them?

Excluding a user for matchmaking, by the way, makes sense. There was a bot author who PM'ed me because he thought that my bot sends him challenges to often. So, I just added him to a blocklist on the server side. Not sure whether it's important to accept challenges from such bots, though.

MarkZH:
I just tried an experiment. If account A blocks account B on lichess, then B cannot challenge A. But, A can still challenge B. So, a block list in the matchmaker section makes sense. The question now is do we want to replicate the blocking feature on the lichess website. I'll need to think about this. @AttackingOrDefending any thoughts?

In the meantime, documentation of this configuration is needed in the README file.

matchmaking.py
```diff
@@ -25,7 +26,8 @@ def __init__(self, li, config, user_profile):
         self.last_user_profile_update_time = Timer(5 * 60)  # 5 minutes
         self.min_wait_time = 60  # Wait 60 seconds before creating a new challenge to avoid hitting the api rate limits.
         self.challenge_id = None
-        self.block_list = []
+        self.block_list = copy.copy(self.matchmaking_cfg.get("block_list")) or []
+        assert isinstance(self.block_list, list)
```
MarkZH, line 30: A better place for this check is the `load_config()` function in config.py.

model.py
```diff
@@ -63,6 +63,9 @@ def is_supported(self, config):
             if not self.is_supported_mode(config):
                 return False, ("casual" if self.rated else "rated")
 
+            if self.challenger_name in config.get("block_list", []):
+                return False, "blocked"
```
MarkZH, line 67: The message `"blocked"` is not a valid reason for declining a challenge. See the list of possible reasons in the [Lichess API documentation](https://lichess.org/api#tag/Challenges/operation/challengeDecline). The translation of each reason value can be found in [this file from the lila repository](https://github.com/lichess-org/lila/blob/b0ab35f98e3b535fe93ac653140d77abfa950a65/translation/source/challenge.xml).

AttackingOrDefending:
I think it is a good idea to have a blocklist.

matchmaking.py
AttackingOrDefending: I think it would be better to use `self.block_list = (self.matchmaking_cfg.get("block_list") or []).copy()`, so we don't have to import `copy`.

README.md
```diff
@@ -235,6 +235,7 @@ will precede the `go` command to start thinking with `sd 5`. The other `go_comma
     -rated
     -casual
 ```
+  - `blocklist`: An indented list of usernames from which the challenges are always declined. If this option is not present, then the list is considered empty.
```
MarkZH, line 238: This should be `block_list` to match the config file. Same in the matchmaking section.

model.py
```diff
@@ -63,6 +63,9 @@ def is_supported(self, config):
             if not self.is_supported_mode(config):
                 return False, ("casual" if self.rated else "rated")
 
+            if self.challenger_name in config.get("block_list", []):
```
MarkZH, line 66: It's better to use `config.get("block_list") or []` to make sure a list is the result. If a user has `block_list:` with nothing beneath it, the current expression will return `None`, leading to an error since `X in None` is not valid. Imagine that the user has this in their config.yml:
```yml
challenges:
  block_list:
    - user1
    - user2
```
Then, the user decides to delete the list.

```yml
challenges:
  block_list:
```
This will cause the error.

alex65536:
Is there something else needed to be done in this PR before merging?

MarkZH:
Probably not. We just need to find the time to review and test.

AttackingOrDefending (approved):

MarkZH (approved):

PR #600, opened 2022-11-20
Allow constraining time control for correspondence challenges

mheinzel:
This is a feature I added for my fork, maybe you think it makes sense for others as well. My bot has an increasing number of ongoing correspondence games, many unlimited and ignored by their opponents or progressing super slowly (1 move every 2 weeks). This can be avoided by only accepting challenges up to a small number of days per move, which also excludes unlimited time control.

I included `min_days` for symmetry, but can't really imagine a use case. Happy to remove it if you think that's better.

**Some notes on the code:**

Introducing the `time_control_type` tag seemed like the best way to handle the unlimited case, so I made use of it to distinguish clock and correspondence as well.

I think with these checks the `-1` sentinel values should not be needed anymore, unless something is really weird with the API. And if it is, silently skipping those challenges doesn't sound like the best strategy either.
I could add some extra checks that log an error and decline the challenge if an expected field is missing, though.

Relevant parts of some challenge JSON I captured:

```json
    "speed": "bullet",
    "timeControl": {
      "type": "clock",
      "limit": 60,
      "increment": 0,
      "show": "1+0"
    },
```

```json
    "speed": "correspondence",
    "timeControl": {
      "type": "correspondence",
      "daysPerTurn": 2
    },
```

```json
    "speed": "correspondence",
    "timeControl": {
      "type": "unlimited"
    },
```

### DIFF
--- a/config.yml.default
+++ b/config.yml.default
@@ -112,7 +112,8 @@ challenge:                   # Incoming challenges.
   min_increment: 0           # Minimum amount of increment to accept a challenge.
   max_base: 315360000        # Maximum amount of base time to accept a challenge. The max is 315360000 (10 years).
   min_base: 0                # Minimum amount of base time to accept a challenge.
-  max_days:                  # Maximum number of days per move to accept a challenge for a correspondence game.
+  max_days: 14               # Maximum number of days per move to accept a challenge for a correspondence game.
+                             # Unlimited games can be accepted by removing this field or specifying .inf
   min_days: 1                # Minimum number of days per move to accept a challenge for a correspondence game.
   variants:                  # Chess variants to accept (https://lichess.org/variant).
     - standard
--- a/config.yml.default
+++ b/config.yml.default
@@ -112,6 +112,8 @@ challenge:                   # Incoming challenges.
   min_increment: 0           # Minimum amount of increment to accept a challenge.
   max_base: 315360000        # Maximum amount of base time to accept a challenge. The max is 315360000 (10 years).
   min_base: 0                # Minimum amount of base time to accept a challenge.
+  max_days:                  # Maximum number of days per move to accept a challenge for a correspondence game.
+  min_days: 1                # Minimum number of days per move to accept a challenge for a correspondence game.
   variants:                  # Chess variants to accept (https://lichess.org/variant).
     - standard
 #   - fromPosition
--- a/README.md
+++ b/README.md
@@ -212,6 +212,8 @@ will precede the `go` command to start thinking with `sd 5`. The other `go_comma
   - `min_increment`: The minimum value of time increment.
   - `max_base`: The maximum base time for a game.
   - `min_base`: The minimum base time for a game.
+  - `max_days`: The maximum number of days for a correspondence game.
+  - `min_days`: The minimum number of days for a correspondence game.
   - `variants`: An indented list of chess variants that the bot can handle.
 ```yml
   variants:
--- a/model.py
+++ b/model.py
@@ -12,7 +12,6 @@ class Challenge:
         self.variant = c_info["variant"]["key"]
         self.perf_name = c_info["perf"]["name"]
         self.speed = c_info["speed"]
-        self.time_control_type = c_info.get("timeControl", {}).get("type")
         self.increment = c_info.get("timeControl", {}).get("increment")
         self.base = c_info.get("timeControl", {}).get("limit")
         self.days = c_info.get("timeControl", {}).get("daysPerTurn")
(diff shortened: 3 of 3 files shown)


config.yml.default
```diff
@@ -112,6 +112,8 @@ challenge:                   # Incoming challenges.
   min_increment: 0           # Minimum amount of increment to accept a challenge.
   max_base: 315360000        # Maximum amount of base time to accept a challenge. The max is 315360000 (10 years).
   min_base: 0                # Minimum amount of base time to accept a challenge.
+  max_days:                  # Maximum number of days per move to accept a challenge for a correspondence game.
+  min_days: 1                # Minimum number of days per move to accept a challenge for a correspondence game.
```
mheinzel, line 116: I wasn't quite sure about the best values to put here. 1 is a reasonable default for the minimum, but it seemed wrong to put some arbitrary number to represent "unlimited" or "no maximum". Leaving it empty is interpreted as no upper bound (allowing unlimited games), just as not putting a `max_days` field at all.

I could have also commented them out.

mheinzel:
> I think with these checks the `-1` sentinel values should not be needed anymore

On a more general note, these arbitrary default values for fields seem like a code smell and source of bugs. Instead of failing in an obvious way, they silently go wrong. A correspondence game does not have an initial time. Giving it one can only lead to confusion.

There is a similar thing in the Game class. This even caused a bug ([fixed recently](https://github.com/ShailChoksi/lichess-bot/pull/594/commits/5494df54b4d9639923390471947f32e6a73140af)) that made my bot search for a move for up to 10 years. It wasted more than a day of CPU time before I noticed.

The approach I chose is not necessarily the best, but there must be something better than the current solution.

AttackingOrDefending:
I think that a value of 100 days of more, should be considered unlimited, so that users can change their configs easily, because I don't think that leaving the value empty is intuitive. Internally we use 10 years to denote the maximum time, so this is also another alternative.

Also, I believe that the condition for the parameter to be considered unlimited, whatever that is, should be mentioned in the `README` and `config.yml`.

MarkZH:
This is a good feature to add.

I think `min_days` should be kept. I run my bot on my personal computer, so setting `min_days: 2` would give me some room to not lose on time if my computer is off for a day.

I think  a reasonable default for `max_days` is 14--the longest finite correspondence time control. This prevents the problem of games that never end as described in the OP. To allow unlimited games, the [YAML spec allows the value `.inf` for infinity](https://yaml.org/spec/1.2.2/#10214-floating-point), which would work with both correspondence games and unlimited games. The test for accepting an unlimited game would be `return math.isinf(days_max)`. This could also be used for `max_base` and `max_increment`, but that's another PR.

As I mentioned in [the other PR](https://github.com/ShailChoksi/lichess-bot/pull/601#issuecomment-1321222433), you're right about the need for more structured handling of JSON messages from lichess.

mheinzel:
> I think that a value of 100 days of more, should be considered unlimited

I get where you're coming from, but picking arbitrary numbers like "10 years in milliseconds" seems unintuitive and just waiting to go wrong. Having it as a constant with a helpful comment at the top of the file would make it better (more discoverable), but still not ideal.

> To allow unlimited games, the [YAML spec allows the value .inf for infinity](https://yaml.org/spec/1.2.2/#10214-floating-point)

I like this better as a default, wherever we want to encode "no limit". This is strictly more meaningful and easier to check for than saying "there is a limit, but it's 10 years, so not really a limit in practice". Replacing other config defaults with `math.inf` seems promising to me, but either way we'll be stuck with the current values in existing config files copied from the template.

Maybe a lesson from that is that it's best to encourage users to remove/comment out a field if they want the default? We need to handle that case anyways, so why come up with a second way to encode it that we're then stuck with?

mheinzel:
> That would change the bot's behaviour for everyone upgrading. It seems like a better default, but is it not nicer to keep the previous behaviour as a default and just allow users to opt into the new limits?

I think I understand now. In model.py we want `math.inf` as fallback if the value is not specified (so including for existing users with existing config files). In config.yml.default we want `14` as a reasonable default for new users or people purposely copying it.

> Introducing the time_control_type tag seemed like the best way to handle the unlimited case

I got rid of that as it was kind of redundant, but I stand by my other choices to:

1. use `None` instead of `-1` for missing fields in the Challenge constructor.
   It is closer to how the Lichess API works and makes it easier to distinguish clock vs. correspondence vs. unlimited.
2. handling the three time control cases separately.
   It is more verbose, but makes it clearer what the actual conditions are that should be checked, and that each case should be considered by someone changing the code.

mheinzel:
By the way, while checking the Lichess source code I saw that [even unlimited games do get aborted after some time of inactivity], which is currently at [21 days](https://github.com/lichess-org/lila/blob/2898a23ffe2c59c73a55d6e5d6e136830958bc88/modules/game/src/main/Game.scala#L717).

A bit weird, but also makes sense. At least that means I won't have to resign those games to get rid of them.

MarkZH (approved):

AttackingOrDefending (approved):

PR #601, opened 2022-11-20
Fix for missing secondsLeft field in unlimited games

mheinzel:
This bug was introduced in #594. Unlimited games don't contain a `secondsLeft` field, which causes an exception.

I made a minimal change to get the bot working again, but the correctness still hinges on the invariant that only games with a `secondsLeft` entry are put into the `low_time_games` queue, which is not obvious and might break again in the future.

It is another case of certain fields not being present on all types of games or challenges (clock/correspondence/unlimited), but not distinguishing between them in a principled way. Is there interest in a discussion on how to prevent similar issues in the future? Be it through improved testing or changes in the code, but I think the project could benefit from tackling this source of bugs.

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -381,7 +381,10 @@ def enough_time_to_queue(event, config):
     move_time = corr_cfg.get("move_time") or 60
     minimum_time = (checkin_time + move_time) * 10
     game = event["game"]
-    return not game["isMyTurn"] or game["secondsLeft"] > minimum_time
+    seconds_left = game.get("secondsLeft")
+    return (not game["isMyTurn"]
+            or seconds_left is None
+            or seconds_left > minimum_time)
 
 
 def handle_challenge(event, li, challenge_queue, challenge_config, user_profile, matchmaker):
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -17,6 +17,7 @@ import sys
 import os
 import io
 import copy
+import math
 from config import load_config
 from conversation import Conversation, ChatLine
 from timer import Timer
@@ -302,7 +303,7 @@ def check_in_on_correspondence_games(pool,
 
 
 def start_low_time_games(low_time_games, active_games, max_games, pool, play_game_args):
-    low_time_games.sort(key=lambda g: g["secondsLeft"])
+    low_time_games.sort(key=lambda g: g.get("secondsLeft", math.inf))
     while low_time_games and len(active_games) < max_games:
         game_id = low_time_games.pop(0)["id"]
         active_games.add(game_id)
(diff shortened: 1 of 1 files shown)


AttackingOrDefending:
I agree that we should do something to avoid similar problems. We can add a comment in line 384 explaining that this is done because unlimited games don't have `secondsLeft`. We can also common bugs in `CONTRIBUTING.md`, so that contributors check that they haven't done any of these common mistakes. Adding one test for correspondence and one for unlimited, shouldn't be that hard. We will just have to capture the messages from lichess in these types of games and add them in `test_bot/lichess.py` and lower `correspondence.move_time` in the config.

MarkZH:
The fix in this PR is good. A comment indicating that `seconds_left is None` means that the game is unlimited would be useful as well.

I agree with the need for handling messages from lichess in a more structured way. I don't have any clear ideas right now. We could start with a function that unifies all of the `(event.get("thing") or {}).get("thing2") or default_value` logic in one place. Something like
```python
def get_lichess_value(message, *args, default=None):
    section = message
    for parameter in args[:-1]:
        section = section.get(parameter) or {}
    value = section.get(args[-1])
    if value is None and default is None:
        raise ValueError(f'No value in {message} for {args}')
    return value if value is not None else default
```
This would change [this line from model.py](https://github.com/ShailChoksi/lichess-bot/blob/ea4ea5619d472745ce8c9c28591cd238015712c2/model.py#L102) from
```python
        self.perf_name = (json.get("perf") or {}).get("name", "{perf?}")
```
to
```
        self.perf_name = get_lichess_value (json, "perf", "name", default="{perf?}")
```
This way, default values are clearly labeled and all of the logic happens in one place. If the API changes, we'll get error messages in a format that is easy to interpret.

mheinzel:
After looking at the other PR, I also applied the idea of using `math.inf` here. I think it's fine in this case, using it in such a small scope.

> I agree with the need for handling messages from lichess in a more structured way.

My question was meant to be about missing/placeholder values in general, including in our own classes. What are good conventions/policies there, e.g. where default/placeholder values should be used and where not?

Some examples of what such a convention could look like (without putting too much thought into it):

1. For config values, all missing values should be replaced by default values (0/infinity/...) in a central place, so code using them can always assume they are there (not None) and there's a single source of truth.
   (Currently we have `get()` all over the place and e.g. `checkin_period`'s default of 600 is currently duplicated to two places in the code)

2. For other values that somehow do not apply (e.g. base time for correspondence), they are left None. Any piece of code relying on them should check if they are None. For example, distinguish the clock, correspondence and unlimited case and handle each separately.

Having such a convention clearly stated can go a long way to make the code more consistent and understandable, avoiding bugs.

MarkZH (approved):

PR #608, opened 2022-12-03
Config defaults

MarkZH:
Set all configuration default values in a single location.

### DIFF
--- a/config.py
+++ b/config.py
@@ -93,6 +93,7 @@ def insert_default_values(CONFIG):
     set_config_default(CONFIG, "challenge", key="max_days", default=math.inf)
     set_config_default(CONFIG, "challenge", key="min_days", default=1)
     set_config_default(CONFIG, "challenge", key="block_list", default=[])
+    set_config_default(CONFIG, "correspondence", key="checkin_period", default=600)
     set_config_default(CONFIG, "matchmaking", key="challenge_timeout", default=30, force_falsey_values=True)
     CONFIG["matchmaking"]["challenge_timeout"] = max(CONFIG["matchmaking"]["challenge_timeout"], 1)
     set_config_default(CONFIG, "matchmaking", key="block_list", default=[], force_falsey_values=True)
--- a/config.py
+++ b/config.py
@@ -95,6 +95,8 @@ def insert_default_values(CONFIG):
     set_config_default(CONFIG, "matchmaking", key="opponent_min_rating", default=600, force_falsey_values=True)
     set_config_default(CONFIG, "matchmaking", key="opponent_max_rating", default=4000, force_falsey_values=True)
     set_config_default(CONFIG, "matchmaking", key="opponent_allow_tos_violation", default=True)
+    set_config_default(CONFIG, "matchmaking", key="challenge_variant", default="random")
+    set_config_default(CONFIG, "matchmaking", key="challenge_mode", default="random")
 
 
 def log_config(CONFIG):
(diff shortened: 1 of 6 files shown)


MarkZH:
This is a work-in-progress to show an idea before too many changes happen.

MarkZH:
This completes engine_wrapper.py. Thoughts? Comments? Good idea/bad idea?

MarkZH:
So, after all that, I might have a better idea.

Instead of fixing up nested dictionaries with defaults and all the `.get()`ing they require everywhere else, we write a `Configuration` class with members representing all the options we care to retrieve. For example, instead of checking whether the bot accepts challenges from other bots by calling
```python
if not config.get("challenge").get("accept_bot") and self.challenger_is_bot:
```
we write
```python
if not config.accept_challenges_from_bots and self.challenger_is_bot:
```
Within the `Configuration.__init__()` method, we put the complicated nesting dictionary code
```python
class Configuration:
    def __init__(self, yaml_input):
        # ...
        self.accept_challenges_from_bots = yaml_input["challenge"].get("accept_bot") or False
```
This is similar to the way the classes in model.py (`Challenge`, `Game`, and `Player`) work.

Thoughts?

AttackingOrDefending:
I think it is a good idea do add a class like that, but I would prefer to keep the structure of `config.yml` because then we would have 50+ options because we care about all the variables of `config.yml`. What we can do is read `config.yml.default` and update it with the values of `config.yml`, so we have all the needed values. Then we can create a class and set `__getattr__` if we want to use dot notation or just use the dict normally with brackets and not `get()`.

mheinzel:
> What we can do is read config.yml.default and update it with the values of config.yml, so we have all the needed values.

I think the values in config.yml.default are useful as a template for users to start from, but not necessarily what we want to use then nothing is specified in the config. See for example [a related comment](https://github.com/ShailChoksi/lichess-bot/pull/600#issuecomment-1321322218).

> Thoughts?

I like the idea. And if I understand it correctly, it also wouldn't require more boilerplate than what is in this PR.

The main question I'm still unsure about is about the best way to handle nested keys like `set_config_default(CONFIG, "engine", key="silence_stderr", default=False)`.
If we squash this into a single field name like "engine_silence_stderr", the mapping between fields and the YAML structure is not quite as clear anymore and (more importantly), we lose the ability to pass subsets of the config into specific parts of the code.
Most of the code only needs some part (e.g. the matchmaking engine part), so it should not have access to the rest. This helps me understand code, since I immediately know that the other options cannot play a role in a function's behaviour.

So do we want some additional classes like `EngineConfig` which get nested into the main `Config`, like `self.engine = EngineConfig(yaml_input)`?

AttackingOrDefending:
> So do we want some additional classes like `EngineConfig` which get nested into the main `Config`, like `self.engine = EngineConfig(yaml_input)`?

We can use something like:
```python
class Configuration:
    def __init__(self, yaml_input):
        self.yaml_input = yaml_input

    def __getattr__(self, attr):
        value = self.yaml_input[attr]
        if isinstance(value, dict):
            return Configuration(value)
        return value
```
To get a subconfig or a value.

> I think the values in config.yml.default are useful as a template for users to start from, but not necessarily what we want to use then nothing is specified in the config.

I still think that it is worth it. It would simplify the code and we won't have to add every new parameter in `config.yml.default` and the new config class. The change in behavior for unlimited games isn't that big of a change. We could also add deprecation warnings for these changes they are important.

MarkZH:
I created the `Configuration` class along the lines described by @AttackingOrDefending. For now, I've kept the code that sets defaults as is instead of reading config.yml.default for two reasons:
1. The defaults in the code are different from what's in the file.
2. I want to guard against the program having errors due to users renaming config.yml.default instead of copying it.

P.S. I've come to dislike pickles (and [pickle](https://docs.python.org/3/library/pickle.html)) more than I already have. It took way to long to figure out that my class needed to have `__getstate__` and `__setstate__` methods due to the `__getattr__` method interfering with pytest trying to pickle everything.

AttackingOrDefending (approved):

PR #611, opened 2022-12-15
Ensure proper termination of sub-processes before attempting to restart

noobpwnftw:


noobpwnftw:
Not sure how to resmultiprocessing

MarkZH:
The test fails because `restart` is initially set to `True` and is only set to `False` in the `if __name__ == "__main__":` section.

Issue #607, opened 2022-12-02
Bot randomly stops matchmaking and accepting requests


Quintium:
A day or two after starting, the bot hangs and stops matchmaking and accepting requests. This was already reported in #573 , but the author marked it as closed after an update had apparently fixed the bug. The bug stills happens to me even after re-cloning the repo several times.

Here's the verbose logs:
```                      
[11/30/22 22:35:06] DEBUG    <UciProtocol (pid=15357)>: >> info      engine.py:1035
                             score cp -1288 depth 1 nodes 71 time 33               
                             nps 2113 pv                                           
                    DEBUG    <UciProtocol (pid=15357)>: >> info      engine.py:1035
                             score cp -1288 depth 2 nodes 203 time                 
                             34 nps 5867 pv                                        
                    DEBUG    <UciProtocol (pid=15357)>: >> info      engine.py:1035
                             score cp -1399 depth 3 nodes 1208 time                
                             40 nps 29499 pv                                       
                    DEBUG    <UciProtocol (pid=15357)>: >> info      engine.py:1035
                             score cp -1807 depth 4 nodes 7769 time                
                             61 nps 126721 pv                                      
                    DEBUG    <UciProtocol (pid=15357)>: >> info      engine.py:1035
                             score cp -1932 depth 5 nodes 12905 time               
                             88 nps 145648 pv                                      
                    DEBUG    <UciProtocol (pid=15357)>: >> info      engine.py:1035
                             score cp -2458 depth 6 nodes 26965 time               
                             128 nps 210047 pv                                     
                    DEBUG    <UciProtocol (pid=15357)>: >> info      engine.py:1035
                             score mate -3 depth 7 nodes 42325 time                
                             200 nps 211608 pv d1d7 a4d7 g7g6 g5g6                 
                             g8h8 g6g7                                             
                    DEBUG    <UciProtocol (pid=15357)>: >> bestmove  engine.py:1035
                             d1d7                                                  
                    INFO     Evaluation: #-3                  engine_wrapper.py:264
                    INFO     Depth: 7                         engine_wrapper.py:264
                    INFO     Nodes: 42.3K                     engine_wrapper.py:264
                    INFO     Speed: 211.6Knps                 engine_wrapper.py:264
                    INFO     PV: 69...Rd7+ 70. Bxd7 g6 71.    engine_wrapper.py:264
                             Qxg6+ Kh8 72. Qg7#                                    
[...]

PR #612, opened 2022-12-16
Ensure proper termination of sub-processes before attempting to restart

noobpwnftw:


noobpwnftw:
There were three problems that should be fixed now:
1. Timeout on streams is needed or they can take a long time if not forever to fail with my Internet
2. `play_game` can get stuck in backoff with `board` is uninitialized while leaving the function
3. When control stream fails but not game stream or in a different order, it can cause races to resume active games

noobpwnftw:
Additional exception catching for `check_online_status` and `get_ongoing_games`. Also stop engine pondering in case a book move is found, it makes no sense as next time it'll be another position to begin search with.

MarkZH:
@noobpwnftw Do you have an idea of how much more work remains for this PR will be done? The latest change seems to be expanding the scope of this PR beyond ensuring that game processes are properly terminated in the event of errors. The handling of the matchmaking challenge ID looks like a good change, but I think it belongs in a new PR.

noobpwnftw:
I hope the end is near. Hangs are not easily reproducible, so usually I wake up and fix the last thing shown in the logs.

MarkZH:
OK. If the changes are relevant to the hangs, then that's fine. It's not fun debugging when the bugs can take days to show up.

AttackingOrDefending:
Does this fix #607?

noobpwnftw:
It should.

AttackingOrDefending (approved):

AttackingOrDefending:
Is there any reason why `multiprocessing.set_start_method('spawn')` isn't run on tests?

noobpwnftw:
Tests don't call `__main__`, which is required https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_start_method

Platform defaults here:
https://docs.python.org/3/library/multiprocessing.html#multiprocessing.get_start_method

MarkZH (approved):

PR #620, opened 2022-12-25
Add type hints

AttackingOrDefending:
Adds type hints to all functions.

### DIFF
--- a/config.py
+++ b/config.py
@@ -3,6 +3,7 @@ import os
 import os.path
 import logging
 import math
+from abc import ABCMeta
 from enum import Enum
 from typing import Dict, Any
 CONFIG_DICT_TYPE = Dict[str, Any]
@@ -27,7 +28,7 @@ class Configuration:
         data = self.config.get(name)
         return Configuration(data) if isinstance(data, dict) else data
 
-    def items(self):
+    def items(self) -> Any:
         return self.config.items()
 
     def __bool__(self) -> bool:
@@ -45,7 +46,7 @@ def config_assert(assertion: bool, error_message: str) -> None:
         raise Exception(error_message)
 
 
-def check_config_section(config: CONFIG_DICT_TYPE, data_name: str, data_type: type, subsection: str = "") -> None:
+def check_config_section(config: CONFIG_DICT_TYPE, data_name: str, data_type: ABCMeta, subsection: str = "") -> None:
     config_part = config[subsection] if subsection else config
     sub = f"`{subsection}` sub" if subsection else ""
     data_location = f"`{data_name}` subsection in `{subsection}`" if subsection else f"Section `{data_name}`"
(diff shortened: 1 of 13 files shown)


AttackingOrDefending:
This should be ready for review.
I only added type hints to all lines with `def`, not to every variable. This is intended for another PR.
I didn't add mypy to github actions because to get the number of errors down to 0, will require a lot of changes to the existing code.
There are 191 mypy errors remaining (using `mypy . --strict --exclude=.venv`).
In `lichess-bot.py`, all the different `Dict[str, Any]`s were added, so that the user will be able to differentiate between the different dicts. For mypy they are the exact same.
I used `queue.Queue` and `list` event though we use a proxy of them, because I couldn't use the proxies, but they work the exact same way.
To use `dict`, `list`, `x | None`, `x | y` instead of `Dict`, `List`, `Optional[x]`, `Union[x, y]`, would require dropping support for python 3.7 and 3.8.

config.py
```diff
@@ -3,41 +3,50 @@
...
-from matchmaking import DelayType
+from abc import ABCMeta
+from enum import Enum
+from typing import Dict, Any
+CONFIG_DICT_TYPE = Dict[str, Any]
 
 logger = logging.getLogger(__name__)
 
 
+class DelayType(str, Enum):
```
MarkZH, line 14: Why was `DelayType` copied here?

AttackingOrDefending: Because there was a circular import. This is why `MAX_CHAT_MESSAGE_LEN` was also copied. Because I added `from config import Configuration` in matchmaking.py I couldn't import anything from matchmaking.

MarkZH: For `MAX_CHAT_MESSAGE_LEN`, you can delete the definition in engine_wrapper.py and use `lichess.MAX_CHAT_MESSAGE_LEN` inside the `get_stats()` method.

Similarly for `DelayType`, delete the definition in matchmaking.py and use `config.DelayType` instead.

This removes duplicate definitions.

AttackingOrDefending:
The new tests for python 3.8 found some errors here. They were easy to fix, but unfortunately the quotes will have to be kept in the types initialized in `lichess-bot.py` (e.g. `CONTROL_QUEUE_TYPE = "queue.Queue[EVENT_TYPE]"`), because python 3.8 doesn't support using `Queue[...]`.

MarkZH (approved):

MarkZH:
Is the idea here like the flake8: put the structure in place and then slowly fix up the errors revealed?

AttackingOrDefending:
I added mypy on github actions, but it will always be failing until we fix all the erros. It is somewhat like `flake8` (as you can also see [here](https://mypy.readthedocs.io/en/stable/existing_code.html)), but it is hard to fix all the errors. Also, `lichess-bot.py` will always raise an error, because I added quotes around some types as mentioned in my previous comment, until we drop support for `python 3.8`.

PR #622, opened 2022-12-28
Fine grained rate limiting

### LINKED ISSUE #621: Rate limit hangs the event loop
For some reason since today, matchmaking challenges get rate limited even if they were made less than once per minute(daily quota somewhere?), which in turn blocks the event thread for 60s with every request.

MarkZH:
When a bot hits rate limits and lichess responds with a 429, use a 60-second timer on that endpoint instead of pausing the main thread of the program. This should allow game/event streams to continue mostly normally.

Will close #621.

### DIFF
--- a/lichess.py
+++ b/lichess.py
@@ -129,7 +129,7 @@ class Lichess:
         return self.api_get(ENDPOINTS["game"], game_id)
 
     def upgrade_to_bot_account(self):
-        return self.api_post(ENDPOINTS, "upgrade")
+        return self.api_post(ENDPOINTS["upgrade"])
 
     def make_move(self, game_id, move):
         return self.api_post(ENDPOINTS["move"], game_id, move.move,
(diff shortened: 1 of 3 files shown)


MarkZH:
Testing will be needed since it can take days for a rate limit to show up.

noobpwnftw:
Maybe it should throw exceptions instead of silently returning empty.

AttackingOrDefending:
In line 78 `self.set_rate_limit_timer(path_template, delay)` should be changed to `self.set_rate_limit_delay(path_template, delay)`.
In line 105 `self.set_rate_limit_delay(60)` should be changed to `self.set_rate_limit_delay(endpoint_name, 60)`.
In line 127 `return self.rate_limit_time_left[path_template].time_until_expiration()` should be changed to `return self.rate_limit_timers[path_template].time_until_expiration()`.

MarkZH:
Is the behavior correct? Is there anywhere else that needs to handle the `RateLimited` exception?

AttackingOrDefending (approved):

Issue #623, opened 2022-12-29
Process leaks eventually resulting in OOM


tm512:
Since setting up lichess-bot, I've noticed that my server eventually becomes unresponsive, forcing me to hard reset, with OOM errors spammed into my system logs. Monitoring the system using top, I notice that lichess-bot will occasionally spawn but fail to kill instances of my chess engine, often in pairs with sequential PIDs which makes me think it's spawning multiple engines at once.

I'm not entirely sure what kind of situation causes this, if anything in particular, since I've failed to catch it happening in real-time to see what lichess-bot prints out to the console, but I can update this with additional info if I finally manage to catch it. I'll probably start piping the output through tee to log everything.

If it's relevant, I'm running lichess-bot on Python 3.9.14 on DragonflyBSD.

MarkZH:
You can write lichess-bot log messages to a file with `-l filename.txt` and add verbosity with `-v`. I'm interested in seeing these logs since this is the first I've heard of this problem. Please upload the entire log from start to OOM error to https://gist.github.com/.

tm512:
Looking at the log it appears that lichess-bot is sending the `stop` UCI command but not following it with `quit` like it normally does. As soon as the engine gets `stop` it responds with `bestmove`, so it seems like it might have something to do with the game being ended mid-search? The log is huge and cluttered by output from a concurrent game so I'm restarting it with concurrency set to 1 so I can get a clean log snippet to paste if the additional log information is necessary.

MarkZH:
As a first try at fixing this issue, we've added the `isready/readyok` query between the final `stop` and `quit` commands. If the problem is caused by the communication channel being shutdown before the engine can finish responding to the `stop` command, this may help.

tm512:
Still happens even after [73577de](https://github.com/ShailChoksi/lichess-bot/commit/73577de9bfe8501e23452570a2445f01721e939a). Here's a gist of the problem happening where it leaks two instances in succession: https://gist.github.com/tm512/7b9dc09872c12af9358afc005117afa8

It kinda seems like the connection to lichess gets interrupted, causing lichess-bot to spawn a new engine instance without sending `quit` to the previously spawned one, considering all of these are spawned for the exact same game ID.

AttackingOrDefending:
[...]

PR #625, opened 2022-12-30
Improve logging

AttackingOrDefending:
- Logs the game_id of engine searches
- Logs the pid of every game_id
- Adds the id to `model.Game.__str__` and `model.Challenge.__str__`
- Added ping. We ping the engine before we close it. Not sure if this helps. The idea came from #623, so we make sure that the engine is alive when we quit, but I don't think this solves the problem.

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -37,7 +37,7 @@ def create_engine(config):
         raise ValueError(
             f"    Invalid engine type: {engine_type}. Expected xboard, uci, or homemade.")
     options = remove_managed_options(cfg.lookup(f"{engine_type}_options") or {})
-    logger.debug(f"Starting engine: '{commands}'")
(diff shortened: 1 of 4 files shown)


engine_wrapper.py
```diff
@@ -37,7 +37,7 @@ def create_engine(config):
         raise ValueError(
             f"    Invalid engine type: {engine_type}. Expected xboard, uci, or homemade.")
     options = remove_managed_options(cfg.lookup(f"{engine_type}_options") or {})
-    logger.debug(f"Starting engine: {' '.join(commands)}")
+    logger.debug(f"Starting engine: '{commands}'")
```
MarkZH, line 40: Delete the single quotes in this string to get correct coloring in the console log.

MarkZH (approved):

PR #627, opened 2023-01-01
Always close the engine

### LINKED ISSUE #623: Process leaks eventually resulting in OOM

AttackingOrDefending:
Makes sure that the engine is closed. The only error this won't catch is if it happens inside `__init__` in `UCIEngine`, `XBoardEngine` or `MinimalEngine`.

fixes #623

### DIFF
--- a/model.py
+++ b/model.py
@@ -98,7 +98,8 @@ class Challenge:
         return f'{self.challenger_title or ""} {self.challenger_name}'.strip()
 
     def __str__(self):
-        return f"{self.perf_name} {self.mode()} challenge from {self.challenger_full_name()}({self.challenger_rating}) ({self.id})"
+        return f"{self.perf_name} {self.mode()} challenge from {self.challenger_full_name()}({self.challenger_rating})" \
+               f" ({self.id})"
 
     def __repr__(self):
         return self.__str__()
--- a/model.py
+++ b/model.py
@@ -98,8 +98,8 @@ class Challenge:
         return f'{self.challenger_title or ""} {self.challenger_name}'.strip()
 
     def __str__(self):
-        return f"{self.perf_name} {self.mode()} challenge from {self.challenger_full_name()}({self.challenger_rating})" \
-               f" ({self.id})"
+        return (f"{self.perf_name} {self.mode()} challenge from {self.challenger_full_name()}({self.challenger_rating})"
+                f" ({self.id})")
 
     def __repr__(self):
         return self.__str__()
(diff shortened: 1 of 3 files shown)


MarkZH:
There's a way to do this without increasing the complexity of `play_game()` using a [context manager decorator](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) on `create_engine()`. The the `create_engine()` function would look like this:
```python
# engine_wrapper.py

from contextlib import contextmanager

@contextmanager
def create_engine(config):

    # ... parse options

    engine = Engine(commands, options, stderr, cfg.draw_or_resign, cwd=cfg.working_dir)
    try:
        yield engine
    finally:
        engine.stop()
        engine.ping()
        engine.quit()
```
Then, `create_engine()` and succeeding `try:` in `play_game()` is replaced with:
```python
# lichess-bot.py

    with engine_wrapper.create_engine(config) as engine:
        # ... the rest
```
The `finally:` block associated with that `try:` can be deleted since it will be handled by the `finally:` in `create_engine()`.

model.py
```diff
@@ -98,7 +98,8 @@ def challenger_full_name(self):
         return f'{self.challenger_title or ""} {self.challenger_name}'.strip()
 
     def __str__(self):
-        return f"{self.perf_name} {self.mode()} challenge from {self.challenger_full_name()}({self.challenger_rating}) ({self.id})"
+        return f"{self.perf_name} {self.mode()} challenge from {self.challenger_full_name()}({self.challenger_rating})" \
+               f" ({self.id})"
```
MarkZH, line 102: Putting parentheses around the multi-line string is less fragile than escaping the newline. Accidentally putting space characters after the backslash will cause errors well after the program starts.

MarkZH:
Let me know once this is ready to merge.

AttackingOrDefending:
Ready.

MarkZH (approved):

PR #638, opened 2023-01-21
Python versioning scheme

MarkZH:
This commit provides a way to check if a recent enough version of Python is being used and a way to alert users to the need to upgrade Python so that new features can be used.

All versioning information is now centralized in a separate file. Now the commit history of lichess-bot.py won't be cluttered with version number updates.

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -17,6 +17,9 @@ import os
 import io
 import copy
 import math
+import sys
+import datetime
+import yaml
 from config import load_config, Configuration
 from conversation import Conversation, ChatLine
 from timer import Timer
@@ -40,7 +43,10 @@ POOL_TYPE = multiprocessing.Pool
 
 logger = logging.getLogger(__name__)
 
-__version__ = "2023.1.21.1"
+with open("versioning.yml") as version_file:
+    versioning_info = yaml.safe_load(version_file)
+
+__version__ = versioning_info["lichess_bot_version"]
 
 terminated = False
 restart = True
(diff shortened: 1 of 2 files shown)


MarkZH:
The deprecation date in the file is something we should discuss, I just picked it as an example of how this function works. It should probably be later, though no later than October 2024, since that's when 3.8 is EOL.

AttackingOrDefending (approved):

AttackingOrDefending:
I think the deprecation date should be before September 2023. Not that many people seem to be using old python versions. We can leave the date as is, and if a lot of people use 3.8 we can postpone it.

MarkZH:
I have some more changes to make. Namely, the exception should only be thrown if the user tries to use a newer version of lichess-bot that requires a newer Python, not if the deprecation date has passed. If the user wants to use an older version of lichess-bot, they should be able to.

MarkZH:
Now, a user can keep using an older version of lichess-bot if they want to use a deprecated version of Python. Plus, the code was cleaned up.

As for the current deprecation date, I left it as is. I think three month's warning should be enough to see if anyone wants to keep using Python 3.8. 

One thing I've noticed: the reason that Python 3.7 stopped working was due to an import that was not available in 3.7. This new function would not catch that since the bad import would throw a program-terminating exception beforehand. Perhaps the function could be moved above all of the imports that are not needed for it to run so that the user would get a better error message.

Finally, to prepare for upgrading python versions, we can label PRs with `requires-3.8`, `requires-3.9`, and so on so that the next deprecation date can be planned.

AttackingOrDefending:
Can you change lines 779-782 and `version_str` (line 775) to an f-string?

PR #639, opened 2023-01-22
Run fewer GitHub actions

AttackingOrDefending:
GitHub actions will now only be run when there is a PR, or the push commit message contains the text `Update version` (which it already does because we use `Update version to x.y.z.n`). This is done to avoid having to rerun as many failures. Also, update SF and LC0 versions, and use `DEBUG` instead of `INFO`.

### DIFF
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -9,7 +9,8 @@ on:
     branches: [ master ]
 
 jobs:
-  build:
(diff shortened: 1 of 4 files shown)


MarkZH:
Ready to merge?

AttackingOrDefending:
Yes.

MarkZH (approved):

PR #640, opened 2023-01-22
Add type hints to variables

AttackingOrDefending:
From the 96 mypy errors that are left (run with `mypy --strict . --exclude=.venv`), 90 are in `lichess-bot.py` and `engine_wrapper.py`.

### DIFF
--- a/conversation.py
+++ b/conversation.py
@@ -55,11 +55,11 @@ class Conversation:
 
     def send_message(self, room: str, message: str) -> None:
         if message:
-            self.send_reply(ChatLine({"room": room}), message)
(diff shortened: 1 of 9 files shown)


MarkZH:
Do you want me to start a PR fixing up lichess-bot.py or engine-wrapper.py? Or do you want to handle that?

AttackingOrDefending:
I guess we can merge this and create new PR for the remaining errors. A big problem is that in `engine_wrapper.py` I use `EngineWrapper` instead of `Union[UCIEngine, XBoardEngine, ExampleEngine]` because there would be a circular import. I will start working on adding type hints to lichess-bot. We can create a branch that we both add type hints to.

MarkZH (approved):

PR #644, opened 2023-01-25
Update config.yml.default

Selamiyet:


MarkZH:
Each user needs to have their own unique token. There's no point in putting a token into the default config file. If this is your bot's token, you should delete it and create a new one so other people seeing this PR won't be able to impersonate your account.

PR #647, opened 2023-01-29
Add more type hints

AttackingOrDefending:
Down to 3 significant errors. There are 34 errors caused by the quotes in the start of `lichess-bot.py`. We will be able to remove the quotes when we drop support for `3.8`. The errors are:

- 2 in `lichess-bot.py`
- 1 in `test_bot.py`

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -82,19 +82,19 @@ def watch_control_stream(control_queue: CONTROL_QUEUE_TYPE, li: lichess.Lichess)
             for line in lines:
                 if line:
                     event = json.loads(line.decode("utf-8"))
-                    control_queue.put_nowait(event)
+                    control_queue.put_nowait(event)  # type: ignore[attr-defined]
                 else:
-                    control_queue.put_nowait({"type": "ping"})
+                    control_queue.put_nowait({"type": "ping"})  # type: ignore[attr-defined]
         except Exception:
             break
 
-    control_queue.put_nowait({"type": "terminated"})
+    control_queue.put_nowait({"type": "terminated"})  # type: ignore[attr-defined]
 
 
 def do_correspondence_ping(control_queue: CONTROL_QUEUE_TYPE, period: int) -> None:
     while not terminated:
         time.sleep(period)
-        control_queue.put_nowait({"type": "correspondence_ping"})
+        control_queue.put_nowait({"type": "correspondence_ping"})  # type: ignore[attr-defined]
 
 
 def logging_configurer(level: int, filename: Optional[str]) -> None:
@@ -119,12 +119,12 @@ def logging_listener_proc(queue: LOGGING_QUEUE_TYPE, level: int, log_filename: O
     logging_configurer(level, log_filename)
     logger = logging.getLogger()
     while not terminated:
-        task = queue.get()
+        task = queue.get()  # type: ignore[attr-defined]
         try:
             logger.handle(task)
         except Exception:
             pass
-        queue.task_done()
+        queue.task_done()  # type: ignore[attr-defined]
 
 
 def game_logging_configurer(queue: Union[CONTROL_QUEUE_TYPE, LOGGING_QUEUE_TYPE], level: int) -> None:
@@ -143,7 +143,7 @@ def start(li: lichess.Lichess, user_profile: USER_PROFILE_TYPE, config: Configur
           log_filename: Optional[str], one_game: bool = False) -> None:
     logger.info(f"You're now connected to {config.url} and awaiting challenges.")
     manager = multiprocessing.Manager()
-    challenge_queue: MULTIPROCESSING_LIST_TYPE = manager.list()
+    challenge_queue: MULTIPROCESSING_LIST_TYPE = manager.list()  # type: ignore[assignment]
     control_queue: CONTROL_QUEUE_TYPE = manager.Queue()
     control_stream = multiprocessing.Process(target=watch_control_stream, args=(control_queue, li))
     control_stream.start()
@@ -230,7 +230,7 @@ def lichess_bot_main(li: lichess.Lichess,
 
             if event["type"] == "terminated":
                 restart = True
-                control_queue.task_done()
+                control_queue.task_done()  # type: ignore[attr-defined]
                 break
             elif event["type"] in ["local_game_done", "gameFinish"]:
                 active_games.discard(event["game"]["id"])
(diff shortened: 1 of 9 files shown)


MarkZH:
For lichess-bot.py, you can put `# type: ignore[valid-type]` at the top of the file to ignore all the quoted type annotations. It can be removed once we require python 3.8. This way, we'll can get a clean mypy run until then.

MarkZH:
For the `EngineWrapper` types, I think I have a different solution. I'll make a PR later.

MarkZH:
Correction, the correct command for disabling the `[valid-type]` check is
```python
# mypy: disable-error-code=valid-type
```

AttackingOrDefending:
Done, but this doesn't fix the `attr-defined` errors (e.g. `lichess-bot.py:303: error: CORRESPONDENCE_QUEUE_TYPE? has no attribute "qsize"  [attr-defined]`).

AttackingOrDefending:
For some errors, like in `test_bot.py` line 154, we can fix the errors by adding some unnecessary code. Changing
```python
win = board.outcome().winner == chess.BLACK
```
to
```python
outcome = board.outcome()
win = False
if isinstance(outcome, chess.Outcome):
    win = outcome.winner == chess.BLACK
```
Do you think we should add a little unecessary code or use `# type: ignore`?

MarkZH:
Another way to write the test_bot.py line is
```python
    outcome = board.outcome()
    win = outcome.winner == chess.BLACK if outcome else False
```
Likewise, in test_bot/lichess.py
```python
        if move.move:
            uci_move: str = move.move.uci()
        else:
            raise RuntimeError("Engine attempted to make null move.")
```
The `else:` branch is unlikely, but it is a better error message than `AttributeError: 'NoneType' object has no attribute 'uci'.`

As for the `[attr-defined]` errors, I think putting an `ignore` statement on the lines with errors due to stringy types is good for now (and on `restart` on line 170 of test_bot/test_bot.py). Mypy has an option [`--warn-unused-ignores`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-warn-unused-ignores) that will alert us when the ignore statements are no longer necessary (like once we require Python 3.8). We may also consider adding the option [`--warn-redundant-casts`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-warn-redundant-casts) to make sure we're not overdoing it on the explicit type annotations.

AttackingOrDefending:
`--warn-unused-ignores` and `--warn-redundant-casts` are already enabled by `--strict`.
I fixed most of the errors but there are 3 remaining that I don't know how to fix. The PR is ready, but it should be tested for bugs.

lichess-bot.py
```diff
@@ -685,24 +687,35 @@ def print_pgn_game_record(li: lichess.Lichess, config: Configuration, game: mode
     game_path = os.path.join(config.pgn_directory, game_file_name)
 
     lichess_game_record = chess.pgn.read_game(io.StringIO(li.get_game_pgn(game.id)))
+    if not isinstance(lichess_game_record, chess.pgn.Game):
+        return
     try:
         # Recall previously written PGN file to retain engine evaluations.
         with open(game_path) as game_data:
             game_record = chess.pgn.read_game(game_data)
```
MarkZH, line 695: It would be better to set `game_record` to `lichess_game_record` rather than returning early:

`game_record = chess.pgn.read_game(game_data) or lichess_game_record`

lichess-bot.py
```diff
@@ -685,24 +687,35 @@ def print_pgn_game_record(li: lichess.Lichess, config: Configuration, game: mode
     game_path = os.path.join(config.pgn_directory, game_file_name)
 
     lichess_game_record = chess.pgn.read_game(io.StringIO(li.get_game_pgn(game.id)))
+    if not isinstance(lichess_game_record, chess.pgn.Game):
```
MarkZH, line 690: `lichess_game_record = chess.pgn.read_game(io.StringIO(li.get_game_pgn(game.id))) or chess.pgn.Game()`

This will negate the next `if not isistance`. The only way this line could fail is if there is a communication error with lichess, in which case the raised exception would end the function anyway.

MarkZH:
Here's a diff that fixes `print_pgn_game_record()`. I don't think returning early is correct, as that doesn't allow the PGN to be saved nor the engine to write its evaluations. This also takes care of the complexity. I use the fact that `node.is_end()` is the same as `node.next() == None`.

```diff
diff --git a/lichess-bot.py b/lichess-bot.py
index 0b9455d..aaba849 100644
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -686,15 +686,11 @@ def print_pgn_game_record(li: lichess.Lichess, config: Configuration, game: mode
     game_file_name = "".join(c for c in game_file_name if c not in '<>:"/\\|?*')
     game_path = os.path.join(config.pgn_directory, game_file_name)
 
-    lichess_game_record = chess.pgn.read_game(io.StringIO(li.get_game_pgn(game.id)))
-    if not isinstance(lichess_game_record, chess.pgn.Game):
-        return
+    lichess_game_record = chess.pgn.read_game(io.StringIO(li.get_game_pgn(game.id))) or chess.pgn.Game()
     try:
         # Recall previously written PGN file to retain engine evaluations.
         with open(game_path) as game_data:
-            game_record = chess.pgn.read_game(game_data)
-        if not isinstance(game_record, chess.pgn.Game):
-            return
+            game_record = chess.pgn.read_game(game_data) or lichess_game_record
         game_record.headers.update(lichess_game_record.headers)
     except FileNotFoundError:
         game_record = lichess_game_record
@@ -706,16 +702,11 @@ def print_pgn_game_record(li: lichess.Lichess, config: Configuration, game: mode
         if next_node is None or next_node.move != move:
             current_node = current_node.add_main_variation(move)
         else:
-            optional_current_node = current_node.next()
-            if not isinstance(optional_current_node, chess.pgn.ChildNode):
-                break
-            current_node = optional_current_node
+            current_node = next_node
 
-        if not lichess_node.is_end():
-            optional_lichess_node = lichess_node.next()
-            if not isinstance(optional_lichess_node, chess.pgn.ChildNode):
-                break
-            lichess_node = optional_lichess_node
+        next_lichess_node = lichess_node.next()
+        if next_lichess_node:
+            lichess_node = next_lichess_node
             current_node.set_clock(lichess_node.clock())
             if current_node.comment != lichess_node.comment:
                 current_node.comment = f"{current_node.comment} {lichess_node.comment}".strip()
```

MarkZH:
As for the rest of the errors:

1. I would put ignore statements on all the remaining `[attr-defined]` errors. The queue stringy types we'll fix with Python 3.9. The one on test_bot/test_bot.py line 173 seems to be a bug in mypy.
2. Line 426 in lichess-bot.py: The argument should be `Exception` instead of `BaseException`.
3. I don't know what to do about the `List[model.Challenge]` vs `ListProxy[Any]` on line 265 of lichess-bot.py. I would have thought that using `MULTIPROCESSING_LIST_TYPE = multiprocessing.managers.ListProxy[model.Challenge]` would work, but apparently that's not a valid type definition. To get a clean mypy run, we can put an ignore statement here as well.

To keep track of the ignore statements, we can open an issue to list them to make sure we revisit them occasionally.

MarkZH:
Yep. I was editing my post when you posted. So, one more ignore statement.

MarkZH (approved):

PR #658, opened 2023-02-04
Handle errors from Lichess class outside of games

MarkZH:
These changes ignore errors from contacting lichess.org outside of games. In each case, ignoring the error and continuing to run is either acceptable (when declining a challenge) or much less inconvenient than stopping the whole program (updating bot ratings or failing to download the completed game PGN).

Other methods that are called outside of games are already handled.

### DIFF
--- a/lichess.py
+++ b/lichess.py
@@ -223,7 +223,7 @@ class Lichess:
                                           "application/x-www-form-urlencoded"},
                                  raise_for_status=False)
         except Exception:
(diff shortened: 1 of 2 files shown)


MarkZH:
The missing PGN information can be filled in manually from the `model.Game` instance, but that is beyond the scope of this PR.

AttackingOrDefending (approved):

PR #665, opened 2023-02-07
Fix `IndexError` in matchmaking

AttackingOrDefending:
In a config like:
```yml
  challenge_days:
#    - 1
#    - 2
```
There would be an `IndexError` raised in line 93 of `matchmaking.py`.


MarkZH:
I think this PR requires one more change. Zero counts as a falsey value, so a user that wanted their bot to only play 5+0 games could not write their config file like this:
```yaml
matchmaking:
    challenge_initial_time: 5
    challenge_increment: 0
```
After processing, this would result in
```yaml
matchmaking:
    challenge_initial_time:
        - 5
    challenge_increment:
        - 2
```
The function `change_value_to_list()` was written to accommodate users that wrote single-value parameters like this instead of as a list.

Lines 66-69 of config.py should change from
```python
    if force_falsey_values:
        subconfig[key] = subconfig.get(key) or default
    else:
        subconfig.setdefault(key, default)
```
to 
```python
    if force_falsey_values:
        if subconfig.get(key) is None:
            subconfig[key] = default
    else:
        subconfig.setdefault(key, default)
```
This way, only `None` counts as a falsey value, rather than `0`, `""`, `{}`, etc. Only empty entries should defaulted by `force_falsey_values`.

MarkZH:
The error in testing happens because the `engine: working_dir` value in config.yml.default is the empty string. My suggestion causes empty strings to no longer be replaced when `force_falsey_values=True`. My fault for not testing it on my machine. I think the inner check should be
```python
if subconfig.get(key) in [None, ""]:
    subconfig[key] = default
```
In this context, I don't think there's a semantic difference between an empty value `param: ` and an empty string value `param: ""`, whereas other falsey values like `0` have definite meaning.

PR #684, opened 2023-04-06
Delay rechallenging based on specific challenge decline reason

MarkZH:
Lichess now includes a machine-readable key for the reason a challenge was declined. This key can be used to not rechallenge a potential opponent with game configurations that would lead to rejection. For example, if the user's bot challenges another bot with a rated blitz standard game and the challenged bot declines with a `declineReasonKey` of `timecontrol`, then lichess-bot will delay creating another blitz challenge for that opponent.


MarkZH:
I'm thinking of replacing the timers with boolean values. This would simplify things a lot. I've noticed that the connection to lichess gets reset at least once a day, so that would allow a long-running session of lichess-bot to check if any opponents have changed their settings.

MarkZH:
Switched from Timers to bool for avoiding declination of challenges, which means these challenge filters never expire. Let me know if this is a good change.

AttackingOrDefending (approved):

PR #685, opened 2023-04-10
Drop support for Python 3.8

MarkZH:
On May 1, 2023, lichess-bot officially drops support for Python 3.8 and will require at least Python 3.9 going forward. The main benefit will be to finish type hinting with the `Queue` type. Further changes:
- Use the `MutableSequence` type for `MULTIPROCESSING_LIST_TYPE` that describes the `challenge_queue`. This is slightly more generic than the specific type returned by `multiprocessing.Manager().list()`, but `MutableSequence` is still an accurate description of what it is. Functions that only read the `challenge_queue` can use a `Sequence` type to flag modifications as an error.
- Put all packages for running tests in requirements.txt and delete the separate `pip` lines from the github actions.
- Create a function that sets `restart` to `False` in lichess-bot.py when testing. Functions seem to be more visible to mypy than variables.

### DIFF
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest]
-        python: [3.8, "3.11"]
+        python: [3.9, "3.11"]
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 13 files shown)


MarkZH:
This is ready to merge.

AttackingOrDefending (approved):

PR #657, opened 2023-02-03
Use new python-chess engine method to send opponent information

MarkZH:
The python-chess library has been updated so that opponent information can be sent to the engine after the "new" command but before the first move.

Closes #654.

PR #690, opened 2023-04-16
Remove all calls to SimpleEngine.protocol.send_line()

MarkZH:
It is incorrect to call the method `SimpleEngine.protocol.send_line()` outside the context of the event loop used within the other `SimpleEngine` methods. Within lichess-bot, these calls are found within the methods `stop()` and recursively in `report_game_result()` for both `UCIEngine`. In about 10% of `play_game()` threads when lichess-bot is run on Windows, something doesn't get shut down correctly after the game finishes and the debug logged gets spammed by these messages once per second forever:
```
<IocpProactor overlapped#=1 result#=0> is running after closing for 1.0 seconds
<IocpProactor overlapped#=1 result#=0> is running after closing for 2.0 seconds
<IocpProactor overlapped#=1 result#=0> is running after closing for 3.0 seconds
...
```
After deleting the methods, these messages no longer appear in the logs.

`XBoardEngine.get_opponent_info()` also has this problem, but it will be fixed with PR #657.

The main consequence of this change is that the final moves in a game will not be sent to the engines and `result` (e.g., `result 1-0 {White mates}`) commands will not be sent to XBoard engines. I don't think this is a big deal since PGN files are written out after every game if the user chooses. I'll probably create a PR in python-chess to restore this behavior at a later date.

Another consequence--specifically from deleting the `stop()` command on line 168 of engine_wrapper.py in the `play_move()` method--is that a pondering engine will not be stopped if a book/online move is found. I don't think this will happen with any frequency since it should be rare that a position with no book/db entry (resulting in an engine move with pondering) is followed by a position that has an entry.

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -349,9 +349,6 @@ class EngineWrapper:
         name: str = engine_info["name"]
         return name
 
-    def report_game_result(self, game: model.Game, board: chess.Board) -> None:
-        pass
-
     def get_pid(self) -> str:
         pid = "?"
         if self.engine.transport is not None:
@@ -383,10 +380,6 @@ class UCIEngine(EngineWrapper):
             player_type = "computer" if title == "BOT" else "human"
             self.engine.configure({"UCI_Opponent": f"{title} {rating} {player_type} {name}"})
 
-    def report_game_result(self, game: model.Game, board: chess.Board) -> None:
-        if isinstance(self.engine.protocol, chess.engine.UciProtocol):
-            self.engine.protocol._position(board)
-
 
 class XBoardEngine(EngineWrapper):
     def __init__(self, commands: COMMANDS_TYPE, options: OPTIONS_TYPE, stderr: Optional[int],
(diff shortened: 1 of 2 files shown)


MarkZH:
I would like to wait on merging this until after PR #657 is merged to keep `XBoardEngine` and `UCIEngine` equal feature-wise (namely, the opponent info methods).

To find the problem, I used the following python script that is a barebones version of lichess-bot.py. The `IocpProactor overlapped` messages should start appearing within 10-20 minutes when run on Windows. Comment out the `engine.protocol.send_line("stop")` line in the `finally`-block to stop the messages from appearing. I suggest directing output to a file since there will be a lot of output.
```
import chess
import chess.engine
import logging
from multiprocessing.pool import Pool
import multiprocessing
import time

logging.basicConfig(level=logging.DEBUG, force=True)

def play_game():
    try:
        engine = chess.engine.SimpleEngine.popen_uci([r"C:\path\to\engine.exe"])
        board = chess.Board()
        while not board.is_game_over():
            result = engine.play(board, chess.engine.Limit(time=0.1))
            board.push(result.move)
    finally:
        engine.protocol.send_line("stop")
        engine.ping()
        engine.quit()
        engine.close()

if __name__ == "__main__":
    multiprocessing.set_start_method("spawn")
    with Pool(4) as pool:
        while True:
            pool.apply_async(play_game)
            time.sleep(1)
```

MarkZH:
This PR may be redundant after merging #699.

MarkZH:
This PR is entirely redundant.

PR #699, opened 2023-04-26
Send game result to engine after game

### LINKED ISSUE #696: Position command while pondering
It seems that lichess-bot issues a fixed series os commands whenever a match finishes for any reason (checkmate, checkmate against, draw, resignation based on resign_score config, etc).

The series of commands seems to be:
position startpos moves [the complete move list]
stop
isready
quit

Except for the obviously expected "quit" command, I don't undestand why these commands are issued. There might be a reason I don't fully undestand. Well, that's not a big issue. They don't hurt.

Except that the first command (position) seems to be illegal if just previously lichess-bot issued a "go ponder" command. While pondering, I expected either a "stop" or a "ponderhit" commands. Nothing else. Am I wrong?

MarkZH:
This is the proper way to send endgame messages to engines within the event loop used by python-chess.

This should be merged after #690 and after https://github.com/niklasf/python-chess/pull/998.

Closes #696


MarkZH:
The game in your logs continues because the `gameFinish` event comes from the event stream in `lichess_bot_main()`, not the game stream in `play_game()`. There's no threading in the `play_game()` function to watch for game stream events, so the next event--the draw offer, in this case--is not seen until the engine picks a move. The reason it took so long for the engine to pick a move (14 minutes and 9 minutes) is the long time control of 3 hours plus 3 minutes increment and 2 hours plus 2 minutes increment.

The game is removed from the `active_games` set [in this branch of the main event stream](https://github.com/lichess-bot-devs/lichess-bot/blob/666814894c8183da625bde3b05df8d08cfe0352b/lichess-bot.py#L327), specifically the `gameFinish` event at the time you accepted the draw for your bot through the lichess web site. One solution to this would be to only remove games after a `local_game_done` event. We would have to make sure that this event was put in the `control_queue` even in the event the `play_game()` function ended due to an exception. Otherwise, game IDs would pile up in the `active_games` set even after the game finished, which would prevent more games from starting. I think this fix should go in another PR.

AttackingOrDefending (approved):

PR #704, opened 2023-04-30
Use new clock_id to correct Xboard communication

### LINKED ISSUE #703: "level" command is improperly used
xboard engines expect to receive a "level" command to set the time control.

This command should be sent only once at the start of a new game (typically after "new") and subsequently if the time control changes (for example, at the expiration of an initial time control such as 40/2 and the beginning of a new one for the next n moves).

However, lichess-bot sends "level" after each move and apparently reduces the time control section of that command by the time used. For example I see:
level 0 4:57 3.0
at the start of a 5+3 time control game and then after each next move I see different quantities in the time control field, for example:
level 0 4:32 3.0
after a few moves. This is incorrect behavior. The level command is for setting the time control for a game or a portion of a game. The "time" and "otim" commands inform the engine after each move of its own and its opponent's time remaining.

MarkZH:
This will prevent sending redundant `level` and `st` commands to Xboard engines.

Closes #703


MarkZH:
Requires that [this python-chess PR](https://github.com/niklasf/python-chess/pull/1003) is merged and released.

AttackingOrDefending (approved):

PR #710, opened 2023-05-07
Fix difference in Linux/Windows handling of cwd

### LINKED ISSUE #707: Can't load engine binary on Linux (FileNotFoundError)
Hi, I wanted to move my working bot from Windows PC to Linux machine (Raspberry Pi 4), but the bot is failing to load the engine.
I'm sure that the file exists because I'm able to run the engine manually. I'm using the same config.yml file as on Windows (only had to remove .exe extension).

Here's full console output:
obraz

Lichess-bot version: **2023.5.2.1**

MarkZH:
When a user has the engine: working_dir: field set in their config file, the behavior of popen when opening the engine executable is different on Windows and Linux. In Linux, the current working directory (cwd) is changed before starting the executable. In Windows, it's the reverse. This means that Linux clients will not be able to find the executable if engine: working_dir is different than the cwd when lichess-bot starts.

These changes fix the problem by making the path to the engine absolute before starting the process. Also, the documentation is updated to notify users that the cwd change means relative paths to files are now relative to the working_dir.

Closes #707

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ Within the file `config.yml`:
 - Enter the executable name in the `engine: name` field (In Windows you may need to type a name with ".exe", like "lczero.exe")
 - If you want the engine to run in a different directory (e.g., if the engine needs to read or write files at a certain location), enter that directory in the `engine: working_dir` field.
   - If this field is blank or missing, the current directory will be used.
+  - IMPORTANT NOTE: If this field is used, the running engine will look for files and directories (Syzygy tablebases, for example) relative to this path, not the `engine: dir` field.
 - Leave the `weights` field empty or see [LeelaChessZero section](#leelachesszero) for Neural Nets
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -58,6 +58,7 @@ Within the file `config.yml`:
 - Enter the executable name in the `engine: name` field (In Windows you may need to type a name with ".exe", like "lczero.exe")
 - If you want the engine to run in a different directory (e.g., if the engine needs to read or write files at a certain location), enter that directory in the `engine: working_dir` field.
   - If this field is blank or missing, the current directory will be used.
+  - IMPORTANT NOTE: If this field is used, the running engine will look for files and directories (Syzygy tablebases, for example) relative to this path, not the `engine: dir` field.
```
AttackingOrDefending, line 61: I think that if `engine: working_dir` isn't set, the engine will look for files relative to `lichess-bot.py` and not `engine: dir`.

AttackingOrDefending: This should also be changed in `config.yml.default`.

PR #711, opened 2023-05-07
Update README.md due to PEP 668

MarkZH:
Due to conflicts with python modules that are installed from Linux package managers, some Python modules can't be installed through `pip` outside of a virtual environment. Due to [PEP 668](https://peps.python.org/pep-0668/), attempting to use `pip` to install `virtualenv` will fail with a warning about the module being "externally managed." This change updates the README to change the instructions for installing `virtualenv` from `pip` to the OS package manager.


MarkZH:
Looks like merging again worked. I don't know what happened the first time.

PR #718, opened 2023-05-14
Remove duplicated block_list section from `config.yml.default`

eduherminio:
Already present in line 149


MarkZH:
The two block lists serve different purposes. The block list you deleted prevents lichess-bot from challenging the bots on the list. The other block list prevents lichess-bot from accepting challenges from user names on the list. Both examples are needed.

It's worth considering whether or not the two lists should be combined into one.

PR #729, opened 2023-06-04
Add automatic logging

AttackingOrDefending:
There have been a lot of issues were the only info we get is a few lines copied from the console that were logged using `logging.INFO`. This PR attempts to fix this by automatically saving the logs to a file. These logs are kept for 7 days and then deleted, so if a user makes and issue we can tell them to upload the logs, so we have more info. Also, in another PR we should probably add an issue template where we tell users to upload the logs. It is enabled by default.

Pros:
- Automatically saves logs
- Saves all debug logs (the user doesn't have to specify the logging level)
- The logging level is `logging.NOTSET`, so we can add more log statements that don't show up with only `logging.DEBUG`

Cons:
- Can take quite a lot of space (which is why I added a filter for UCI info lines, and they are only saved for 7 days)
- Possible slowdown
- The reason we get only a few lines may be because the bot is hosted and the owner doesn't have access to the logs so this won't help

@MarkZH Do you think we need this? Also, what are some unimportant XBoard lines that can be added to the filter?


MarkZH:
I think this is a good idea. If you want to filter [XBoard thinking output](https://www.gnu.org/software/xboard/engine-intf.html#10), filter anything that starts with 4 integers separated by whitespace. In regular expression syntax, something like this: `(\d+\s+){4}`

The README should have a section near the top with instructions for filing bug reports including pasting log contents into the bug description (or linking with [Github gist](https://gist.github.com/)).

lichess-bot.py
```diff
@@ -132,19 +134,44 @@ def logging_configurer(level: int, filename: Optional[str]) -> None:
         file_handler.setFormatter(file_formatter)
         all_handlers.append(file_handler)
 
+    if auto_log_filename:
+        if not os.path.isdir("./log/"):
+            os.mkdir("./log/")
```
MarkZH, line 139: This `if` block can be replaced by a call to `os.makedirs("./log/", exist_ok=True)`

https://docs.python.org/3/library/os.html#os.makedirs

lichess-bot.py
```diff
@@ -132,19 +134,44 @@ def logging_configurer(level: int, filename: Optional[str]) -> None:
...
+
+        # Clear old logs.
+        seven_days = 7 * 24 * 60 * 60
```
MarkZH, line 142: `seven_days = datetime.timedelta(days=7).total_seconds()`

lichess-bot.py
```diff
@@ -132,19 +134,44 @@ def logging_configurer(level: int, filename: Optional[str]) -> None:
...
+        for file in os.listdir("./log/"):
+            path = os.path.join("./log/", file)
+            if os.path.getmtime(path) * seven_days < time.time():
```
MarkZH, line 145: The multiplication should be addition.

lichess-bot.py
MarkZH: Also, I think the name `log` is likely to conflict with existing directories that user's may use to store logs. A name like `lichess_bot_weekly_logs` is less likely to collide with user data.

MarkZH: Also, also, the log directory can be derived from `auto_log_file_name` with `os.path.dirname()`

https://docs.python.org/3/library/os.path.html#os.path.dirname

lichess-bot.py
```diff
@@ -926,10 +955,15 @@ def start_lichess_bot() -> None:
...
     parser.add_argument("-l", "--logfile", help="Record all console output to a log file.", default=None)
+    parser.add_argument("--disable_auto_logging", action="store_true", help="Disable automatic logging.")
     args = parser.parse_args()
 
     logging_level = logging.DEBUG if args.v else logging.INFO
-    logging_configurer(logging_level, args.logfile)
+    auto_log_filename = None
+    if not args.disable_auto_logging:
+        now = datetime.datetime.now()
+        auto_log_filename = f"./log/{now.year}-{now.month}-{now.day}_{now.hour}-{now.minute}-{now.second}.txt"
```
MarkZH, line 965: The file names will sort incorrectly without zero-padding the parts of the date.

`auto_log_filename = now.strftime("./log/%Y-%m-%d_%H-%M-%S.txt")`

This is probably the best place to change the log folder name (see above comments for line 139).

AttackingOrDefending:
I decided not to filter the logs, and instead use a RotatingFileHandler, with a max size of 100MB. This means that only the last 100-200MB of logs from every session run in the last 7 days will be stored at any given time. Every file will still be cleared 7 days after the last edit.

AttackingOrDefending:
It is also possible to compress old logs so that they take less space. For example:
```python
def handle_old_logs(directory: str) -> None:
    seven_days = datetime.timedelta(days=7).total_seconds()
    for file in os.listdir(directory):
        path = os.path.join(directory, file)
        if os.path.getmtime(path) + seven_days < time.time():
            os.remove(path)
        elif not path.endswith(".compressed_log"):
            with open(path) as file:
                contents = file.read()
            with lzma.open(f"{path}.compressed_log", "w") as file:
                file.write(contents.encode("utf-8"))
            os.remove(path)
```
Compressed log files can then be decompressed with:
```python
with lzma.open(f"{path}.compressed_log", "r") as file:
    contents = file.read().decode()
```
which can be provided in a file inside lichess-bot.
Compressing turned a 5678KB log to 376KB.
Should we add compression or not?

MarkZH:
I have some uncertainties about this.

I think seven days is too long. Most people will see a problem and only want the logs from the last run. I don't think anyone is going to go searching through multiple days of logs to find a problem. Anyone who would is probably already saving their own logs. Just saving a log file only for the last session should suffice--or maybe a single day. This would also negate the need for compression.

If the rolling file handler is kept, we need a way to save the configuration. The rolling file handler will lose the configuration if the log exceeds 200 MB due to only 1 rollover file begin saved.

If compression is kept, the compressed files should end with `.gz` so operating systems and users can identify what they are and use external tools on them.

AttackingOrDefending:
Now automatic logging is simplified. If lichess-bot isn't running there will be 2 logs saved: `recent.log` (the last session) and `old.log` (the 2nd to last session). When lichess-bot is run, `old.log` will be deleted, `recent.log` will be renamed to `old.log` and we will write into `recent.log`.

MarkZH:
The following diff fixes two problems:

1. `logging_configurer` is called multiple times due to threading, so the handling of old logs needs a parameter to make sure it only happens once per session.
2. Using `logging.NOTSET` didn't seem to work right. I used `logging.DEBUG` for the overall `logger` level and set the levels on each handler individually.

```diff
diff --git a/lichess-bot.py b/lichess-bot.py
index 2c4040e..a1ca0fe 100644
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -123,7 +123,7 @@ def handle_old_logs(auto_log_filename: str) -> None:
         os.rename(auto_log_filename, old_path)
 
 
-def logging_configurer(level: int, filename: Optional[str], auto_log_filename: Optional[str]) -> None:
+def logging_configurer(level: int, filename: Optional[str], auto_log_filename: Optional[str], delete_old_logs: bool) -> None:
     """
     Configure the logger.
 
@@ -134,6 +134,7 @@ def logging_configurer(level: int, filename: Optional[str], auto_log_filename: O
     console_handler = RichHandler()
     console_formatter = logging.Formatter("%(message)s")
     console_handler.setFormatter(console_formatter)
+    console_handler.setLevel(level)
     all_handlers: list[logging.Handler] = [console_handler]
 
     if filename:
@@ -141,24 +142,26 @@ def logging_configurer(level: int, filename: Optional[str], auto_log_filename: O
         FORMAT = "%(asctime)s %(name)s %(levelname)s %(message)s"
         file_formatter = logging.Formatter(FORMAT)
         file_handler.setFormatter(file_formatter)
+        file_handler.setLevel(level)
         all_handlers.append(file_handler)
 
     if auto_log_filename:
         os.makedirs(os.path.dirname(auto_log_filename), exist_ok=True)
 
         # Clear old logs.
-        handle_old_logs(auto_log_filename)
+        if delete_old_logs:
+            handle_old_logs(auto_log_filename)
 
         # Set up automatic logging.
         auto_file_handler = logging.FileHandler(auto_log_filename, delay=True)
-        auto_file_handler.setLevel(logging.NOTSET)
+        auto_file_handler.setLevel(logging.DEBUG)
 
         FORMAT = "%(asctime)s %(name)s %(levelname)s %(message)s"
         file_formatter = logging.Formatter(FORMAT)
         auto_file_handler.setFormatter(file_formatter)
         all_handlers.append(auto_file_handler)
 
-    logging.basicConfig(level=level,
+    logging.basicConfig(level=logging.DEBUG,
                         handlers=all_handlers,
                         force=True)
 
@@ -171,7 +174,7 @@ def logging_listener_proc(queue: LOGGING_QUEUE_TYPE, level: int, log_filename: O
     This allows the logs from inside a thread to be printed.
     They are added to the queue, so they are printed outside the thread.
     """
-    logging_configurer(level, log_filename, auto_log_filename)
+    logging_configurer(level, log_filename, auto_log_filename, False)
     logger = logging.getLogger()
     while not terminated:
         task = queue.get()
@@ -963,7 +966,7 @@ def start_lichess_bot() -> None:
     auto_log_filename = None
     if not args.disable_auto_logging:
         auto_log_filename = "./lichess_bot_auto_logs/recent.log"
-    logging_configurer(logging_level, args.logfile, auto_log_filename)
+    logging_configurer(logging_level, args.logfile, auto_log_filename, True)
     logger.info(intro(), extra={"highlighter": None})
     CONFIG = load_config(args.config or "./config.yml")
     max_retries = CONFIG.engine.online_moves.max_retries
```

MarkZH:
In line 194 of lichess-bot.py (`game_logging_configurer()`), the line `root.setLevel(level)` needs to be changed to `root.setLevel(logging.DEBUG)` so in-game debugging messages are logged to the automatic logs. Since this means that the `level` argument of that function is no longer neeeded, that simple change had a cascading effect on all functions that call it, resulting in the diff below.
```diff
diff --git a/lichess-bot.py b/lichess-bot.py
index a1ca0fe..02fa062 100644
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -185,13 +185,13 @@ def logging_listener_proc(queue: LOGGING_QUEUE_TYPE, level: int, log_filename: O
         queue.task_done()
 
 
-def game_logging_configurer(queue: Union[CONTROL_QUEUE_TYPE, LOGGING_QUEUE_TYPE], level: int) -> None:
+def game_logging_configurer(queue: Union[CONTROL_QUEUE_TYPE, LOGGING_QUEUE_TYPE]) -> None:
     """Configure the game logger."""
     h = logging.handlers.QueueHandler(queue)
     root = logging.getLogger()
     root.handlers.clear()
     root.addHandler(h)
-    root.setLevel(level)
+    root.setLevel(logging.DEBUG)
 
 
 def game_error_handler(error: BaseException) -> None:
@@ -236,7 +236,6 @@ def start(li: lichess.Lichess, user_profile: USER_PROFILE_TYPE, config: Configur
         lichess_bot_main(li,
                          user_profile,
                          config,
-                         logging_level,
                          challenge_queue,
                          control_queue,
                          correspondence_queue,
@@ -265,7 +264,6 @@ def log_proc_count(change: str, active_games: set[str]) -> None:
 def lichess_bot_main(li: lichess.Lichess,
                      user_profile: USER_PROFILE_TYPE,
                      config: Configuration,
-                     logging_level: int,
                      challenge_queue: MULTIPROCESSING_LIST_TYPE,
                      control_queue: CONTROL_QUEUE_TYPE,
                      correspondence_queue: CORRESPONDENCE_QUEUE_TYPE,
@@ -309,8 +307,7 @@ def lichess_bot_main(li: lichess.Lichess,
                       "config": config,
                       "challenge_queue": challenge_queue,
                       "correspondence_queue": correspondence_queue,
-                      "logging_queue": logging_queue,
-                      "logging_level": logging_level}
+                      "logging_queue": logging_queue}
 
     recent_bot_challenges: defaultdict[str, list[Timer]] = defaultdict(list)
 
@@ -547,8 +544,7 @@ def play_game(li: lichess.Lichess,
               config: Configuration,
               challenge_queue: MULTIPROCESSING_LIST_TYPE,
               correspondence_queue: CORRESPONDENCE_QUEUE_TYPE,
-              logging_queue: LOGGING_QUEUE_TYPE,
-              logging_level: int) -> None:
+              logging_queue: LOGGING_QUEUE_TYPE) -> None:
     """
     Play a game.
 
@@ -560,9 +556,8 @@ def play_game(li: lichess.Lichess,
     :param challenge_queue: The queue containing the challenges.
     :param correspondence_queue: The queue containing the correspondence games.
     :param logging_queue: The logging queue. Used by `logging_listener_proc`.
-    :param logging_level: The logging level. Either `logging.INFO` or `logging.DEBUG`.
     """
-    game_logging_configurer(logging_queue, logging_level)
+    game_logging_configurer(logging_queue)
     logger = logging.getLogger(__name__)
 
     response = li.get_game_stream(game_id)
```

AttackingOrDefending:
I also added `%(filename)s` and `(lineno)d` which was already included in the console logger but not in the file logger.

PR #738, opened 2023-06-14
Print traceback & function parameters when backing off

AttackingOrDefending:
This will make debugging easier. For example in #732 we only get `2023-06-13 22:00:20,075 backoff INFO Backing off play_game(...) for 3.3s (ValueError: end of string while looking for conversion specifier)` instead of the full traceback.


MarkZH:
Go ahead and merge if this is ready.

PR #744, opened 2023-06-19
Deprecate move quality "good" for EGTBs

AttackingOrDefending:
Move quality `good` can lead to 3-fold repetition. Example: https://lichess.org/HoJ22Ge7/white#130

### DIFF
--- a/config.py
+++ b/config.py
@@ -267,8 +267,13 @@ def validate_config(CONFIG: CONFIG_DICT_TYPE) -> None:
             online_section = (CONFIG["engine"].get(section) or {}).get(subsection) or {}
             config_assert(online_section.get("move_quality") != "suggest" or not online_section.get("enabled"),
                           f"XBoard engines can't be used with `move_quality` set to `suggest` in {subsection}.")
-            if online_section.get("move_quality") == "good":
-                raise DeprecationWarning("`move_quality` `good` is deprecated and will be removed soon.")
+
+    for section, subsection in (("online_moves", "online_egtb"),
+                                ("lichess_bot_tbs", "syzygy"),
+                                ("lichess_bot_tbs", "gaviota")):
+        online_section = (CONFIG["engine"].get(section) or {}).get(subsection) or {}
+        if online_section.get("move_quality") == "good":
+            logger.warning(DeprecationWarning(f"`move_quality` `good` in {subsection} is deprecated and will be removed soon."))
 
     matchmaking = CONFIG.get("matchmaking") or {}
     matchmaking_enabled = matchmaking.get("allow_matchmaking") or False
--- a/config.py
+++ b/config.py
@@ -267,6 +267,8 @@ def validate_config(CONFIG: CONFIG_DICT_TYPE) -> None:
             online_section = (CONFIG["engine"].get(section) or {}).get(subsection) or {}
             config_assert(online_section.get("move_quality") != "suggest" or not online_section.get("enabled"),
                           f"XBoard engines can't be used with `move_quality` set to `suggest` in {subsection}.")
+            if online_section.get("move_quality") == "good":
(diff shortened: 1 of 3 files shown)


engine_wrapper.py
```diff
@@ -875,6 +875,8 @@ def get_online_egtb_move(li: lichess.Lichess, board: chess.Board, game: model.Ga
         return None, -3
 
     quality = online_egtb_cfg.move_quality
+    if quality == "good":
+        logger.warning(DeprecationWarning("`move_quality` `good` is deprecated and will be removed soon."))
```
MarkZH, line 879: Do we need to warn on every move or only after loading the config file?

MarkZH (approved):

MarkZH:
I don't understand the remaining changes in engine_wrapper.py. Where is the `KeyError` raised and what does the code in the `except` block do?

lucasart:
The feature works as expected. But its implications are often misunderstood. The problem is that choosing a move that keeps WDL unchanged is insufficient to guarantee progress. Hence DTZ tables.

OTOH following the best algorithm, even though results in perfect conversion from the game result standpoint, may choose to convert in unnaturally convoluted ways.

Hence the hybrid idea of filtering for you the WDL optimal moves, and letting the engine search the best one, so that it can play a quick and natural finish instead of being enslaved to the slow (but guaranteed correct) DTZ minimisation.

It's basically a trade off.

PS: Why deprecte and not remove? If you want people to fix their config.yml you should remove and throw an error.

lucasart:
The method implemented by top engines, is to follow "best", unless a mate is found using "good" (and searching for a mate).

This requires syzygy code inside the engine to probe WDL and DTZ, of course. Perhaps this could be piloted by an external python script. But I would say this is too much complexity. Best to simply remove the "good" feature completely, and let those who want fine grained logic code it inside their engine.

lucasart:
Or just leave as is, and accept that "good" is a trade off.

MarkZH:
@lucasart We do a deprecation step any time a feature is removed or some change requires the user to change their setup--config settings or python version, for two examples. We don't do this when adding features because new features are off by default and the user doesn't have to do anything to keep their setup the same. The deprecation warning on startup gives users a chance to let us know if they disagree with what we are planning, just like you're doing right now.

I'll leave your other questions to @AttackingOrDefending, who has written most of the book and tablebase code. I don't use those sources with my bot.

AttackingOrDefending:
@MarkZH Syzygy TBs have `.rtbz` files which store DTZ (distance to zeroing) and `.rtbw` files which store WDL (win draw loss). The `except` block catches the `KeyError` exception raised by `python-chess` when it doesn't find `.rtbz` files. In this case, we are going to use only `.rtbw` files, to filter the moves and tell the engine to search onlly the moves with the best WDL. If `KeyError` is raised again then it means that there are no `.rtbw` files for this endgame.

config.py
```diff
@@ -268,6 +268,14 @@ def validate_config(CONFIG: CONFIG_DICT_TYPE) -> None:
             config_assert(online_section.get("move_quality") != "suggest" or not online_section.get("enabled"),
                           f"XBoard engines can't be used with `move_quality` set to `suggest` in {subsection}.")
 
+    for section, subsection in (("online_moves", "online_egtb"),
+                                ("lichess_bot_tbs", "syzygy"),
+                                ("lichess_bot_tbs", "gaviota")):
+        online_section = (CONFIG["engine"].get(section) or {}).get(subsection) or {}
+        if online_section.get("move_quality") == "good":
```
MarkZH, line 275: Do we want to print the warning if `online_section.get("enabled") == False`?

AttackingOrDefending: The default value is `best`, so they must have changed it to `good`. I think it is better to warn them, because they could enable it in the future and get unexpected results/errors.

MarkZH (approved):
Go ahead and merge if this is ready.

Issue #749, opened 2023-06-21
UnboundLocalError: local variable 'move' referenced before assignment


TheYoBots:
Relevant normal logs:
```
[06/21/23 20:28:35] INFO     Backing off play_game(...) for       _common.py:105
                             189.6s (UnboundLocalError: local                   
                             variable 'move' referenced before                  
                             assignment) 
```
Relavant verbose logs:
```
                    INFO     Backing off play_game(...) for 1.0s  _common.py:105
                             (UnboundLocalError: local variable                 
                             'move' referenced before assignment)               
                    DEBUG    Backing off 1.0 seconds after 1 tries lichess.py:62
                             calling function <function play_game               
                             at 0x7f72f40d6af0> with args () and                
                             kwargs {'li': <lichess.Lichess object              
                             at 0x7f72f5bd6070>, 'control_queue': ...
                             ...
                    DEBUG    Exception: Traceback (most recent     lichess.py:64
                             call last):                                        
                             File                                             
                             "/usr/local/lib/python3.9/dist-packag              
                             es/backoff/_sync.py", line 105, in                 
                             retry                                              
                                 ret = target(*args, **kwargs)                  
                               File "//lichess/lichess-bot.py",                 
                             line 620, in play_game                             
                                 engine.play_move(board,                        
                               File "/lichess/engine_wrapper.py",               
                             line 160, in play_move                             
                                 best_move = get_book_move(board,               
                             game, polyglot_cfg)                                
                               File "/lichess/engine_wrapper.py",               
                             line 704, in get_book_move                         
                                 if move is not None:                           
                             UnboundLocalError: local variable                  
                             'move' referenced before assignment
```

MarkZH:
Can you post the `engine: polyglot:` section of your config?

TheYoBots:
```yml
polyglot:
    enabled: true            # Activate polyglot book.
    book:
      standard:              # List of book file paths for variant standard.
        - lichess/book1.bin
        - lichess/book2.bin
        - lichess/book3.bin
        - lichess/book3.bin
#     atomic:                # List of book file paths for variant atomic.
#       - engines/atomicbook1.bin
#       - engines/atomicbook2.bin
#     etc.
#     Use the same pattern for 'giveaway' (antichess), 'crazyhouse', 'horde', 'kingofthehill', 'racingkings' and '3check' as well.
```

The dir is correct. When I put a wrong dir I get the following error:
```
[...]

PR #745, opened 2023-06-19
Add lichess opening explorer

AttackingOrDefending:
Adds the lichess opening explorer to the available online opening books. In `README.md` `source` is added twice because it is completely different than `online_egtb`'s `source` parameter.

PR #751, opened 2023-06-23
Check for valid DB lookup options

MarkZH:
Since the method of move selection from chess DBs of various sorts is chosen from a list, make sure the value in the user's config is one of those values.

Inspired by issues #743 and #749.


MarkZH:
This should be merged after #744 and #745 so it can be adjusted for all databases.

config.py
```diff
@@ -290,6 +290,40 @@ def validate_config(CONFIG: CONFIG_DICT_TYPE) -> None:
...
+                         "online_egtb": ["good", "best", "suggest"]}
+    for db_name, valid_selections in selection_choices.items():
+        is_online = db_name != "polyglot"
+        db_section = (CONFIG["engine"].get("online_moves") or {}) if is_online else CONFIG["engine"]
+        db_config = db_section.get(db_name)
+        select_key = "selection" if db_name == "polyglot" else "move_quality"
+        selection = db_config.get(select_key)
+        select = f"{'online_moves:' if is_online else ''}{db_name}:{select_key}"
+        config_assert(selection in valid_selections,
+                      f"'{selection}' is not a valid 'engine:{select}' value. "
```
AttackingOrDefending, line 305: The single quotes in `'{selection}'` and `'engine:{select}'` should probably change to backticks like on line 313 for consistency. Or change the backticks in line 313 to singe quotes.

AttackingOrDefending (approved):

PR #764, opened 2023-06-28
Fix restart

AttackingOrDefending:
Closes logging handlers. lichess-bot would get an error, because it would attempt to move `recent.log` to `old.log` but `recent.log` was used by a file handler.

I added `logger.debug(f"Exception: {traceback.format_exc()}")` because lichess-bot didn't restart because it was offline on lichess.org but because `watch_control_stream` added `{"type": "terminated"}` to the `control_queue`.

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -20,7 +20,6 @@ import copy
 import math
 import sys
 import yaml
-import traceback
 from config import load_config, Configuration
 from conversation import Conversation, ChatLine
 from timer import Timer
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -20,6 +20,7 @@ import copy
 import math
 import sys
 import yaml
+import traceback
 from config import load_config, Configuration
 from conversation import Conversation, ChatLine
 from timer import Timer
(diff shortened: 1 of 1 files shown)


lichess-bot.py
```diff
@@ -97,6 +98,7 @@ def watch_control_stream(control_queue: CONTROL_QUEUE_TYPE, li: lichess.Lichess)
                 else:
                     control_queue.put_nowait({"type": "ping"})
         except Exception:
+            logger.debug(f"Exception: {traceback.format_exc()}")
```
MarkZH, line 101: This logging line won't work (at least, not in Windows) because it is in a different thread. Look at the first lines of `play_game()` to see what logging with threads looks like. You can add the exception to the `terminated` message and then handle it in the `if event["type"] == "terminated":` block of `lichess_bot_main()`--line 321. Something like this:
```python
def watch_control_stream(control_queue: CONTROL_QUEUE_TYPE, li: lichess.Lichess) -> None:
    """Put the events in a queue."""

    error = None
    while not terminated:
        try:
            response = li.get_event_stream()
            lines = response.iter_lines()
            for line in lines:
                if line:
                    event = json.loads(line.decode("utf-8"))
                    control_queue.put_nowait(event)
                else:
                    control_queue.put_nowait({"type": "ping"})
        except Exception as err:
            error = err
            break

    control_queue.put_nowait({"type": "terminated", "error": error})
```
Later, in lichess_bot_main(),
```python
            if event["type"] == "terminated":
                restart = True
                logger.debug("Terminating exception:", exc_info=event["error"])
                control_queue.task_done()
                break
```

The `exc_info` parameter will capture the exception information for logging, so you don't need to use `traceback`. If you want to print the stack trace as well, add `stack_info=event["error"]` to the `logger.debug` call.

MarkZH (approved):

PR #769, opened 2023-07-03
Separate Standard and Chess960 Books

TheYoBots:
Previously to add a book for chess960, it had to be added in the 'standard' section. With this change there's a separate section for standard and chess960 opening books.

### DIFF
--- a/config.yml.default
+++ b/config.yml.default
@@ -19,7 +19,7 @@ engine:                      # Engine settings.
 #       - engines/atomicbook1.bin
 #       - engines/atomicbook2.bin
 #     etc.
-#     Use the same pattern for 'giveaway' (antichess), 'crazyhouse', 'horde', 'kingofthehill', 'racingkings' and '3check' as well.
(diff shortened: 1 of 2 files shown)


TheYoBots:
I'm not really sure as to how I should go about fixing flake8 errors, but a possible fix could be this:
```py
variant = "chess960" if board.chess960 else str(board.uci_variant)
```
and in the `config.yml` I could change `standard` to `chess` which honestly isn't too confusing right?

AttackingOrDefending:
The solution you proposed would require adding additional code for backwards compatibility. Another not so clean way to fix this is:
```python
if board.chess960:
    variant = "chess960"
else:
    variant = str(board.uci_variant).replace("chess", "standard")
```
In the code that you currently have, line 690 should change to `variant = str(board.uci_variant)`.

TheYoBots:
What about this:
```py
if board.chess960:
    variant = "chess960"
else:
    variant = "standard" if board.uci_variant == "chess" else str(board.uci_variant)
```
Though I don't know if it would pass the flake8 test... I'm not sure how it works.

TheYoBots:
I fixed all the merge conflicts. Is there any other change that I should make?

AttackingOrDefending (approved):

MarkZH (approved):

Issue #773, opened 2023-07-05
How could I get my engine to receive the Wtime and Btime of a lichess game?


vibranium21:
I am improving my engine's time management, and I love to test my engine out on lichess. My engine is not uci. So I use the strategies.py to connect the bot to lichess. Is there a way to receive wtime and btime without making the engine Uci?

AttackingOrDefending:
Your engine already gets the time but it is under `*args` which is why it isn't shown. If you remove `*args` you can use `time_limit`.
```python
class MyEngine(ExampleEngine):
    def search(self, board: chess.Board, time_limit: chess.engine.Limit, ponder: bool, draw_offered: bool,
               root_moves: MOVE) -> chess.engine.PlayResult:
        wtime = time_limit.white_clock
        btime = time_limit.black_clock
        winc = time_limit.white_inc
        binc = time_limit.black_inc
        CODE HERE
        return PlayResult(engine_move, None)
```

MarkZH:
@vibranium21 We've added another example homemade engine in strategies.py to demonstrate how to get the game clock and other information to your engine. Essentially, you change the arguments to the search method. Please reopen this issue if you have trouble with this in the future.

PR #774, opened 2023-07-05
Add one more example homemade engine

AttackingOrDefending:
Add one more example homemade engine so that it better showcases its capablities. Inpsired by #773.

### DIFF
--- a/strategies.py
+++ b/strategies.py
@@ -57,11 +57,24 @@ class FirstMove(ExampleEngine):
 
 
 class ComboEngine(ExampleEngine):
-    """Get a move using multiple different methods."""
+    """
+    Get a move using multiple different methods.
+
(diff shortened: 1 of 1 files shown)


strategies.py
```diff
@@ -53,3 +54,31 @@ def search(self, board: chess.Board, *args: Any) -> PlayResult:
         moves = list(board.legal_moves)
         moves.sort(key=str)
         return PlayResult(moves[0], None)
+
+
+class ComboEngine(ExampleEngine):
+    """Get a move using multiple different methods."""
```
MarkZH, line 60: I think it's important to mention that this demonstrates a `search()` method that can use time limits and other game information.

MarkZH (approved):

PR #775, opened 2023-07-06
Option to require increment for bullet games

### LINKED ISSUE #770: Add bullet with increment only option
Can you add this option to prevent 1+0 challenges (farming) for low-end bots ?

MarkZH:
This option prevents opponents from taking advantage of engines that time out often during short games by requiring challenges to have a minimum one-second increment for bullet games.

Closes #770


MarkZH:
Right now, this applies to all challengers (human and bot). Let me know if this should only apply to bots.

EmptikBest:
@MarkZH Yeah I think it would be better if it only applied to bots since what human can time-out a bot anyway?

AttackingOrDefending (approved):

AttackingOrDefending:
The comment in `config.yml.default` should be changed to make it obvious that this only applies in bots. You can merge this after changing the comment. Also, add this option to the wiki.

PR #789, opened 2023-07-29
Unbridle fake_think_time

lucasart:
This adresses 4 problems I have experienced with the current formula:
- min(initial, remaining) is replaced by remaining. Indeed, time usage should increase when the remaining time increases above the initial time, as a result of accumulated increments.
- the accel formula is replaced by a smooth one, with the terminal condition preserved: accel = 1/3 for 120 half-moves played, but that is unfloored and slowly decayes towards zero beyond that. Just a slight difference that accel starts immediately from 10, instead of 20. This is more coherent, as the whole logic starts at 10 (for which accel=1, then accel decays by 1% for each ply).
- remove the cap of 5 seconds, which is an arbitrary value that corresponds to nothing. In long time control this made no sense.
- increase the fraction of remaining time (before accel factor) from 0.15 ~= 1/67 to 0.025 = 1/40.

Hopefully, the new formulas are more elegant and easier to understand.

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -673,9 +673,9 @@ def say_hello(conversation: Conversation, hello: str, hello_spectators: str, boa
 def fake_thinking(config: Configuration, board: chess.Board, game: model.Game) -> None:
     """Wait some time before starting to search for a move."""
     if config.fake_think_time and len(board.move_stack) > 9:
-        delay = min(game.clock_initial, game.my_remaining_seconds()) * 0.015
-        accel = 1 - max(0, min(100, len(board.move_stack) - 20)) / 150
-        sleep = min(5, delay * accel)
+        delay = game.my_remaining_seconds() * 0.025
+        accel = 0.99 ** (len(board.move_stack) - 10)
+        sleep = delay * accel
         time.sleep(sleep)
 
 


lucasart:
For real life validation of how this patch behaves, you can have a look at recent games (since the time of this writing) from my bot accounts: Demolito_L5 and Demolito_L6. lichess shows time usage per move, and time remaingin graph, hopefully you can get a good sense by looking at games played at various time control (these bots are setup to play lots of varied (time, inc) pairs).

lucasart:
Hum ... Now I am getting quite a few time losses. It seems to happen when multiple games are being played concurrently.

I wonder if this is a bug that gets _exposed_ by this patch, which the current formula was made to hide (by being extremely conservative). Isn't this time.sleep() stopping the whole world instead of only the one game in which it occurs?

MarkZH:
The function `time.sleep()` should only pause the thread it is called in. But, `fake_thinking()` is called in every game thread, so every game should have this pause on every move.

I think there are two crucial differences between your code and the original. First, the original code has a maximum sleep time of 5 seconds, whereas your code has unbounded sleep. Second, in the original code, the sleep time gets shorter as the game goes on, whereas your code causes longer sleeps as the game progresses.

For reference, I believe the following is equivalent to the original code. I just tried to simplify the math.
```python
def fake_thinking(config: Configuration, board: chess.Board, game: model.Game) -> None:
    """Wait some time before starting to search for a move."""
    if config.fake_think_time and len(board.move_stack) > 9:

        def clamp(low, high, value):
            return max(low, min(high, value))

        delay = min(game.clock_initial, game.my_remaining_seconds())
        accel = 130 - clamp(20, 120, len(board.move_stack))
        sleep = min(5, delay * accel * 1e-4)
        time.sleep(sleep)
```

lucasart:
> First, the original code has a maximum sleep time of 5 seconds, whereas your code has unbounded sleep.

Yes, as explained above, this is a feature. I see no logical reason for a fixed cap.

> Second, in the original code, the sleep time gets shorter as the game goes on, whereas your code causes longer sleeps as the game progresses.
 the original code.

No, my `accel` formula starts at 1, like the original, and goes down to 0.33 when plies=100, like the original. But it continues to decay towards zero beyond that.

PS: my erratic time usage/losses were unrelated. I reimplemented the fake sleep in my engine, and experienced the same problem. What fixed it was to acquire the wake lock on Android (to prevent it from sending termux to sleep)

AttackingOrDefending:
The new fake thinking time decays normally. It usually returns a higher sleep value than the original one but I don't think that is a problem.
Figure_1
Figure_2
Figure_3

MarkZH:
I see. I misread the exponentiation operator `**` as multiplication `*`.

I think the change results in too large an increase in fake thinking time. The following plot compares the time left on the clock for a 10min+0 game with an engine that spends 1 second thinking about each move.
Plot of time remaining comparison between original code and this PR (intitial time 10 minutes, engine search time 1 second)
In the middle game, an engine will have 106 fewer seconds on the clock than before.

By adjusting the `delay` line to `delay = remaining_seconds * 0.015`, engines will have a similar amount of time for thinking until the endgame. The maximum difference in the middle game is 30 seconds.
Plot of time remaining comparison between original code and an adjustment of this PR (intitial time 10 minutes, engine search time 1 second)

lucasart:
> I think the change results in too large an increase in fake thinking time. The following plot compares the time left on the clock for a 10min+0 game with an engine that spends 1 second thinking about each move.

Ok. I see where you are coming from. My assumption was that one should only use fake_think for engines that play instantly. Those are typically the "weak" bots obtained by limiting depth/node to such a degree that engine is bad enough to lose against humans.

Of course we use 0.15 instead of 0.25, to avoid rocking the boat. But I think weare missing anopportunity to properly fixthe logic.

What I would recommend is to combine both times instead of adding them: if engine takes T1 secons, and fake_think requires T2 seconds, we should do the time.sleep after the engine think, and wait for `max(T1, T2) - T1` (which may be zero, no need to call sleep there).

What do you think?

AttackingOrDefending:
I think it is a good idea to sleep after the engine thinks. It was also mentioned in #295 but it wasn't implemented.

MarkZH (approved):

AttackingOrDefending (approved):

PR #796, opened 2023-08-04
combine fake_think_time with time already consumed

lucasart:
There are 3 different delays:
1/ time spend by the engine
2/ rate_limiting_delay
3/ fake_think_time formula

The problem is each of these is unaware of the other 2, and we just add them all up. This is a recipie for disaster. Instead we want to spend the max.

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -107,7 +107,8 @@ class EngineWrapper:
                   can_ponder: bool,
                   is_correspondence: bool,
                   correspondence_move_time: int,
-                  engine_cfg: config.Configuration) -> None:
+                  engine_cfg: config.Configuration,
+                  min_time: float) -> None:
         """
         Play a move.
 
@@ -120,6 +121,7 @@ class EngineWrapper:
         :param is_correspondence: Whether this is a correspondence or unlimited game.
         :param correspondence_move_time: The time the engine will think if `is_correspondence` is true.
         :param engine_cfg: Options for external moves (e.g. from an opening book), and for engine resignation and draw offers.
+        :param min_time: minimum time to spend, in seconds.
         :return: The move to play.
         """
         polyglot_cfg = engine_cfg.polyglot
@@ -155,6 +157,11 @@ class EngineWrapper:
                 time_limit = game_clock_time(board, game, start_time, move_overhead)
 
             best_move = self.search(board, time_limit, can_ponder, draw_offered, best_move)
+
+        # Heed min_time
+        elapsed = (time.perf_counter_ns() - start_time) / 1e9
+        if elapsed < min_time:
+            time.sleep(min_time - elapsed)
 
         self.add_comment(best_move, board)
         self.print_stats()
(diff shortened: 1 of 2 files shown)


lucasart:
Sorry, this is not commitable as is. The waiting must be done in `play_move()` before `li.make_move()` is called. I will fix this and report back, but hopefully this simple code shows you what I want to do.

AttackingOrDefending:
As mentioned in https://github.com/lichess-bot-devs/lichess-bot/pull/789#issuecomment-1665182850, we can sleep for `rate_limiting_delay` in the opponents time. Since you are going to move this function inside `engine_wrapper`, we can store when the last move was played and see if `delay_seconds` seconds have passed before sleeping. Something like:
```python
combined_delay = max(fake_think_time, last_move_played + rate_limiting_delay - time_now, consumed_sec) - consumed_sec
```
Also, `consumed_sec` doesn't have to be inside `max`, since you already check if `combined_delay` is greater than 0.

MarkZH:
As long as everything is being moved into `EngineWrapper`, I think something like the following would be the simplest
```python
    def play_move(self, ...)
        fake_thinking_timer = Timer(fake_think_time())

        # search for move ...

        if best_move.resigned and len(board.move_stack) >= 2:
            li.resign(game.id)
        else:
            time.sleep(max(fake_thinking_timer.time_until_expiration(), 0))
            li.make_move(game.id, best_move)
            time.sleep(self.rate_limiting_delay)
```

lucasart:
Updated the code, with a slightly less ambitious version: do not touch the rate_limiting_delay (broken) logic. On the other hand, I have improved the fake_think_time formula, to account for move_overhead, which was not done before.

PS: As I look at the code, I find it extremely confusing how we mix `s,ms,ns` without a consistent naming convention. I would advise to follow Python's time library: no suffix for seconds, _ns for nanoseconds, and (by extension) _ms for milliseconds. But that would be a much bigger PR...

MarkZH:
I agree that time can be better handled. I think a better solution is to use `datetime` and `timedelta` objects instead of `float` and `int` so nobody has to keep track of units. Getting values from the configuration file would then look like `datetime.timedelta(milliseconds=config.move_overhead)` so there's no ambiguity and no need for `1e9` and `1000` values everywhere. `Timer` objects from `timer.py` could also be helpful.

engine_wrapper.py
```diff
@@ -120,6 +121,7 @@ def play_move(self,
         :param is_correspondence: Whether this is a correspondence or unlimited game.
         :param correspondence_move_time: The time the engine will think if `is_correspondence` is true.
         :param engine_cfg: Options for external moves (e.g. from an opening book), and for engine resignation and draw offers.
+        :param min_time: minimum time to spend, in seconds.
```
AttackingOrDefending, line 124: Capitalize the m. `Minimum time to spend, in seconds.`

lucasart: fixed

MarkZH (approved):

AttackingOrDefending (approved):

PR #797, opened 2023-08-05
Add source field to the engine info

### LINKED ISSUE #792: Eval during move taken from Book, Online, TBs must be different
When the bot is playing a move taken from `opening_book` no eval gets printed. And when the bot is playing a move taken from any online resource (opening explorer, online tbs, etc.) the move score is converted into a format to print eval in the same way it is printed otherwise(?).
What I suggest is that eval should be printed as follows:
1. Opening Book moves:
```
Opening Book: 1.e2e4
```
2. Moves taken from chessdb book:
```
Online ChessDB Book: 1.e2e4 
# and then maybe the chessdb score can be printed(?)
```
3. Moves taken from lichess cloud analysis:
```
Lichess Cloud Analysis: 1.e2e4
# and then maybe the data like depth, knodes given by the endpoint
```
4. Moves taken from Lichess Opening Explorer
```
Lichess Opening Explorer: 1.e2e4
# and then info like which database is was taken from (master, player or lichess games). And if player from which player(?)
```
5. Lichess/ChessDB online tablebases:
```
Lichess Online EGTB: 54.Qc2#
# or ChessDB instead of lichess if got from chessdb. and followed by wdl score (1, 0, -1) and dtm (probably something like Moves to mate: 3
```
6. Bot Tablebases:
```
Syzygy EGTB: 54.Qc2 # Gaviota instead of Syzygy
# followed by wdl score and dtm like in online tb
```

I got this idea after seeing a few BOTs that use BotLi: https://github.com/Torom/BotLi

AttackingOrDefending:
Adds a new field in the engine info called `source`. Possible options are: `Engine`, `Opening Book`, `Lichess EGTB`, `ChessDB EGTB`, `Lichess Opening Explorer (Masters)`, `Lichess Opening Explorer (Player)`, `Lichess Opening Explorer (Lichess)`, `ChessDB`, `Lichess Cloud Analysis`, `Syzygy EGTB`, `Gaviota EGTB`.

closes #792

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -307,6 +307,19 @@ class EngineWrapper:
             return f"{round(number / 1e3, 1)}K"
         return str(number)
 
+    def to_readable_value(self, stat: str, info: MOVE_INFO_TYPE) -> str:
+        readable: dict[str, Callable[[Any], str]] = {"score": self.readable_score, "wdl": self.readable_wdl,
+                                                     "hashfull": lambda x: f"{round(x / 10, 1)}%",
+                                                     "nodes": self.readable_number,
+                                                     "nps": lambda x: f"{self.readable_number(x)}nps",
+                                                     "tbhits": self.readable_number,
+                                                     "cpuload": lambda x: f"{round(x / 10, 1)}%"}
+
+        def identity(x: Any) -> str:
+            return str(x)
+
+        return str(readable.get(stat, identity)(info[stat]))
+
     def get_stats(self, for_chat: bool = False) -> list[str]:
         """
         Get the stats returned by the engine.
@@ -315,19 +328,6 @@ class EngineWrapper:
         """
         can_index = self.move_commentary and self.move_commentary[-1]
         info: MOVE_INFO_TYPE = self.move_commentary[-1].copy() if can_index else {}
-
-        def to_readable_value(stat: str, info: MOVE_INFO_TYPE) -> str:
-            readable: dict[str, Callable[[Any], str]] = {"score": self.readable_score, "wdl": self.readable_wdl,
-                                                         "hashfull": lambda x: f"{round(x / 10, 1)}%",
-                                                         "nodes": self.readable_number,
-                                                         "nps": lambda x: f"{self.readable_number(x)}nps",
-                                                         "tbhits": self.readable_number,
-                                                         "cpuload": lambda x: f"{round(x / 10, 1)}%"}
-
-            def identity(x: Any) -> str:
-                return str(x)
-
-            return str(readable.get(stat, identity)(info[stat]))
 
         def to_readable_key(stat: str, value: Any) -> str:
             readable = {"wdl": "winrate", "ponderpv": "PV", "nps": "speed", "score": "evaluation"}
@@ -344,7 +344,7 @@ class EngineWrapper:
 
         stats = ["Source", "Evaluation", "Winrate", "Depth", "Nodes", "Speed", "Pv"]
         if for_chat and "Pv" in info:
-            bot_stats = [f"{stat}: {to_readable_value(stat, info)}"
+            bot_stats = [f"{stat}: {self.to_readable_value(stat, info)}"
                          for stat in stats if stat in info and stat != "Pv"]
             len_bot_stats = len(", ".join(bot_stats)) + PONDERPV_CHARACTERS
             ponder_pv = info["Pv"].split()
@@ -358,7 +358,7 @@ class EngineWrapper:
                 pass
             if not info["Pv"]:
                 info.pop("Pv")
-        return [f"{stat}: {to_readable_value(stat, info)}" for stat in stats if stat in info]
+        return [f"{stat}: {self.to_readable_value(stat, info)}" for stat in stats if stat in info]
 
     def get_opponent_info(self, game: model.Game) -> None:
         """Get the opponent's information and sends it to the engine."""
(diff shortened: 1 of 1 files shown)


TheYoBots:
Is this ready for testing?

engine_wrapper.py
```diff
@@ -316,41 +330,36 @@ def get_stats(self, for_chat: bool = False) -> list[str]:
...
+                stat = "source"
             return stat.title()
 
-        stats = ["score", "wdl", "depth", "nodes", "nps", "ponderpv"]
-        if for_chat and "ponderpv" in info:
-            bot_stats = [f"{to_readable_key(stat)}: {to_readable_value(stat, info)}"
-                         for stat in stats if stat in info and stat != "ponderpv"]
+        info = {to_readable_key(key, value): value for (key, value) in info.items()}
+        if "Source" not in info:
+            info["Source"] = "Engine"
```
TheYoBots, line 344: This should be changed to this:
```py
        info = {to_readable_key(stat, value): value for (stat, value) in info.items()}
        if "Source" not in info:
            info["Source"] = "Engine"
```
Otherwise the `info["Source"]` field will always be `"Engine"`

MarkZH: How does changing the name of the variable from `key` to `stat` change anything?

engine_wrapper.py
```diff
@@ -316,41 +330,36 @@ def get_stats(self, for_chat: bool = False) -> list[str]:
...
-
-        def to_readable_key(stat: str) -> str:
+        def to_readable_key(stat: str, value: Any) -> str:
             readable = {"wdl": "winrate", "ponderpv": "PV", "nps": "speed", "score": "evaluation"}
             stat = readable.get(stat, stat)
+            if stat == "string" and value in ["Opening Book", "Lichess Opening Explorer (Masters)",
+                                              "Lichess Opening Explorer (Player)", "Lichess Opening Explorer (Lichess)",
+                                              "Lichess EGTB", "ChessDB EGTB", "Lichess Cloud Analysis", "ChessDB"]:
+                stat = "source"
             return stat.title()
```
MarkZH, line 340: I think there's an easier way to do this without needing to list every possible source. In steps:
1. Prefix all of the `"string"` comments with `"lichess-bot-source:"`. For example, in line 668: `return chess.engine.PlayResult(move, None, {"string": "lichess-bot-source:Opening Book"})`
2. Rename this inner function `def to_readable_item()` with the same arguments.
3. The `if`-statement becomes `if stat == "string" and value.startswith("lichess-bot-source:"):`
4. If so, `stat = "source"` and `value = value.split(":", 1)[1]`
5. `return stat.title(), value`
6. Finally, `info = dict(to_readable_item(key, value) for (key, value) in info.items())`

engine_wrapper.py
TheYoBots: I was testing this code and I found 2 errors:
1. Source is always "Engine", which gets fixed by my previous comment.
2. `to_readable_value` doesn't really seem to execute properly. My stats are printed like this:
```
[08/07/23 10:27:31] INFO     Source: Engine                engine_wrapper.py:233
                    INFO     Evaluation: PovScore(Cp(0),   engine_wrapper.py:233
                             WHITE)                                             
                    INFO     Depth: 51                     engine_wrapper.py:233
                    INFO     Nodes: 22616579               engine_wrapper.py:233
                    INFO     Speed: 1360313                engine_wrapper.py:233
                    INFO     Pv: 60. Rg8+ Kh5 61. Re5+ Kh6 engine_wrapper.py:233
                             62. Rge8 Kg6 63. R5e7 Kh5 64.                      
                             Rh8+ Kg6 65. Rg8+ Kh6 66.                          
                             Rh8+
```

engine_wrapper.py
AttackingOrDefending: Done.

TheYoBots:
The latest commit fixes all issue I faced with this. Another minor change could be to edit this part:
```py
        readable: dict[str, Callable[[Any], str]] = {"Evaluation": self.readable_score, "Winrate": self.readable_wdl,
                                                     "Hashfull": lambda x: f"{round(x / 10, 1)}%",
                                                     "Nodes": self.readable_number,
                                                     "Speed": lambda x: f"{self.readable_number(x)}nps",
                                                     "Tbhits": self.readable_number,
                                                     "Cpuload": lambda x: f"{round(x / 10, 1)}%"}
```
to something like this:
```py
        readable: dict[str, Callable[[Any], str]] = {"Evaluation": self.readable_score, "Winrate": self.readable_wdl,
                                                     "Nodes": self.readable_number,
                                                     "Speed": lambda x: f"{self.readable_number(x)}nps"}
```

Also `Winrate` never gets printed anywhere, yet it is present here.

AttackingOrDefending:
These options aren't printed by default, but they can be added in line 345. For example, you can change `stats = ["Source", "Evaluation", "Winrate", "Depth", "Nodes", "Speed", "Pv"]` to `stats = ["Source", "Evaluation", "Winrate", "Depth", "Nodes", "Speed", "Hashfull", "Tbhits", "Pv"]`. They were added to support all the options in https://python-chess.readthedocs.io/en/latest/engine.html#chess.engine.InfoDict. To show `Winrate` you should add `UCI_ShowWDL: true` under `uci_options`. I think we should probably add it to `config.yml.default`. This option is supported by Stockfish and Leela, and probably more engines.

MarkZH (approved):

PR #800, opened 2023-08-09
Reset logging config before exiting

MarkZH:
Because the main thread logs to a queue that is then handled by a listener process, terminating the listener process stops logging for the rest of the program. This means that, if lichess-bot is stopped by an exception, then the exception logging at the bottom of the file (line 1029) does not do anything.

This change reverts the main thread back to direct logging before terminating the logging queue thread.


MarkZH:
This change can be tested by raising an exception within `lichess_bot_main`. Before this change, the exception isn't logged; afterwards, it is.

AttackingOrDefending (approved):

PR #801, opened 2023-08-10
Replace ints and floats with datetime.timedeltas

MarkZH:
This PR replaces all `int` and `float` data that represent time with proper types.
- Duration: bare `int` and `float` replaced by `datetime.timedelta`
- Stopwatch: two calls to `time.perf_counter()` replaced with `Timer` and calls to `time_since_reset()`
- `Timer` internals: replace `time.time()` (which can be altered by system clock changes like daylight saving time) with the monotonic `time.perf_counter()`
- Variables tagged with `_ms`, `_sec`, etc.: explicit conversions to `timedelta` with functions named `msec()`, `seconds()`, etc.

The main benefit of these changes is that there is no need to keep track of what units a given `int` or `float` variable is using to store its time. There is no more need for `* 1000` or `/ 1e9` or `* 60` operations to get numbers into common units. Everything is handled by the `timedelta` type. The result of `seconds(5) + msec(50)` is correctly computed without manually coded conversions.

Other bugs fixed:
- test_bot/lichess.py line 84: `wtime` used instead of `btime`

Note:
The only time variables that were not converted to `datetime.timedelta` are the time control variables in the functions that create a new challenge (`create_challenge()` and `choose_opponent()`). Since these functions select values from lists in the user's config file and then immediately send them to `Lichess.challenge()`, the round trip from `int` to `timedelta` and back to `int` seemed pointless.

### DIFF
--- a/test_bot/test_bot.py
+++ b/test_bot/test_bot.py
@@ -3,7 +3,6 @@ import pytest
 import zipfile
 import requests
 import time
-import datetime
 import yaml
 import chess
 import chess.engine
@@ -14,7 +13,7 @@ import stat
 import shutil
 import importlib
 import config
-from timer import Timer, to_seconds
+from timer import Timer, to_seconds, seconds
 from typing import Any
 if __name__ == "__main__":
     sys.exit(f"The script {os.path.basename(__file__)} should only be run by pytest.")
@@ -87,8 +86,8 @@ def thread_for_test() -> None:
     open("./logs/states.txt", "w").close()
     open("./logs/result.txt", "w").close()
 
-    start_time = datetime.timedelta(seconds=10)
-    increment = datetime.timedelta(seconds=0.1)
+    start_time = seconds(10)
+    increment = seconds(0.1)
 
     board = chess.Board()
     wtime = start_time
--- a/test_bot/test_bot.py
+++ b/test_bot/test_bot.py
@@ -3,6 +3,7 @@ import pytest
 import zipfile
 import requests
 import time
+import datetime
 import yaml
 import chess
 import chess.engine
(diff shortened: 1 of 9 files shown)


AttackingOrDefending:
Instaed of using `timedelta / msec(1)` which isn't that easy to understand what it does, we can use:
```python
class Timedelta(datetime.timedelta):
    def total_milliseconds(self) -> float:
        return self / msec(1)
```
Also, to allow us to use years we can use:
```python
def get_timedelta(years=0, days=0, seconds=0, microseconds=0, milliseconds=0, minutes=0, hours=0, weeks=0):
    return Timedelta(days=years * 365 + days, seconds=seconds, microseconds=microseconds, milliseconds=milliseconds,
                     minutes=minutes, hours=hours, weeks=weeks)
```
and instead of using `import datetime` we can use `from timer import get_timedelta`.

MarkZH:
I opted for a couple more functions instead of a whole new class. It seemed simpler.

Also, the first post of this PR has been updated to explain when the time control variables were not converted to use `datetime.timedelta`.

test_bot/test_bot.py
```diff
@@ -85,15 +87,15 @@ def thread_for_test() -> None:
...
+    start_time = datetime.timedelta(seconds=10)
+    increment = datetime.timedelta(seconds=0.1)
 
     board = chess.Board()
     wtime = start_time
     btime = start_time
 
     with open("./logs/states.txt", "w") as file:
-        file.write(f"\n{wtime},{btime}")
+        file.write(f"\n{wtime.total_seconds()},{btime.total_seconds()}")
```
AttackingOrDefending, line 98: Change `.total_seconds()` to `to_seconds(wtime or btime)`

AttackingOrDefending (approved):

test_bot/test_bot.py
AttackingOrDefending: You can marge this PR after changing this.

PR #805, opened 2023-08-12
Add UCI_ShowWDL by default

AttackingOrDefending:
Adds `UCI_ShowWDL` under `uci_options` so that the winrate is printed.

### DIFF
--- a/config.yml.default
+++ b/config.yml.default
@@ -90,6 +90,7 @@ engine:                            # Engine settings.
     Threads: 4                     # Max CPU threads the engine can use.
     Hash: 512                      # Max memory (in megabytes) the engine can allocate.
     SyzygyPath: "./syzygy/"        # Paths to Syzygy endgame tablebases that the engine reads.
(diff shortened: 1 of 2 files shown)


TheYoBots:
Isn't this handled by python-chess?
https://github.com/niklasf/python-chess/blob/35b961c17b7a6dcbad98f4f3bfb9df115a28db29/chess/engine.py#L1426

MarkZH:
I think stopping lichess-bot because of an unsupported option is better. Users need to know when they send an option that their engine doesn't understand, and quitting with an error message lets them know that there is something that needs fixing. Printing a warning at the start of a game is easy to miss.

Maybe we could start up an engine at the end of `load_config()`? That would cause lichess-bot to quit with an error before any games are started.

AttackingOrDefending:
#811 implemented this, so I removed the code and left only the `UCI_ShowWDL` line.

MarkZH (approved):

PR #810, opened 2023-08-13
Add sync-wiki workflow

TheYoBots:
This allows users to edit the wiki by simply making changes in the `./wiki/*` dir. I've imported all commits/changes made to the wiki and added them to the `wiki` dir and even added a new workflow that will run whenever there is a change in the `wiki` dir. The first workflow will give an error since there are no changes to the `wiki` dir as compared to the [wiki page](https://github.com/lichess-bot-devs/lichess-bot/wiki), but afterwards any changes made to the `wiki` dir will directly be reflected in the wiki.

If you don't want this error on first run then either I can make a change to the `wiki` dir now or I can add the tag `--allow-empty-commit` before commiting changes like this: `git commit --allow-empty-commit "Update wiki"`.

### DIFF
--- a/.github/workflows/sync-wiki.yml
+++ b/.github/workflows/sync-wiki.yml
@@ -10,6 +10,7 @@ on:
 
 jobs:
   sync:
+    if: github.repository != github.event.repository.full_name
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repository
--- a/.github/workflows/sync-wiki.yml
+++ b/.github/workflows/sync-wiki.yml
@@ -10,7 +10,6 @@ on:
 
 jobs:
   sync:
-    if: "!contains(github.event.head_commit.message, 'Auto update wiki/Home.md')"
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repository
--- a/.github/workflows/sync-wiki.yml
+++ b/.github/workflows/sync-wiki.yml
@@ -6,9 +6,11 @@ on:
       - master
     paths:
       - 'wiki/**'
+      - 'README.md'
 
 jobs:
   sync:
+    if: "!contains(github.event.head_commit.message, 'Auto update wiki/Home.md')"
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repository
@@ -18,12 +20,20 @@ jobs:
         with:
           repository: "${{ github.repository }}.wiki"
           path: "lichess-bot.wiki"
-      - name: Sync files to wiki
+      - name: Set github bot global config
+        run: |
+          git config --global user.email "actions@github.com"
+          git config --global user.name "actions-user"
+      - name: Sync README.md to wiki/Home.md
+        run: |
+          cp -r $GITHUB_WORKSPACE/README.md $GITHUB_WORKSPACE/wiki/Home.md
+          git add wiki/Home.md
+          git commit -m "Auto update wiki/Home.md"
+          git push
+      - name: Sync all files to wiki
         run: |
           cp -r $GITHUB_WORKSPACE/wiki/* $GITHUB_WORKSPACE/lichess-bot.wiki
           cd $GITHUB_WORKSPACE/lichess-bot.wiki
-          git config --global user.email "actions@github.com"
-          git config --global user.name "Wiki sync"
           git add .
-          git commit -m "Update wiki"
-          git push --force
+          git commit -m "Auto update wiki"
+          git push
--- a/.github/workflows/sync-wiki.yml
+++ b/.github/workflows/sync-wiki.yml
@@ -20,17 +20,33 @@ jobs:
         with:
           repository: "${{ github.repository }}.wiki"
           path: "lichess-bot.wiki"
+      - name: Set wiki path
+        uses: dorny/paths-filter@v2
+        id: wiki
+        with:
+          filters: |
+            src:
+              - 'wiki/**'
+      - name: Set README.md path
+        uses: dorny/paths-filter@v2
+        id: readme
+        with:
+          filters: |
+            src:
+              - 'README.md'
       - name: Set github bot global config
         run: |
           git config --global user.email "actions@github.com"
           git config --global user.name "actions-user"
       - name: Sync README.md to wiki/Home.md
+        if: steps.readme.outputs.src == 'true'
         run: |
           cp -r $GITHUB_WORKSPACE/README.md $GITHUB_WORKSPACE/wiki/Home.md
           git add wiki/Home.md
           git commit -m "Auto update wiki/Home.md"
           git push
       - name: Sync all files to wiki
+        if: steps.wiki.outputs.src == 'true'
         run: |
           cp -r $GITHUB_WORKSPACE/wiki/* $GITHUB_WORKSPACE/lichess-bot.wiki
           cd $GITHUB_WORKSPACE/lichess-bot.wiki
--- a/.github/workflows/sync-wiki.yml
+++ b/.github/workflows/sync-wiki.yml
@@ -42,7 +42,7 @@ jobs:
               - 'README.md'
       - name: Prevent Conflicts
         if: |
-          steps.readme.outputs.src == 'true' &&
+          steps.home.outputs.src == 'true' &&
           steps.readme.outputs.src == 'true'
         run: |
             echo "Error: Conflicting changes. Edit either the README.md file or the wiki/Home.md file, not both."
--- a/.github/workflows/sync-wiki.yml
+++ b/.github/workflows/sync-wiki.yml
@@ -1,0 +1,29 @@
+name: Sync wiki
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'wiki/**'
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+      - name: Checkout wiki
+        uses: actions/checkout@v3
+        with:
+          repository: "${{ github.repository }}.wiki"
+          path: "lichess-bot.wiki"
+      - name: Sync files to wiki
+        run: |
+          cp -r $GITHUB_WORKSPACE/wiki/* $GITHUB_WORKSPACE/lichess-bot.wiki
+          cd $GITHUB_WORKSPACE/lichess-bot.wiki
+          git config --global user.email "actions@github.com"
+          git config --global user.name "Wiki sync"
+          git add .
+          git commit -m "Update wiki"
+          git push --force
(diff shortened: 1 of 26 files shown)


TheYoBots:
With this change though you must be careful. Incase you edit the wiki page directly then it won't be synced in the github repo, so the next time a commit is made to the wiki dir it will erase the changes directly made in the wiki page.

AttackingOrDefending:
Very nice. Now users will get the wiki when they download lichess-bot and we can create PRs for wiki changes.

I think `README.md` should change to:
```md
## Steps
1. [Install lichess-bot](./wiki/How-to-Install.md)
2. [Create a lichess OAuth token](./wiki/How-to-create-a-Lichess-OAuth-token.md)
3. [Upgrade to a BOT account](./wiki/Upgrade-to-a-BOT-account.md)
4. [Setup the engine](./wiki/Setup-the-engine.md)
5. [Configure lichess-bot](./wiki/Configure-lichess-bot.md)
6. [Run lichess-bot](./wiki/How-to-Run-lichess%E2%80%90bot.md)

## Advanced options
- [Create a custom engine](./wiki/Create-a-custom-engine.md)

<br />

## Acknowledgements
Thanks to the Lichess team, especially T. Alexander Lystad and Thibault Duplessis for working with the LeelaChessZero team to get this API up. Thanks to the [Niklas Fiekas](https://github.com/niklasf) and his [python-chess](https://github.com/niklasf/python-chess) code which allows engine communication seamlessly.

## License
lichess-bot is licensed under the AGPLv3 (or any later version at your option). Check out the [LICENSE file](./LICENSE) for the full text.
```
so that it points to the files instead of the links.

<br />

Possible improvements (could be made in this or in another PR):
If we switch from markdown to asciidoc we can add if statements to see whether we should point to files or links. Example: [see here](https://github.com/AttackingOrDefending/lichess-bot/blob/t/wiki/How-to-Install.adoc).
I used:
```asciidoc
ifdef::env-github[]
**Next step**: https://github.com/lichess-bot-devs/lichess-bot/wiki/How-to-create-a-Lichess-OAuth-token[Create a Lichess OAuth token]
endif::[]
ifndef::env-github[]
**Next step**: link:./How-to-create-a-Lichess-OAuth-token.md[Create a Lichess OAuth token]
endif::[]
```
Another possible improvement would be to sync `wiki/Home.md` to `README.md` since they are the same file. This has to be done after switching to asciidoc if we change `README.md` to point to files.

MarkZH:
If the links are changed to the files in the repo, wouldn't clicking on them take readers out of the wiki space and then show code pages like this?

https://github.com/lichess-bot-devs/lichess-bot/blob/master/README.md

At the very least, we'd lose the table of contents on the right.

AttackingOrDefending:
No, because we check if we are on GitHub and change what we do accordingly. If you go to the above [link](https://github.com/AttackingOrDefending/lichess-bot/blob/t/wiki/How-to-Install.adoc) and click the first `Next step: Create a Lichess OAuth token` below mac and linux it will take you to the wiki. If you download the wiki and open it with PyCharm it will take you to the next file. So no one gets out of the wiki if they are in the wiki. If we sync the `README` and `wiki/Home`, then we will have to use the if else also in the `README`.

TheYoBots:
I'm not sure what changes are really wanted here, so I'll make a list:

- [ ] Change Markdown files to asciidoc (.md to .adoc)
About this I don't think it has anything related to my current pull request so I'm not sure if I should do it? It you think I should, then tell me and I will.

- [ ] Simplify README.md file to direct to docs in the wiki folder
Honestly I think that directing to the wiki page makes more sense since it looks much more neater than the readme files in the `wiki` dir and it displays all other wiki pages there too without showing other code files.

- [ ] sync `wiki/Home.md` to `README.md`
I personally think this is a good change considering that some changes might be made to the `README.md` and the `Home.md` wiki file is the same as the `README.md` file.

Put a tick on which ever changes you want me to implement. Owners of repositories have access to editing comments. Just click on the box and it will change (you don't need to do `edit comment`) then the next time I see it, I'll make the change.

AttackingOrDefending:
Just do number 3.

The main reason that I wanted to make `README.md` point to files is so that people that have lichess-bot downloaded can use it. We can do that after switching to asciidoc.

We can do number 1 in another PR.

.github/workflows/sync-wiki.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Checkout wiki
+        uses: actions/checkout@v3
+        with:
+          repository: "${{ github.repository }}.wiki"
+          path: "lichess-bot.wiki"
+      - name: Sync files to wiki
+        run: |
+          git config --global user.email "actions@github.com"
+          git config --global user.name "Wiki sync"
+          cp -r $GITHUB_WORKSPACE/README.md $GITHUB_WORKSPACE/wiki/Home.md
```
TheYoBots, line 25: Instead of `cp -r $GITHUB_WORKSPACE/README.md $GITHUB_WORKSPACE/wiki/Home.md` over here I tried [this](https://github.com/TheYoBots/lichess-bot/commit/c9e4c9db281b966e2e282bdc81dcad9b1e0d26cf). But for some reason this works locally but on github I get the following error: `fatal: bad revision 'HEAD~1'`. Although that would be a much better change.

AttackingOrDefending:
Change `git commit -m "Update wiki"` to `git commit -m "Auto update wiki"`.

`git config --global user.name "Wiki sync"` should change to something like `git config --global user.name "wiki-sync[bot]"` or `git config --global user.name "Wiki-Sync-BOT"`. The 1st option in more in line with `github-actions[bot]` while the 2nd is similar to `Version-BOT`.

I'm not sure about the user that should make this changes. I think there are 4 options:
- actions-user (currently used)
- github-actions[bot] (`41898282+github-actions[bot]@users.noreply.github.com`)
- Renaming `Version-BOT` to `lichess-bot[bot]` and having it do all automated lichess-bot tasks
- Creating a new account like `Version-BOT` but for wiki sync

.github/workflows/sync-wiki.yml
```diff
@@ -0,0 +1,37 @@
...
+          git add wiki/Home.md
+          git commit -m "Auto update wiki/Home.md"
+          git push
+      - name: Sync all files to wiki
+        run: |
+          cp -r $GITHUB_WORKSPACE/wiki/* $GITHUB_WORKSPACE/lichess-bot.wiki
+          cd $GITHUB_WORKSPACE/lichess-bot.wiki
+          git add .
```
AttackingOrDefending, line 35: Change this also to `git add wiki/*`.

TheYoBots:
`git config --global user.name "Wiki sync"` doesn't rename the bot to `wiki sync` to name will still remain `actions-user`. Maybe what I did was confusing. If you want I can change that line to `git config --global user.name "actions-user"`. But we can't rename a GitHub Bot.

Regarding which bot to use, I simply used actions-user because I had used it earlier and knew how it works. If you want I can change it to use the version-bot that's currently being used. Though I don't think it was necessary to create a separate bot just for this repository since there are already other bots that could do the same jobs.

TheYoBots:
Can the actions-user bot be used for even versioning? Because the version bot really can't be used to test any changes to the versioning github action since it uses private ssh key. actions-user is even an official github bot.

MarkZH:
This last change to the versioning workflow isn't correct. Changing from `&&` to `||` means that the only way the version bot won't run is if a commit message contains *both* `Auto update version` and `Auto update wiki/Home.md`. This change will cause an infinite loop of version updates.

AttackingOrDefending:
@TheYoBots Yes, action-user can be used for everything.

@MarkZH While the last change is wrong, it shouldn't cause an infinite loop because version bot only updates `versioning.yml` and `actions-user` only works if the `README.md` is changed. Also, since version bot is being run without the `GITHUB_TOKEN` it doesn't trigger any workflows. This should also be the case with actions-user.

The extra condition makes it so that version-bot won't be run on forks.

TheYoBots:
> This last change to the versioning workflow isn't correct. Changing from `&&` to `||` means that the only way the version bot won't run is if a commit message contains _both_ `Auto update version` and `Auto update wiki/Home.md`. This change will cause an infinite loop of version updates.

If I'm right the `&&` condition runs only when both are true, whereas the `||` condition will run when at least 1 of the conditions are true. That's why I put `||`.

TheYoBots:
This PR is ready. Tell me there are any other changes you want me to make here.

AttackingOrDefending:
What I like about version-bot is that it is obvious from the name what it does, while with action-user it isn't as obvious, so I prefer keeping version-bot for version updates. For syncing the wiki I'm not sure which bot we should use. @MarkZH Which bots do you think we should use for version updates and syncing the wiki? The options that I see are:
- `actions-user` only for wiki sync or for both tasks
- `github-actions[bot]` (`41898282+github-actions[bot]@users.noreply.github.com`) only for wiki sync or for both tasks
- Renaming `Version-BOT` to `lichess-bot[bot]` and having it do all automated lichess-bot tasks
- Creating a new account like `Version-BOT` but for wiki sync, and keeping `Version-BOT` for version updates

MarkZH:
I think using `actions-user` for both is fine. I've never logged into the Version-BOT account to do anything, and not needing to maintain such an account seems like an improvement. The commit message titles should make it obvious which process created the commit: `Auto update version`, `Auto update wiki`, etc.

TheYoBots:
Currently the commit messages are as follows:
- Version: "Auto update version to {new-version-date}"
- Wiki: "Auto update wiki"
- Updates to `Home.md` synced to `README.md`: "Auto update README.md"
- Updates to `README.md` synced to `Home.md`: "Auto update wiki/Home.md"

.github/workflows/sync-wiki.yml
```diff
@@ -0,0 +1,65 @@
...
+  push:
+    branches:
+      - master
+    paths:
+      - 'wiki/**'
+      - 'README.md'
+
+jobs:
+  sync:
+    if: github.repository != github.event.repository.full_name
```
AttackingOrDefending, line 13: Should be `if: github.repository == github.event.repository.full_name` not `!=`.

.github/workflows/versioning.yml
```diff
@@ -6,9 +6,9 @@ on:
 
 jobs:
   versioning:
-      if: "!contains(github.event.head_commit.message, 'Auto update version')"
-      runs-on: ubuntu-latest
-      steps:
+    if: github.repository != github.event.repository.full_name
```
AttackingOrDefending, line 9: Should be `if: github.repository == github.event.repository.full_name` not `!=`.

AttackingOrDefending:
Would it be possible to have the commit message be `"Auto update wiki/Home.md (INITIAL COMMIT MESSAGE HERE)"`? For example, if we update `README.md` with the commit message `Add new wiki page`, It would be nice if the next commit messages were `"Auto update wiki/Home.md (Add new wiki page)"` and then `"Auto update wiki (Add new wiki page)"`. It can probably be done with something like `"Auto update wiki/Home.md (${{ github.event.head_commit.message }})"`.

AttackingOrDefending (approved):

MarkZH (approved):

PR #819, opened 2023-08-25
Update python-test.yml

M-DinhHoangViet:


AttackingOrDefending:
There is no reason to add double quotes to `3.9`. Also, we only run tests on the minimum and the maximum supported version of python because the tests take a lot of resources and time.

PR #821, opened 2023-08-25
Add time to source field

TheYoBots:
It is a field used by many engines. Would be good if it is present like hashfull/tbhits for people who are interested to enable. https://python-chess.readthedocs.io/en/latest/engine.html#chess.engine.InfoDict

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -306,6 +306,14 @@ class EngineWrapper:
         wdl_percentage = round(wdl.relative.expectation() * 100, 1)
         return f"{wdl_percentage}%"
 
+    def readable_time(self, number: int) -> str:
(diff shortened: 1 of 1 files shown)


AttackingOrDefending (approved):

MarkZH:
I think you need to add `"Movetime"` to `stats = ["Source", "Evaluation", "Winrate", "Depth", "Nodes", "Speed", "Pv"]` on line 363.

TheYoBots:
> I think you need to add `"Movetime"` to `stats = ["Source", "Evaluation", "Winrate", "Depth", "Nodes", "Speed", "Pv"]` on line 363.

That's only if we want to print it when the eval commands is sent. I just added it like an option like how tbhits and hashfull is present.

MarkZH (approved):

PR #838, opened 2023-09-16
Math simplification

MarkZH:
Two changes that should not affect calculation results:
1. Use `math.copysign()` to simplify expressions that get the sign of a number--like `dtz` or `dtm`.
2. In `piecewise_function()` specify whether each range border is inclusive (`<=`) or exclusive (`<`). This way, exclusive borders can be specified exactly without needing to use off-by-one values.

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -20,7 +20,7 @@ import model
 import lichess
 from config import Configuration
 from timer import Timer, msec, seconds, msec_str, sec_str, to_seconds
-from typing import Any, Optional, Union
+from typing import Any, Optional, Union, Literal
 OPTIONS_TYPE = dict[str, Any]
 MOVE_INFO_TYPE = dict[str, Any]
 COMMANDS_TYPE = list[str]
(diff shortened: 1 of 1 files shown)


MarkZH:
Please check that the changed formulas do not change the results.

AttackingOrDefending:
There is a difference in the function `dtz_to_wdl`. In the original code, a value of 100 returns 1 while in the new version it returns 2. As you can see from [this position](https://syzygy-tables.info/?fen=4k3/2nr4/8/6R1/8/6N1/3B4/4K3_w_-_-_0_1), a dtz of 100 is still a draw. We can add a comment, saying that a dtz of 100 has a wdl of +/-1 and not +/-2. It just requires changing `(100, 'i', 2)` to `(100, 'e', 2)`. This was the only difference found after testing all values from -35000 to +35000.

MarkZH:
That was my mistake when changing the argument to the new format. I don't think a comment is necessary.

AttackingOrDefending:
A dtz of 100 means that after 100 plies have been played, the next move will zero the 50-move counter, so in total it would take 101 plies. This is why a dtz of 100 is a draw. I think mentioning this in a comment will be good because it isn't very intuitive. I don't know if adding it in a comment in `dtz_to_wdl` is the best idea though, because we use `piecewise_function` in a lot of places.

AttackingOrDefending (approved):

PR #843, opened 2023-09-24
Mypy fixes

MarkZH:
Better organized code to remove need for changes that are only to satisfy mypy.

### DIFF
--- a/engine_wrapper.py
+++ b/engine_wrapper.py
@@ -1098,7 +1098,7 @@ def get_syzygy(board: chess.Board, game: model.Game,
             # Attempt to only get the WDL score. It returns moves of quality="suggest", even if quality is set to "best".
             try:
                 moves = score_syzygy_moves(board, lambda tablebase, b: -tablebase.probe_wdl(b), tablebase)
-                best_wdl = int(max(moves.values()))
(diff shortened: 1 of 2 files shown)


engine_wrapper.py
```diff
@@ -1098,7 +1098,7 @@ def get_syzygy(board: chess.Board, game: model.Game,
             # Attempt to only get the WDL score. It returns moves of quality="suggest", even if quality is set to "best".
             try:
                 moves = score_syzygy_moves(board, lambda tablebase, b: -tablebase.probe_wdl(b), tablebase)
-                best_wdl = int(max(moves.values()))  # int is there only for mypy.
+                best_wdl = int(max(moves.values()))
```
AttackingOrDefending, line 1101: Why was this comment removed? I don't think anything changed in `get_syzygy`.

MarkZH:
I was looking for code that was only written to satisfy mypy by searching for comments that mentioned mypy. It seemed like code near those comments was more complicated than it needed to be. When I couldn't find anyway to remove the `int` cast or make `best_wdl = int(max(moves.values()))` or any of the surrounding code simpler, I deleted the comment. I suppose it does need explaining why the value is cast to an `int`.

AttackingOrDefending (approved):

PR #847, opened 2023-09-30
Only use local_game_done events to signal games are over

MarkZH:
Since `local_game_done` events are only put in the `control_queue` when the game thread ends, this change will prevent `gameFinish` events from prematurely removing games from the `active_games`, allowing for more simultaneous game threads than the user's configuration specifies (as mentioned [here](https://github.com/lichess-bot-devs/lichess-bot/pull/699#issuecomment-1660490533)). To make sure that game threads that exit due to errors do not leave behind zombie game IDs in `active_games` the error handler for game threads was changed to put `local_game_done` events in the queue. This change provided an opportunity to attempt saving the game PGN since a game-ending error would skip that step.

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -326,7 +326,7 @@ def lichess_bot_main(li: lichess.Lichess,
                 active_games.discard(event["game"]["id"])
                 matchmaker.game_done()
                 log_proc_count("Freed", active_games)
-                save_pgn_record(event, config, li, user_profile["username"])
+                save_pgn_record(event, config, user_profile["username"])
                 one_game_completed = True
             elif event["type"] == "challenge":
                 handle_challenge(event, li, challenge_queue, config.challenge, user_profile, matchmaker, recent_bot_challenges)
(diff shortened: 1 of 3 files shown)


lichess-bot.py
```diff
@@ -887,16 +898,25 @@ def pgn_game_record(li: lichess.Lichess, config: Configuration, game: model.Game
...
 
 
-def get_game_file_path(config: Configuration, game: model.Game, force_single: bool = False) -> str:
+def get_game_file_path(config: Configuration,
+                       game_id: str,
+                       white_name: str,
+                       black_name: str,
+                       user_name: str,
+                       game_is_over: bool,
+                       *, force_single: bool = False) -> str:
```
AttackingOrDefending, line 907: Why is `, *,` used?

MarkZH: To force the use of `force_single=True` rather than just `True` in the last parameter. I was looking at these lines:
```python
game_path = get_game_file_path(config, game_id, white_name, black_name, user_name, game_is_over)
single_game_path = get_game_file_path(config, game_id, white_name, black_name, user_name, game_is_over, force_single=True)
```
Without the star, get `single_game_path` line could look like this:
```python
single_game_path = get_game_file_path(config, game_id, white_name, black_name, user_name, game_is_over, True)
```
with no obvious clue as to what `True` means. I had considered doing the same thing for the `game_is_over` parameter, but the function isn't called with a literal True or False. I could move the `*` before `game_is_over` for consistency.

AttackingOrDefending: I don't think there is any reason to move it. The variable name `game_is_over` and `is_game_over(game)` shows what the parameter does.

AttackingOrDefending (approved):

PR #854, opened 2023-10-11
Add latest Python version to tests

MarkZH:


MarkZH:
I don't know why the test is failing. Running `pytest` and `lichess-bot` with python 3.12 works on my (Windows) computer. I'll investigate later (especially if I can set up a Linux environment).

MarkZH:
This needs to wait until other dependencies announce support for Python 3.12 (and for it to be available on more platforms).

PR #859, opened 2023-10-17
Check that response isn't None in HTTPError 

MarkZH:
Also, use lichess.is_final instead of duplicating it in lichess-bot.py


MarkZH:
The PR incorporates #858.

AttackingOrDefending (approved):

PR #865, opened 2023-11-04
Add the option to prefer high/low rated players

### LINKED ISSUE #840: Make matchmaking time delay shorter
Hi 
Can I suggest you guys to make the matchmaking_time_delay in config.yml to 0.1 minute (10 seconds) to make the bot challenge more opponent 
P/S: can you make matchmaking better by searching from the highest elo to lower elo opponents

AttackingOrDefending:
The bot can now prefer high rated players, low rated players, or have no preference at all.
EDIT: Will fix the tests later.

closes #840

### DIFF
--- a/matchmaking.py
+++ b/matchmaking.py
@@ -150,12 +150,12 @@ class Matchmaking:
                     game_type: str) -> list[int]:
         """Get the weight for each bot. A higher weights means the bot is more likely to get challenged."""
         if rating_preference == "high":
-            # A bot with max_rating rating will be twice as likely to get picked than a bt with min_rating rating.
+            # A bot with max_rating rating will be twice as likely to get picked than a bot with min_rating rating.
             reduce_ratings_by = min(min_rating - (max_rating - min_rating), min_rating - 1)
             # or, reduce_ratings_by = min(2 * min_rating - max_rating, min_rating - 1)
             weights = [bot.get("perfs", {}).get(game_type, {}).get("rating", 0) - reduce_ratings_by for bot in online_bots]
         elif rating_preference == "low":
-            # A bot with min_rating rating will be twice as likely to get picked than a bt with max_rating rating.
+            # A bot with min_rating rating will be twice as likely to get picked than a bot with max_rating rating.
             reduce_ratings_by = max(max_rating - (min_rating - max_rating), max_rating + 1)
             # or, reduce_ratings_by = max(2 * max_rating - min_rating, max_rating + 1)
             weights = [(reduce_ratings_by - bot.get("perfs", {}).get(game_type, {}).get("rating", 0))
--- a/matchmaking.py
+++ b/matchmaking.py
@@ -146,14 +146,16 @@ class Matchmaking:
             except Exception:
                 pass
 
-    def is_suitable_opponent(bot: USER_PROFILE_TYPE) -> bool:
-        perf = bot.get("perfs", {}).get(game_type, {})
-        return (bot["username"] != self.username()
-                and not self.in_block_list(bot["username"])
-                and not bot.get("disabled")
-                and (allow_tos_violation or not bot.get("tosViolation"))  # Terms of Service violation.
-                and perf.get("games", 0) > 0
-                and min_rating <= perf.get("rating", 0) <= max_rating)
+    def get_weights(online_bots: list[USER_PROFILE_TYPE], rating_preference: str, min_rating: int, max_rating: int
+                    ) -> list[int]:
+        weights = [1] * len(online_bots)
+        if rating_preference == "high":
+            reduce_ratings_by = min(min_rating - (max_rating - min_rating), min_rating - 1)
+            weights = [bot.get("perfs", {}).get(game_type, {}).get("rating", 0) - reduce_ratings_by for bot in online_bots]
+        elif rating_preference == "low":
+            reduce_ratings_by = min(min_rating - (max_rating - min_rating), min_rating - 1)
+            weights = [(max_rating - bot.get("perfs", {}).get(game_type, {}).get("rating", 0)) - reduce_ratings_by
+                       for bot in online_bots]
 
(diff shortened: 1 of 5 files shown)


matchmaking.py
```diff
@@ -146,6 +146,19 @@ def update_user_profile(self) -> None:
             except Exception:
                 pass
 
+    def get_weights(self, online_bots: list[USER_PROFILE_TYPE], rating_preference: str, min_rating: int, max_rating: int,
+                    game_type: str) -> list[int]:
+        """Get the weight for each bot. A higher weights means the bot is more likely to get challenged."""
+        weights = [1] * len(online_bots)
```
mheinzel, line 152: This is mostly about style, but I don't quite like declaring the default case upfront, where it is not clear yet that it's only a fallback. To me it read as if it was always used, but it's only used if neither "high" nor "low" apply. So why not like this:

```python
if rating_preference == "high":
    return [... for bot in online_bots]
elif rating_preference == "low":
    return [... for bot in online_bots]
else:
    return [1] * len(online_bots)
```

To me, this makes it clearer that there are three separate cases to consider.

matchmaking.py
```diff
@@ -146,6 +146,19 @@ def update_user_profile(self) -> None:
             except Exception:
                 pass
 
+    def get_weights(self, online_bots: list[USER_PROFILE_TYPE], rating_preference: str, min_rating: int, max_rating: int,
+                    game_type: str) -> list[int]:
+        """Get the weight for each bot. A higher weights means the bot is more likely to get challenged."""
+        weights = [1] * len(online_bots)
+        if rating_preference == "high":
+            reduce_ratings_by = min(min_rating - (max_rating - min_rating), min_rating - 1)
```
mheinzel, line 154: Where does this calculation come from? Isn't `min_rating - (max_rating - min_rating)` just `max_rating`? Then `min` only really makes sense here in the case where `min_rating > max_rating`, which seems like it should never happen.

Maybe I'm missing something, but then a clarifying comment would be helpful for future readers.

mheinzel (commented):
The feature seems useful to me, but I was wondering about some of the calculations.

Also, config options like this one (that have a few special values and everything else falls back to a default behavior) are always error-prone. I feel like it won't take long until someone has a typo in their config, e.g. "hihg", which just gets ignored silently.
Some similar fields have a check for this in `config.py`, see e.g. `valid_pgn_grouping_options`. So maybe it would be a good idea to do something similar for the rating preference.

I could also imagine this kind of check being done similarly to (or even as part of) `set_config_default`. Just as the `default` value, we could optionally list the valid values of each field, in a uniform declarative way. But that change might be better as a separate refactoring.

matchmaking.py
AttackingOrDefending: This calculation is the same as `2*min_rating-max_rating`. With this change, a bot with `max_rating` rating will be twice as likely to get picked as a bot with `min_rating`. I will add a comment.

AttackingOrDefending:
I will add checks for valid option values.

matchmaking.py
```diff
@@ -149,14 +149,15 @@ def update_user_profile(self) -> None:
...
                     game_type: str) -> list[int]:
         """Get the weight for each bot. A higher weights means the bot is more likely to get challenged."""
-        weights = [1] * len(online_bots)
         if rating_preference == "high":
             reduce_ratings_by = min(min_rating - (max_rating - min_rating), min_rating - 1)
             weights = [bot.get("perfs", {}).get(game_type, {}).get("rating", 0) - reduce_ratings_by for bot in online_bots]
         elif rating_preference == "low":
             reduce_ratings_by = min(min_rating - (max_rating - min_rating), min_rating - 1)
             weights = [(max_rating - bot.get("perfs", {}).get(game_type, {}).get("rating", 0)) - reduce_ratings_by
                        for bot in online_bots]
```
MarkZH: This formula isn't correct for `"low"`. It can result in negative weights. The correct formula and usage is
```python
reduce_ratings_by = max(max_rating - (min_rating - max_rating), max_rating + 1)
# or, reduce_ratings_by = max(2*max_rating - min_rating, max_rating + 1)
weights = [(reduce_ratings_by - bot.get("perfs", {}).get(game_type, {}).get("rating", 0))
           for bot in online_bots]
```
The variables `max_rating` and `min_rating` switch roles, `min()` becomes `max()`, and the order of subtraction is reversed.

matchmaking.py
```diff
@@ -146,6 +146,20 @@ def update_user_profile(self) -> None:
...
+    def get_weights(self, online_bots: list[USER_PROFILE_TYPE], rating_preference: str, min_rating: int, max_rating: int,
+                    game_type: str) -> list[int]:
+        """Get the weight for each bot. A higher weights means the bot is more likely to get challenged."""
+        if rating_preference == "high":
+            reduce_ratings_by = min(min_rating - (max_rating - min_rating), min_rating - 1)
+            weights = [bot.get("perfs", {}).get(game_type, {}).get("rating", 0) - reduce_ratings_by for bot in online_bots]
+        elif rating_preference == "low":
+            reduce_ratings_by = min(min_rating - (max_rating - min_rating), min_rating - 1)
+            weights = [(max_rating - bot.get("perfs", {}).get(game_type, {}).get("rating", 0)) - reduce_ratings_by
+                       for bot in online_bots]
```
MarkZH, line 158: This formula isn't correct for `"low"`. It can result in negative weights. The correct formula and usage is
```python
reduce_ratings_by = max(max_rating - (min_rating - max_rating), max_rating + 1)
# or, reduce_ratings_by = max(2*max_rating - min_rating, max_rating + 1)
weights = [(reduce_ratings_by - bot.get("perfs", {}).get(game_type, {}).get("rating", 0))
           for bot in online_bots]
```
The variables `max_rating` and `min_rating` switch roles, `min()` becomes `max()`, and the order of subtraction is reversed.

MarkZH (approved):

PR #902, opened 2024-02-03
Quit lichess-bot after all games finish

MarkZH:
## Type of pull request:
- [ ] Bug fix
- [x] Feature
- [ ] Other

## Description:

If a user wants to quit lichess-bot without interrupting in-progress games, they can add `quit_after_all_games_finish: true` to their configuration file. Then, pressing Ctrl-C will stop new game from being started--whether from matchmaking, challenges, or correspondence check-ins--and wait for all game processes to exit before quitting. The user no longer needs to forfeit games or wait around for current games to end in order to quit lichess-bot.

## Related Issues:

N/A

## Checklist:

- [x] I have read and followed the [contribution guidelines](/CONTRIBUTING.md).

N/A

### DIFF
--- a/config.yml.default
+++ b/config.yml.default
@@ -115,6 +115,7 @@ abort_time: 30                     # Time to abort a game in seconds when there 
 fake_think_time: false             # Artificially slow down the bot to pretend like it's thinking.
 rate_limiting_delay: 0             # Time (in ms) to delay after sending a move to prevent "Too Many Requests" errors.
 move_overhead: 2000                # Increase if your bot flags games too often.
+quit_after_all_games_finish: false # If set to true, then pressing Ctrl-C to quit will only stop lichess-bot after all current games have finished.
 
 correspondence:
   move_time: 60                    # Time in seconds to search in correspondence games.
--- a/lib/config.py
+++ b/lib/config.py
@@ -153,6 +153,7 @@ def insert_default_values(CONFIG: CONFIG_DICT_TYPE) -> None:
     """
     set_config_default(CONFIG, key="abort_time", default=20)
     set_config_default(CONFIG, key="move_overhead", default=1000)
+    set_config_default(CONFIG, key="quit_after_all_games_finish", default=False)
     set_config_default(CONFIG, key="rate_limiting_delay", default=0)
     set_config_default(CONFIG, key="pgn_file_grouping", default="game", force_empty_values=True)
     set_config_default(CONFIG, "engine", key="working_dir", default=os.getcwd(), force_empty_values=True)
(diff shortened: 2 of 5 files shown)


MarkZH:
Annoying hack warning: In [`logging_listener_proc()`](https://github.com/lichess-bot-devs/lichess-bot/pull/902/files#diff-e326008cac463869bad7488e0e9b058e652abe5c59e980c0476f05bb86b10508R183), the `queue.get()` call had to be changed to the non-blocking version because the SIGTERM signal seemed to do irreparable damage to the state of the queue if received during a blocking `get()`. In my first attempts at this PR, a SIGTERM would cause every other call to `get()` to return `None`. These weren't extra queue items. Half of the log lines would be lost. It is only by using `get(block=False)` that this problem is avoided. The [`time.sleep(0.1)`](https://github.com/lichess-bot-devs/lichess-bot/pull/902/files#diff-e326008cac463869bad7488e0e9b058e652abe5c59e980c0476f05bb86b10508R185) is necessary to prevent constant calls to `queue.get(block=False)` from overheating a CPU.

This is probably the same issue that required the new `if event is None:` check in [`next_event()`](https://github.com/lichess-bot-devs/lichess-bot/pull/902/files#diff-e326008cac463869bad7488e0e9b058e652abe5c59e980c0476f05bb86b10508R380). I put in a simple check instead of the full fix like in `logging_listener_proc()` because the first SIGTERM signal causes lichess-bot to stop listening for messages from lichess.org, so a more complete fix didn't seem necessary. I could be wrong.

AttackingOrDefending:
Nice idea.

While testing this PR my engine closed and was restarted in order to finish the game. Note: Even though `DEBUG Received SIGINT. Terminating client.` is seen twice in the logs I only clicked it once, and so I didn't receive any quitting now messages (which is the expected behavior).

Also, probably due to the delay in `logging_listener_proc` the logs aren't completed both in the console and the log file.

b

Log file:
[recent.log](https://github.com/lichess-bot-devs/lichess-bot/files/14151243/recent.log)

I think that it would be better if the bot aborted all abortable games instead of finishing them before closing.

AttackingOrDefending:
The logs are now completed, but the engine is still restarted when I press Ctrl+C the first time. You can see it in the logs below.

[old.log](https://github.com/lichess-bot-devs/lichess-bot/files/14157607/old.log)
[recent.log](https://github.com/lichess-bot-devs/lichess-bot/files/14157608/recent.log)
[recent2.log](https://github.com/lichess-bot-devs/lichess-bot/files/14157621/recent2.log)

For example, in `old.log`, the pid of the engine before pressing Ctrl+C is 13972, while after it is 5164 because python-chess started a new engine. In `recent2.log`, the engine even attempted to restart twice before lichess-bot closed.

MarkZH:
That should do it.

AttackingOrDefending (approved):

PR #905, opened 2024-02-03
Pytest

AttackingOrDefending:
## Type of pull request:
- [ ] Bug fix
- [ ] Feature
- [x] Other

## Description:

Doesn't copy `test_bot\lichess.py` to `lichess.py` and the back. In the past, when the tests timed out the wrong `lichess.py` file would be in `lib/` so that the user would have to copy it back manually.

## Related Issues:

None

## Checklist:

- [x] I have read and followed the [contribution guidelines](/CONTRIBUTING.md).

### DIFF
--- a/lichess-bot.py
+++ b/lichess-bot.py
@@ -3,7 +3,7 @@ import argparse
 import chess
 import chess.pgn
 from chess.variant import find_variant
-from lib import engine_wrapper, model, lichess, matchmaking
+from lib import engine_wrapper, model, matchmaking
 import json
 import logging
 import logging.handlers
@@ -30,6 +30,10 @@ from http.client import RemoteDisconnected
 from queue import Queue
 from multiprocessing.pool import Pool
 from typing import Any, Optional, Union
+if __name__ == "main":
+    from lib import lichess
+else:
+    from test_bot import lichess
 USER_PROFILE_TYPE = dict[str, Any]
 EVENT_TYPE = dict[str, Any]
 PLAY_GAME_ARGS_TYPE = dict[str, Any]
--- a/test_bot/test_bot.py
+++ b/test_bot/test_bot.py
@@ -1,4 +1,6 @@
 """Test lichess-bot."""
+import logging
+
 import pytest
 import zipfile
 import requests
@@ -18,8 +20,6 @@ from lib.timer import Timer, to_seconds, seconds
 from typing import Any
 if __name__ == "__main__":
     sys.exit(f"The script {os.path.basename(__file__)} should only be run by pytest.")
-shutil.copyfile("lib/lichess.py", "lib/correct_lichess.py")
-shutil.copyfile("test_bot/lichess.py", "lib/lichess.py")
 lichess_bot = importlib.import_module("lichess-bot")
 
 platform = sys.platform
@@ -185,7 +185,7 @@ def run_bot(raw_config: dict[str, Any], logging_level: int, opponent_path: str =
     config.insert_default_values(raw_config)
     CONFIG = config.Configuration(raw_config)
     lichess_bot.logger.info(lichess_bot.intro())
-    li = lichess_bot.lichess.Lichess(CONFIG.token, CONFIG.url, lichess_bot.__version__)
+    li = lichess_bot.lichess.Lichess(CONFIG.token, CONFIG.url, lichess_bot.__version__, logging_level, 1)
 
     user_profile = li.get_profile()
     username = user_profile["username"]
--- a/test_bot/test_bot.py
+++ b/test_bot/test_bot.py
@@ -1,6 +1,4 @@
 """Test lichess-bot."""
-import logging
-
 import pytest
 import zipfile
 import requests
--- a/test_bot/conftest.py
+++ b/test_bot/conftest.py
@@ -6,8 +6,6 @@ from typing import Any
 
 def pytest_sessionfinish(session: Any, exitstatus: Any) -> None:
     """Remove files created when testing lichess-bot."""
-    shutil.copyfile("lib/correct_lichess.py", "lib/lichess.py")
-    os.remove("lib/correct_lichess.py")
     if os.path.exists("TEMP") and not os.getenv("GITHUB_ACTIONS"):
         shutil.rmtree("TEMP")
     if os.path.exists("logs"):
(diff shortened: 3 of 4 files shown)


AttackingOrDefending:
I've also tried:
```python
def get_lichess():
    if __name__ == "main":
        from lib import lichess as lichess1
        return lichess1
    else:  # __name__ == "lichess-bot"
        from test_bot import lichess as lichess2
        return lichess2
```
but it raises even more errors.

MarkZH:
Since there's only ever one instance of the `Lichess` class, perhaps a better approach is to create the `li` instance with a runtime-specified type, rather than importing the correct lichess:
```python
    # lichess-bot.py

    # ...
    import test_bot
    # ...
    LICHESS_TYPE = Union[lichess.Lichess, test_bot.lichess.Lichess]

    # ...

    lichess_communicator: type[LICHESS_TYPE]
    if __name__ == "__main__":
        lichess_communicator = lichess.Lichess
    else:
        lichess_communicator = test_bot.lichess.Lichess

    li = lichess_communicator(CONFIG.token, CONFIG.url, __version__, logging_level, max_retries)
```
This will require fixing the arguments and typing of the `test_bot` version of the `Lichess` class to match the original and finding a way to pass the `LICHESS_TYPE` definition around to other modules. But, I think this will be easier.

MarkZH:
I do like this change. This will especially help when I'm changing lichess.py and no longer have to worry about losing changes during testing.

lib/engine_wrapper.py
```diff
@@ -593,7 +595,12 @@ def getHomemadeEngine(name: str) -> type[MinimalEngine]:
     :return: The engine with this name.
     """
     from lib import strategies
-    engine: type[MinimalEngine] = getattr(strategies, name)
+    from test_bot import strategies as test_strategies
+    engine: type[MinimalEngine]
+    if __name__ == "main":
+        engine = getattr(strategies, name)
+    else:
+        engine = getattr(test_strategies, name)
```
MarkZH, line 603: This doesn't work since `__name__` for engine_wrapper.py is never `"__main__"` (or `"main"`). This breaks using homemade engines in normal use since `test_strategies` would always be used.

lichess-bot.py
```diff
@@ -1021,7 +1025,7 @@ def start_lichess_bot() -> None:
 
     max_retries = CONFIG.engine.online_moves.max_retries
     check_python_version()
-    li = lichess.Lichess(CONFIG.token, CONFIG.url, __version__, logging_level, max_retries)
+    li = lichess_communicator(CONFIG.token, CONFIG.url, __version__, logging_level, max_retries)
```
MarkZH, line 1028: With the way this PR has been written, defining `lichess_communicator` isn't needed anymore. `lichess.Lichess` can be used directly here.

test_bot/test_bot.py
```diff
@@ -185,7 +183,7 @@ def run_bot(raw_config: dict[str, Any], logging_level: int, opponent_path: str =
     config.insert_default_values(raw_config)
     CONFIG = config.Configuration(raw_config)
     lichess_bot.logger.info(lichess_bot.intro())
-    li = lichess_bot.lichess.Lichess(CONFIG.token, CONFIG.url, lichess_bot.__version__)
+    li = lichess_bot.test_bot.lichess.Lichess(CONFIG.token, CONFIG.url, lichess_bot.__version__, logging_level, 1)
```
MarkZH, line 186: Is there a reason for not using `import testbot.lichess` and `li = test_bot.lichess.Lichess()`?

lichess-bot.py
```diff
@@ -3,7 +3,7 @@
 import chess
 import chess.pgn
 from chess.variant import find_variant
-from lib import engine_wrapper, model, lichess, matchmaking
+from lib import engine_wrapper, model, matchmaking, lichess
```
MarkZH, line 6: This change doesn't seem necessary.

MarkZH:
Here's a hacky idea that seems to work. Change the name of the homemade class to something that wouldn't be a legal class name as a signal that this is for testing. In this case, by using hyphens in the class name. That way, users won't be able to accidentally reach the test branch.
```diff
diff --git a/lib/engine_wrapper.py b/lib/engine_wrapper.py
index f64a6ec..4fe668b 100644
--- a/lib/engine_wrapper.py
+++ b/lib/engine_wrapper.py
@@ -12,7 +12,6 @@ import datetime
 import time
 import random
 import math
-import sys
 import test_bot.lichess
 from collections import Counter
 from collections.abc import Callable
@@ -588,6 +587,9 @@ class FillerEngine:
         return method
 
 
+test_suffix = "-for-lichess-bot-testing-only"
+
+
 def getHomemadeEngine(name: str) -> type[MinimalEngine]:
     """
     Get the homemade engine with name `name`. e.g. If `name` is `RandomMove` then we will return `strategies.RandomMove`.
@@ -598,8 +600,8 @@ def getHomemadeEngine(name: str) -> type[MinimalEngine]:
     from lib import strategies
     from test_bot import strategies as test_strategies
     engine: type[MinimalEngine]
-    if "pytest" in sys.modules:
-        engine = getattr(test_strategies, name)
+    if name.endswith(test_suffix):
+        engine = getattr(test_strategies, name.removesuffix(test_suffix))
     else:
         engine = getattr(strategies, name)
     return engine
diff --git a/test_bot/test_bot.py b/test_bot/test_bot.py
index 1439c54..09828a8 100644
--- a/test_bot/test_bot.py
+++ b/test_bot/test_bot.py
@@ -16,6 +16,7 @@ import tarfile
 import test_bot.lichess
 from lib import config
 from lib.timer import Timer, to_seconds, seconds
+from lib.engine_wrapper import test_suffix
 from typing import Any
 if "pytest" not in sys.modules:
     sys.exit(f"The script {os.path.basename(__file__)} should only be run by pytest.")
@@ -292,7 +293,7 @@ def test_homemade() -> None:
     with open("./config.yml.default") as file:
         CONFIG = yaml.safe_load(file)
     CONFIG["token"] = ""
-    CONFIG["engine"]["name"] = "Stockfish"
+    CONFIG["engine"]["name"] = "Stockfish" + test_suffix
     CONFIG["engine"]["protocol"] = "homemade"
     CONFIG["pgn_directory"] = "TEMP/homemade_game_record"
     win = run_bot(CONFIG, logging_level)
```

MarkZH (approved):

PR #924, opened 2024-03-09
More complex challenge filtering and game-dependent option management

### LINKED ISSUE #922: In FromPosition challenges: override selected UCI settings depending on the FEN
I'm the Lc0 dev behind the Leela Piece Odds bots on lichess https://lichess.org/player/bots, each accepting challenges for preselected starting positions from the appropriate side. I encountered the problem that some settings (especially Contempt, potentially others as well) are supposed to depend on the starting position and side to play for optimal performance, as e.g. playing knight odds as white and as black is quite different. However, the way lichess-bot works is that there is a static config.yml used once when starting up the bot, and challenges are received only later.

I already added a functionality to the "accept challenges" logic of `is_supported` in `model.py` which checks a stored file (plain text `fenlist.txt`, changing it to `fenlist.json` for the per-position options) whether the combination of position + color is accepted. I'd like to be able to use some syntax like [fenlist.json](https://github.com/lichess-bot-devs/lichess-bot/files/14495575/fenlist.json) to specify UCI options to override from the values set in `config.yml` when accepting a challenge.

To some extent, having multiple bots (one for each level) already was an attempt to work around the issue, since it obviously allows one config per bot. There are other advantages so I'm not particularly unhappy with it, but there are some significant downsides to it, for example the handling of parallel challenges during times with higher demand. However, as mentioned above it turns out that we also want to use different settings for playing with white vs black, and splitting this up into even more bots doesn't make sense.

**Additional context**
My code so far is still based on the pre-file-reorganization structure but obviously trivial to rebase, so if there is interest on your side to include specific handling of `FromPosition` challenges to check for a list of allowed starting FENs + color into lichess-bot, I'd be happy to contribute. Same goes with the per challenge override of settings, once I got it done -- but I believe I need some assistance in figuring out where exactly to connect the logic "found the position in the `fenlist.json`, this is the list of options to override" to actually overriding the options in the engine setup.

Thanks a lot for your help in advance! I really appreciate how easy it is in general to work with the lichess-bot API and provide interesting features directly to lichess users :)

MarkZH:
## Type of pull request:
- [ ] Bug fix
- [x] Feature
- [ ] Other

## Description:

This creates a way for user's to have more fine-grained control over which challenges are accepted and what bot settings are used on a per-game basis. If the config.yml file does not suit a bot runner's purposes, they can create a python file (extra_game_handlers.py) to specify behavior with python code.

## Related Issues:

Implements ideas from and closes #922.

## Checklist:

N/A

### DIFF
--- a/lib/model.py
+++ b/lib/model.py
@@ -29,7 +29,8 @@ class Challenge:
         self.challenge_target = Player(challenge_info.get("destUser") or {})
         self.from_self = self.challenger.name == user_profile["username"]
         self.initial_fen = challenge_info.get("initialFen", "startpos")
-        self.color = challenge_info["color"]
+        color = challenge_info["color"]
+        self.color = color if color != "random" else challenge_info["finalColor"]
         self.time_control = challenge_info["timeControl"]
 
     def is_supported_variant(self, challenge_cfg: Configuration) -> bool:
--- a/lib/model.py
+++ b/lib/model.py
@@ -28,6 +28,9 @@ class Challenge:
         self.challenger = Player(challenge_info.get("challenger") or {})
         self.opponent = Player(challenge_info.get("destUser") or {})
         self.from_self = self.challenger.name == user_profile["username"]
+        self.initial_fen = challenge_info["initialFen"]
+        self.color = challenge_info["color"]
+        self.time_control = challenge_info["timeControl"]
 
     def is_supported_variant(self, challenge_cfg: Configuration) -> bool:
         """Check whether the variant is supported."""
@@ -102,7 +105,8 @@ class Challenge:
                               or self.decline_due_to(self.is_supported_mode(config), "casual" if self.rated else "rated")
                               or self.decline_due_to(self.challenger.name not in config.block_list, "generic")
                               or self.decline_due_to(self.challenger.name in allowed_opponents, "generic")
-                              or self.decline_due_to(self.is_supported_recent(config, recent_bot_challenges), "later"))
+                              or self.decline_due_to(self.is_supported_recent(config, recent_bot_challenges), "later")
+                              or self.decline_due_to(is_supported_extra(self), "generic"))
 
             return not decline_reason, decline_reason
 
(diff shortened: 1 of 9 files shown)


MarkZH:
I put the extra_game_handlers.py file in .gitignore to prevent future users from trying to push their files into their PRs.

lib/model.py
```diff
@@ -131,6 +135,14 @@ def __repr__(self) -> str:
         return self.__str__()
 
 
+try:
+    from extra_game_handlers import is_supported_extra  # type: ignore
+except ImportError:
+    def is_supported_extra(challenge: Challenge) -> bool:
+        """Accept every challnege by default."""
```
AttackingOrDefending, line 142: There is a typo here. Also, I would prefer a comment more similar to `game_specific_options` like `"""Return True by default not to add any additional restrictions to challenges."""`, because someone may think that their bot will accept every challenge.

lib/model.py
```diff
@@ -28,6 +28,9 @@ def __init__(self, challenge_info: dict[str, Any], user_profile: dict[str, Any])
         self.challenger = Player(challenge_info.get("challenger") or {})
         self.opponent = Player(challenge_info.get("destUser") or {})
         self.from_self = self.challenger.name == user_profile["username"]
+        self.initial_fen = challenge_info["initialFen"]
```
AttackingOrDefending, line 31: There is no `initialFen` in standard games.

AttackingOrDefending:
I think we should already have the file `extra_game_handlers.py` and not expect the user to create it. It will probably contain:
```python
from lib import model
from typing import Any


def game_specific_options(game: model.Game) -> dict[str, Any]:
    """Return an empty dict by default to keep options unchanged."""
    return {}


def is_supported_extra(challenge: model.Challenge) -> bool:
    """Accept every challenge by default."""
    return True
```
This way we wouldn't need the try except block. On the other hand, we would need to move the import in `model.py` inside the function to avoid a circular import, but I still think it's worth it.

AttackingOrDefending (approved):

PR #933, opened 2024-03-18
Type some Lichess API methods as returning None

greg-finley:
## Type of pull request:
- [ ] Bug fix
- [ ] Feature
- [x] Other

## Description:

For endpoints that the [Lichess docs](https://lichess.org/api#tag/Bot/operation/botGameAbort) say just return `{"ok": boolean}`, type it as such.

We could keep extending similar to what I was doing in a different branch: https://github.com/lichess-bot-devs/lichess-bot/compare/master...greg-finley:lichess-bot:types

I'd suggest each TODO is a separate PR probably. Maybe it's even worth making separate issues for all of them (I can do it if you want).

## Related Issues:

Makes progress on https://github.com/lichess-bot-devs/lichess-bot/issues/921

## Checklist:

### DIFF
--- a/lib/lichess.py
+++ b/lib/lichess.py
@@ -357,7 +357,6 @@ class Lichess:
         """Cancel a challenge."""
         self.api_post("cancel", challenge_id, raise_for_status=False)
 
-    # TOOD: Replace with a more specific type
     def online_book_get(self, path: str, params: Optional[dict[str, Any]] = None, stream: bool = False) -> JSON_REPLY_TYPE:
         """Get an external move from online sources (chessdb or lichess.org)."""
         @backoff.on_exception(backoff.constant,
--- a/lib/lichess.py
+++ b/lib/lichess.py
@@ -41,6 +41,7 @@ logger = logging.getLogger(__name__)
 
 MAX_CHAT_MESSAGE_LEN = 140  # The maximum characters in a chat message.
 
+
 class OKResponse(TypedDict):
     """Often given by the API on POSTs or endpoints needing no further action."""
     ok: bool
@@ -268,8 +269,10 @@ class Lichess:
         :param game_id: The id of the game.
         :param move: The move to make.
         """
-        return cast(OKResponse, self.api_post("move", game_id, move.move,
-                             params={"offeringDraw": str(move.draw_offered).lower()}))
+        return cast(OKResponse, self.api_post(
+            "move", game_id, move.move,
+            params={"offeringDraw": str(move.draw_offered).lower()}
+        ))
 
     def chat(self, game_id: str, room: str, text: str) -> OKResponse:
         """
@@ -309,11 +312,12 @@ class Lichess:
     def decline_challenge(self, challenge_id: str, reason: str = "generic") -> OKResponse:
         """Decline a challenge."""
         try:
-            return cast(OKResponse, self.api_post("decline", challenge_id,
-                                 data=f"reason={reason}",
-                                 headers={"Content-Type":
-                                          "application/x-www-form-urlencoded"},
-                                 raise_for_status=False))
+            return cast(OKResponse, self.api_post(
+                "decline", challenge_id,
+                data=f"reason={reason}",
+                headers={"Content-Type": "application/x-www-form-urlencoded"},
+                raise_for_status=False
+            ))
         except Exception:
             return {"ok": False}
 
--- a/lib/lichess.py
+++ b/lib/lichess.py
@@ -10,7 +10,7 @@ import traceback
 from collections import defaultdict
 import datetime
 from lib.timer import Timer, seconds, sec_str
-from typing import Optional, TypedDict, Union, Any, cast
+from typing import Optional, Union, Any
 import chess.engine
 JSON_REPLY_TYPE = dict[str, Any]
 REQUESTS_PAYLOAD_TYPE = dict[str, Any]
@@ -40,12 +40,6 @@ ENDPOINTS = {
 logger = logging.getLogger(__name__)
 
 MAX_CHAT_MESSAGE_LEN = 140  # The maximum characters in a chat message.
-
-
-class OKResponse(TypedDict):
-    """Often given by the API on POSTs or endpoints needing no further action."""
-
-    ok: bool
 
 
 class RateLimited(RuntimeError):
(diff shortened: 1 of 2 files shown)


lib/lichess.py
```diff
@@ -279,44 +283,48 @@ def chat(self, game_id: str, room: str, text: str) -> JSON_REPLY_TYPE:
...
-        return self.api_post("chat", game_id, data=payload)
+        return cast(OKResponse, self.api_post("chat", game_id, data=payload))
 
-    def abort(self, game_id: str) -> JSON_REPLY_TYPE:
+    def abort(self, game_id: str) -> OKResponse:
         """Aborts a game."""
-        return self.api_post("abort", game_id)
+        return cast(OKResponse, self.api_post("abort", game_id))
 
+    # TODO: Align types with test_bot/lichess.py
```
greg-finley, line 295: I'm pondering suggesting to make the `Lichess` class in `test_bot/lichess.py` subclass the one in `lib/lichess.py`, or they both subclass an abstract base class. But this type mismatch will be sorta a blocker to that idea.

MarkZH: What benefit would there be to making subclasses? The two versions of the `get_event_stream()` method are only used here:
https://github.com/lichess-bot-devs/lichess-bot/blob/b3c9239f0a3bf40902c15cb3410279380c9d1d4c/lichess-bot.py#L96-L97
The type of `response` here is `EventStream | requests.models.Response` due to the two different classes returning different types for this method. However, since both return types have an `iter_lines()` method, the type checking passes. I don't see a need to align the return types of the `get_event_stream()` methods of `lib/Lichess` and `test_bot/Lichess`.

greg-finley: OK sounds good, I will drop the TODO comment.

I have seen other mocking strategies in the past where they extended the actual implementation, but seems not necessary here. 👍

MarkZH:
I wonder if `-> None` would be a better annotation for all of these methods. Because both `api_get()` and `api_post()` call `response.raise_for_status()`, the only value of `OKResponse` that is ever returned is `{"ok": True}`. Otherwise, an exception is thrown. And then, `backoff` handles most of the exceptions. The places where `response.raise_for_status()` is not raised--`decline_challenge()`, `challenge()`, and `cancel()`--are places where the failure of the request doesn't matter either because the request will be repeated later or because the failure doesn't matter (declining a challenge and never accepting a challenge have the same effect).

An example of using `None` as the return type is illustrated below:
```python
    def make_move(self, game_id: str, move: chess.engine.PlayResult) -> None:
        """
        Make a move.
        :param game_id: The id of the game.
        :param move: The move to make.
        """
        self.api_post("move", game_id, move.move,
                      params={"offeringDraw": str(move.draw_offered).lower()})
```

MarkZH:
@AttackingOrDefending Any thoughts?

greg-finley:
> I wonder if `-> None` would be a better annotation for all of these methods. Because both `api_get()` and `api_post()` call `response.raise_for_status()`, the only value of `OKResponse` that is ever returned is `{"ok": True}`. Otherwise, an exception is thrown. And then, `backoff` handles most of the exceptions. The places where `response.raise_for_status()` is not raised--`decline_challenge()`, `challenge()`, and `cancel()`--are places where the failure of the request doesn't matter either because the request will be repeated later or because the failure doesn't matter (declining a challenge and never accepting a challenge have the same effect).
> 
> An example of using `None` as the return type is illustrated below:
> 
> ```python
>     def make_move(self, game_id: str, move: chess.engine.PlayResult) -> None:
>         """
>         Make a move.
>         :param game_id: The id of the game.
>         :param move: The move to make.
>         """
>         self.api_post("move", game_id, move.move,
>                       params={"offeringDraw": str(move.draw_offered).lower()})
> ```

Yep, that makes sense! Will do.

AttackingOrDefending:
I like the change to `None` since we don't use the return value. We could also log the return value but I don't think that's needed.

lib/lichess.py
```diff
@@ -253,21 +253,23 @@ def rate_limit_time_left(self, path_template: str) -> datetime.timedelta:
...
 
         :param game_id: The id of the game.
         :param move: The move to make.
         """
-        return self.api_post("move", game_id, move.move,
-                             params={"offeringDraw": str(move.draw_offered).lower()})
+        self.api_post(
+            "move", game_id, move.move,
+            params={"offeringDraw": str(move.draw_offered).lower()}
+        )
```
MarkZH, line 270: To keep the style consistent: (1) put the first line of arguments on the same line as the function name, (2) put the final closing parenthesis on the same line as the last arguments, and (3) line up the arguments with the opening parenthesis of the function.
```python
        self.api_post("move", game_id, move.move,
                      params={"offeringDraw": str(move.draw_offered).lower()})
```

lib/lichess.py
```diff
@@ -296,27 +297,30 @@ def get_game_stream(self, game_id: str) -> requests.models.Response:
...
-                                 data=f"reason={reason}",
-                                 headers={"Content-Type":
-                                          "application/x-www-form-urlencoded"},
-                                 raise_for_status=False)
+            self.api_post(
+                "decline", challenge_id,
+                data=f"reason={reason}",
+                headers={"Content-Type": "application/x-www-form-urlencoded"},
+                raise_for_status=False
+            )
```
MarkZH, line 312: Same comment as previous.

lib/lichess.py
greg-finley: Sounds good.

Btw any interest in black or ruff for this repo? They can handle this sort of formatting stuff automatically.

MarkZH: We've considered it, but we prefer the flexibility to arrange code that makes sense in context and using flake8 to keep things tidy. I just tried running black on this repository and there's maybe 1 in 10 changes I agree with. Personally, I don't like how black puts the closing parenthesis/brace on a new line in a multi-line list, nor putting a comma after the last element of a list.

lib/lichess.py
```diff
@@ -296,27 +295,28 @@ def get_game_stream(self, game_id: str) -> requests.models.Response:
...
-                                 raise_for_status=False)
+            self.api_post("decline", challenge_id,
+                          data=f"reason={reason}",
+                          headers={"Content-Type": "application/x-www-form-urlencoded"},
+                          raise_for_status=False)
         except Exception:
-            return {}
+            pass
 
+    # TODO: Replace with a more specific type
```
MarkZH, line 312: Last thing: delete these TODO comments.

MarkZH (approved):

lib/lichess.py
```diff
@@ -355,10 +353,11 @@ def challenge(self, username: str, payload: REQUESTS_PAYLOAD_TYPE) -> JSON_REPLY
         """Create a challenge."""
         return self.api_post("challenge", username, payload=payload, raise_for_status=False)
 
-    def cancel(self, challenge_id: str) -> JSON_REPLY_TYPE:
+    def cancel(self, challenge_id: str) -> None:
         """Cancel a challenge."""
-        return self.api_post("cancel", challenge_id, raise_for_status=False)
+        self.api_post("cancel", challenge_id, raise_for_status=False)
 
+    # TOOD: Replace with a more specific type
```
AttackingOrDefending, line 360: This comment hasn't been removed.

greg-finley: Whoops, sorry, was just doing a search and missed this

AttackingOrDefending (approved):

Issue #780, opened 2023-07-12
Create and maintain official `lichess-bot` Docker image


eduherminio:
## Problem

I would like to use `lichess-bot` with a custom engine in Windows, Mac and Linux, either AMD or ARM, without having to worry about installing/updating its (python) dependencies.

## Proposed solution

A potential solution to that is to create a Docker image that has all `lichess-bot` dependencies pre-installed.

Ideally I envision that Docker image having this repository checked out there (that is, generating a new image per `lichess-bot` release).

Then it would be a matter of attaching a volume(s) to the container in order to share the engine binaries with it and customize the `config.yml` file to point at it, use the right token and configure the desired bot behavior.

Ideally we would like to achieve an implementation (+ sample scripts/commands) that doesn't require us to enter the Docker container for anything.

I'd be more than happy to get involved into the implementation and maintenance of this solution.

## PoCs

This week I kinda already implemented something like this in [lichess-bot-playground](https://github.com/lynx-chess/lichess-bot-playground), but not in an optimal way (the image, while functional, should be as small as possible and I generated it on top of the not exactly slim `python` image). It also requires entering the container to manually modify the `config.yml` file, so this is still far from a perfect solution.

Back in the day I also implemented something similar in [Lynx_BOT](https://github.com/lynx-chess/Lynx_BOT), but embedding the engine itself in the image.
This is a more advanced solution in terms of not having to manually enter the container, since the problem of providing the desired configuration is solved there with a `docker-compose` file that copies the desired `config.yml` file into the container.

## Summary

My idea would be to create a `lichess-bot` image and provide either a sample docker command (similar to the one in the README of [lichess-bot-playground](https://github.com/lynx-chess/lichess-bot-playground) or a sample docker compose file (see [`docker-compose.yml` file in Lynx_BOT](https://github.com/lynx-chess/Lynx_BOT/blob/main/docker-compose.yml)) that achieve that second purpose of being able to provide the engine binaries and the configuration files without having to manually enter the container generated from the docker image.

This image could be use as a base image for engine images, in a similar fashion as I did with Lynx_BOT: adding the engine binaries on top of it. Note that the custom `config.yml` could also be added to engine images, but since that configuration is more likely to be tweaked often, I'd still recommend to inject it separately and keep engine image versions as wrappers of the equivalent (same number) engine versions that have a certain `lichess-bot` version side to side with it, setup and ready to be used to connect that engine to lichess.

[...]

PR #936, opened 2024-04-01
Docker image

fathzer:
## Type of pull request:
- [ ] Bug fix
- [ x] Feature
- [ ] Other

## Description:
Adds a workflow that creates a minimal, easy to use, Docker image.
It also contains some documentation about the image.

## Related Issues:
This PR is related to #780 issue.

## Checklist:

- [ x] I have read and followed the [contribution guidelines](/docs/CONTRIBUTING.md).
- [ x] I have added necessary documentation (if applicable).
- [ x] The changes pass all existing tests.

## Additional work required
In order to work, the workflow requires:
- A Docker organization (let's say *lichess-bot-devs*) to be created in Docker Hub (I can help if required).
- A `DOCKER_ORGANIZATION` action repository variable containing the name of the Docker Hub organization.
- A `DOCKER_IMG` action repository variable containing the name of the Docker image (let's say *lichess-bot*).
- Two action repository secrets `DOCKER_USER` and `DOCKER_PWD` containing, respectively, the login of the user that has rights to push to the docker hub organization and a token of this user with read, write and delete permissions (available in [https://hub.docker.com/settings/security](https://hub.docker.com/settings/security)).

## How it works?
Every push on master branch will trigger the `build and publish` of the image with the `latest` tag.  
The workflow (Docker CI) can also be executed manually in the Github interface. In this case, it will ask for the tag to apply to the built image.

Probably this process can be improved.
Some pushes (typically changes in wiki folder) may not require an image update. But I was not sure to have a safe way to distinguish those minor updates.
I also added the manual trigger because I was wondering how I could reliably obtain a release version.

## How can you see work before accepting the PR?
My fork built the `fathzer/lichess-bot` image available in Docker hub. Feel free to test it!

PS : Be careful, I'm a newbie in Github actions, and I don't known Python at all ... but I know a little Docker (I have been using it in a lot of professional or personal projects since 2014).


AttackingOrDefending:
I haven't used docker before and I have run into some problems while trying to test your docker. I have downloaded docker desktop and downloaded your image through it.
When I run `docker run -d -v C:\Users\panti\Downloads:/engine --env OPTIONS=-v fathzer/lichess-bot` I get a long hex string and then it quit. In the desktop app, the logs are:
```
Exception: The engine                              
2024-04-02 21:26:17                              ./engines/stockfish_24031109_x6                    
2024-04-02 21:26:17                              4_bmi2.exe file does not exist.
```
The structure is:
```
C:\USERS\PANTI\DOWNLOADS
│   config.yml
│   stockfish_24031109_x64_bmi2.exe
│
└───engines
        stockfish_24031109_x64_bmi2.exe
```
It is able to read the `config.yml` file but not Stockfish.

fathzer:
Hi,

The first thing is as mentioned in 'wiki/How-to-use-the-Docker-image.md': *be aware that this program will run inside the container ... in a [Linux Alpine OS](https://www.alpinelinux.org/)*.
It seems your are trying to run an .exe ... on a linux OS. For sure it can't work.

Nevertheless, I tried to launch Stockfish inside this image and I didn't succeed. Looking at Stockfish web site, there's a lot of Linux different versions. I tried with POPCNT and AVX2. Both do not work in the container (but run on the Ubuntu subsystem on my PC). Maybe there's something wrong between Stockfish and Alpine Linux (it's a very light Linux system, maybe something required by Stockfish is missing?).

I had a look at the [Dockerfile of a Stockfish docker image](https://github.com/SleepyMario/stockfish-docker) based on Alpine. And it seems not as simple as executing the compiled file in the package (they do a make).

I will investigate more tomorrow (now it's late in France ;-)).

AttackingOrDefending:
I downloaded SF for linux and the new image from docker hub but it still won't work because it doesn't find the executable, even though it finds `config.yml`. The folder contents are:
```
C:\USERS\PANTI\DOWNLOADS
│   config.yml
│   stockfish_24040206_x64_bmi2
│
└───engines
        stockfish_24040206_x64_bmi2
```
I'm still using Windows so I don't know if that is the problem, but I don't think it is since the error is 
```
2024-04-04 21:42:59                              Exception: The engine                              
2024-04-04 21:42:59                              engines/stockfish_24040206_x64_                    
2024-04-04 21:42:59                              bmi2 file does not exist.
```
The command I am using: `docker run -d -v C:\Users\panti\Downloads:/engine --env OPTIONS=-v fathzer/lichess-bot`

> My proposal is to do as others do: Several variants.
I haven't pushed this into my PR yet (I will if it seems reasonable to you), 

If other engines can run on alpine (e.g. Leela, Berserk, Caissa, Seer), I think it is fine to have two versions.

fathzer:
Hi,

I think the problem is in your `config.yml` file. If I understand well how lichess-bot works, it seems that lichess-bot searches for the executable in the lichess-bot relative path `engines`, but it is not there, you have a copy in the absolute path `/engine`.

Here is the configuration I use:
Folder content:
```
C:\Users\jeanm\Desktop\lichess-docker-example
    config.yml
    stockfish
          stockfish_24040206_x64_bmi2
          ... (other files from the stockfish package)
```
config.yml:
```
engine:
  dir: "/engine/stockfish"
  name: "stockfish-ubuntu-x86-64-bmi2"
  working_dir: "/engine/stockfish"
```

And the command : `docker run -d -v C:\Users\jeanm\Desktop\lichess-docker-example:/engine --env OPTIONS=-v fathzer/lichess-bot`

If you want to inspect what is really in your container, you can launch a shell with the same configuration using: `docker run -ti -v C:\Users\jeanm\Desktop\lichess-docker-example:/engine --env OPTIONS=-v fathzer/lichess-bot /bin/sh`

If this does not fix the problem, please send the docker container's log (you can retrieve it in the *containers* tab of Docker desktop).

For the Alpine variant, I am pretty sure that most C engines may have some compatibility problem with it. But, fortunately, there's a lot of home made engines on Lichess for which Alpine will be a far better option. For instance ... [mine](https://lichess.org/@/fathzer-jchess) ;-)
It is written in Java and using the fat standard image would only have disadvantages (size and vulnerabilities).

fathzer:
Hi,
I just pushed a commit that create 2 image flavors (a *fat* one, and the alpine one). I have also updated `wiki/How-to-use-the-Docker-image.md`.

AttackingOrDefending:
Thanks for the help in setting the docker up. It works now and can play a game.

Would it be possible to support homemade engines and extra game handlers? Is there any way that users can use custom `homemade.py` and `extra_game_handlers.py` without needing to build their own image? When I tried using a homemade engine I got:
```
2024-04-06 22:13:03                              AttributeError: module                             
2024-04-06 22:13:03                              'homemade' has no attribute                        
2024-04-06 22:13:03                              'CustomBot'
```

Also, can the logs in `lichess_bot_auto_logs` be written outside the docker? I see that docker desktop keeps the logs, so I'm not sure if we need this feature at the moment, but it could be a additional feature.

fathzer:
> Would it be possible to support homemade engines and extra game handlers? Is there any way that users can use custom `homemade.py` and `extra_game_handlers.py` without needing to build their own image?

Yes :-)
Docker allows you to mount your own files into the container's file system. The python files are located in `/lichessbot` (I have to add this information to documentation). So, mounting a file to `/lichessbot/homemade.py` or `/lichessbot/extra_game_handlers.py` works. Something like:
`docker run -d -v C:\Users\me\lichess-docker-example:/engine -v C:\Users\me\lichess-docker-example\myhomemade.py:/lichessbot/homemade.py --env OPTIONS=-v fathzer/lichess-bot`
Of course, if those files require some extra pip installation, it will not work.
But Docker is very versatile ;-) You can also change the launch command to make some pip install before launching `lichess-bot.py`. But making your own image seems a better solution to me.

fathzer:
> Also, can the logs in `lichess_bot_auto_logs` be written outside the docker? I see that docker desktop keeps the logs, so I'm not sure if we need this feature at the moment, but it could be a additional feature.

Ooops, I didn't noticed lichess-bot was writing logs in its file system! It's a (little) problem. Is there any way to deactivate this output?

AttackingOrDefending:
By running `python3 lichess-bot.py --disable_auto_logging` the bot won't save the logs to `lichess_bot_auto_logs`.

Users can also save the games to a local folder by enabling the option in `config.yml`. If this is also a problem, then you can use the code [here](https://github.com/lichess-bot-devs/lichess-bot/compare/master...AttackingOrDefending:lichess-bot:docker?expand=1) to warn users about it. It requires setting the environment variable `LICHESS_BOT_DOCKER` to work.

How to use `homemade` engines and `extra_game_handlers` should also be mentioned in `How-to-use-the-Docker-image.md`.

fathzer:
> By running `python3 lichess-bot.py --disable_auto_logging` the bot won't save the logs to `lichess_bot_auto_logs`.

Sorry, I should have seen this!
I propose that the Docker image always uses this option, as duplicating logs in a file sounds strange in the context of Docker (I will add this in documentation).

> Users can also save the games to a local folder by enabling the option in `config.yml`. If this is also a problem, then you can use the code [here](https://github.com/lichess-bot-devs/lichess-bot/compare/master...AttackingOrDefending:lichess-bot:docker?expand=1) to warn users about it. It requires setting the environment variable `LICHESS_BOT_DOCKER` to work.

I will add the `LICHESS_BOT_DOCKER` environment variable so that the Python process can detect that it is running in Docker.
Saving the games can be a *must have* for some users, and, if the user mounts a volume to save the games, he will have no problems (Unfortunately, I think there's no way from inside the container to check if a volume is mounted or not).
So, maybe, the warning message can be less dissuasive? I propose something like "*Games will be saved to xxx, please ensure this is a mounted volume; Using the Docker's container internal file system will prevent you accessing the saved files and can lead to disk saturation.*"
Of course, I will add a word about that in `How-to-use-the-Docker-image.md`.

> How to use `homemade` engines and `extra_game_handlers` should also be mentioned in `How-to-use-the-Docker-image.md`.

I will add this :-)

AttackingOrDefending:
> I propose that the Docker image always uses this option, as duplicating logs in a file sounds strange in the context of Docker (I will add this in documentation).

That's ok, but I think we should recommend users to use `OPTIONS=-v` to make debugging easier.

> So, maybe, the warning message can be less dissuasive? I propose something like "Games will be saved to xxx, please ensure this is a mounted volume; Using the Docker's container internal file system will prevent you accessing the saved files and can lead to disk saturation."

This is a good warning.

fathzer:
> > Unfortunately, I think there's no way from inside the container to check if a volume is mounted or not
> 
> Could [this](https://stackoverflow.com/a/71000025) help?

Not really ;-). The exposed solution requires the user to remember to mount `/var/run/docker.sock` ... and to have Docker installed in the container. Adding an environment variable into the image seems much safer (It only fails if the user has intentionally overwritten the environment variable in its run command).

fathzer:
Hi,

Since yesterday, I've been wondering about the complexity of the Docker volume mounts needed to use a custom `homemade.py` or `extra_game_handlers.py`.
Currently, you have to make a mount for the `config.yml`, another for the `homemade.py` and yet another for the `extra_game_handlers.py`. It seems too complex to me.
I think that a solution requiring only the mount of a directory containing all these elements would be more elegant and simple to use.

The problem is `homemade.py` and `extra_game_handlers.py` have to be in the lichess-bot root directory and Docker doesn't allow selective mount that would only replace some files of the image by the ones in the docker host directory and preserve others. In other words, if the user mounts a directory with `config.yml`, `homemade.py` and `extra_game_handlers.py` at the location of `lichess-bot.py`, it will erase the whole content of this directory (`lichess-bot.py`, the libraries, etc...).

Here are the solutions I considered (in my order of preference ... But I must admit that my opinion is not very informed). All have in common a single volume mount to a path to be chosen (inside or outside of the lichess-bot root directory), let's say `mount`:
1. Make `engine_wrapper.py` and `model.py` able to import  `homemade.py` and `extra_game_handlers.py` directly from `mount`. Of course, to preserve compatibility, they would do that only when the program is launched under Docker (using an environment variable to check if lichess_bot is running under Docker). As I said, I know nothing about Python, but Google told me [it can be done](https://www.squash.io/how-to-import-files-from-different-folder-in-python/#:~:text=One%20way%20to%20import%20files,were%20in%20the%20current%20directory.).
I suppose we should also modify `engine_wrapper.py` and `model.py` to make them able to work from `mount`.
Maybe I'm dreaming ;-)
2. Have a *pre-launch script* that copies `engine_wrapper.py`, `model.py`, `lichess-bot.py`, `lib`, etc ... to a `work` subfolder of the mounted volume. That would slow down the container start a bit. Maybe one could have permission problems if the container does not work as root (running as root is a vulnerability)  ... but works with the current structure and avoids writing in the container's file system (see next proposal).
3. Have a *pre-launch script* that copies `homemade.py` and `extra_game_handlers.py` beside `lichess-bot.py`. This is not ideal because writing into the container's file system is always tricky (as explained in previous posts). Nevertheless, the *pre-launch script* could check if the copy is required before making it (typically, if the container is restarted and the files currently in the root directory have a modification date equal to the ones in `mount`, there's no need to make a copy).

Of course, we can stay on what is currently running!
Any comment or suggestion is welcome ;-)

AttackingOrDefending:
I too prefer option 1.

fathzer:
> I too prefer option 1.

Nice :-)
I found a very simple solution for `homemade.py`. I've just replaced, in the image, the provided `homemade.py` by a symbolic link to `/engine/homemade.py`. It works even if the file is missing in the mounted volume (for example if the user wants to run an uci engine).

I tried the same with `extra_game_handlers.py`, but, unfortunately, I got an error if the file is missing in the mounted volume. I will investigate a bit more ...

fathzer:
Hi,

I've updated the pull request with the following changes:
- It implements the [first option](https://github.com/lichess-bot-devs/lichess-bot/pull/936#issuecomment-2043038504) discussed earlier.
  - The docker image contains a copy of `extra_game_handlers.py` in `lib\extra_game_handlers_failover.py` (the copy is made while building the image to allow changing `extra_game_handlers.py` on your side with no need to duplicate changes anywhere).
  - `lib/engine_wrapper.py` and `lib/model.py` now search for  `lib.extra_game_handlers_failover` if importing `extra_game_handlers` fails.
- It adds `LICHESS_BOT_DOCKER` environment variable to help Python program detecting if it is running in a Docker container.
- It changes installation direction to `/lichess-bot` (instead of `/lichessbot`), to make it consistent with the project name.
- It changes the configuration folder path to `/lichess-bot/config` instead of `/engine`. This is more consistent with the Linux best naming practices for configuration folders. **Warning**: The command to run the image changes (the volume should be mounted to `/lichess-bot/config` instead of `/engine`.
- It adds the warning message you proposed if the docker container is configured to output pgn files.
- It prevents Python from writing to containers (set PYTHONDONTWRITEBYTECODE to 1).
- It merges `docker/Dockerfile` and `docker/Dockerfile-alpine` to ease maintenance.
- It updates the documentation.

Tell me if you want some more changes or if I made something wrong.

wiki/How-to-use-the-Docker-image.md
```diff
@@ -0,0 +1,63 @@
...
+
+## What if I want to add options to ```lichess-bot.py```?
+
+If you want to pass some options to the ```lichess-bot.py``` executed in the container, add them in the ```OPTIONS``` environment variable.  
+For instance, to launch the bot in verbose mode, run the command:  
+```docker run -d -v /home/me/myEngine:/engine --env OPTIONS=-v lichess-bot-devs/lichess-bot```
+
+## How to know which release of lichess-bot is running?
+Use the following command: ```docker run --rm --entrypoint=cat lichess-bot-devs/lichess-bot lib/versioning.yml```
+
```
AttackingOrDefending, line 58: Leave only one empty line at the end.

AttackingOrDefending:
In the wiki, instead of `/home/me/myEngine:/engine` it should be `/home/me/myEngine:/lichess-bot/config`, and instead of `/home/me/myEngine/extra_game_handlers.py` is should be `/lichess-bot/extra_game_handlers.py`. In general mentions of `/engine` and `/lichessbot` need updating.

I believe we should also mention that `homemade.py` is set up like `extra_game_handlers.py`.

I also think we should mention that in the `config.yml` all paths have to start with `/lichess-bot` and for `dir` and `working_dir` `/lichess-bot/config/`.

Lastly, I think we can look into publishing the images also in github packages (see [here](https://docs.github.com/en/actions/publishing-packages/publishing-docker-images)). It has a maximum stotage of 500MB, but maybe we can get under it with some `zip` or `7z` compression.

fathzer:
> In the wiki, instead of `/home/me/myEngine:/engine` it should be `/home/me/myEngine:/lichess-bot/config`, and instead of `/home/me/myEngine/extra_game_handlers.py` is should be `/lichess-bot/extra_game_handlers.py`. In general mentions of `/engine` and `/lichessbot` need updating.
> 
> I believe we should also mention that `homemade.py` is set up like `extra_game_handlers.py`.
>
Sorry, I didn't update the wiki in the right branch of my project! Now it's fixed.

> I also think we should mention that in the `config.yml` all paths have to start with `/lichess-bot` and for `dir` and `working_dir` `/lichess-bot/config/`.

Done

> Lastly, I think we can look into publishing the images also in github packages (see [here](https://docs.github.com/en/actions/publishing-packages/publishing-docker-images)). It has a maximum stotage of 500MB, but maybe we can get under it with some `zip` or `7z` compression.

I don't know Github package at all, and, to be honest, its interest for public packages is not obvious (Docker hub is the official repository that every Docker user know, like Maven central for Java packages).
Nevertheless, I will have a look at it (not sure to have time this week-end). I already saw it seems to be free and unlimited for publishing public packages :-)

fathzer:
Hi,
I've updated the workflow to publish the images to Docker hub and Github container registry.
It requires that you:
1 - Create a token as explained [here](https://www.youtube.com/watch?v=p12ORMWcn5E).
2 - Create the `GB_REPO_PWD` action secret with the token as value.
3 - Create the `GB_REPO_USER` action secret with `lichess-bot-devs` as value.
4 - Make the package public in the danger zone of its [settings](https://github.com/users/lichess-bot-devs/packages/container/lichess-bot/settings). This step can only be achieved after the first image deployment.

I think we should also talk about versioning. It seems you do not use *Github release* to tag the releases of lichess-bot. Instead, every push to master triggers the `Versioning` workflow that increments the version number.

Currently the `Docker image` workflow is also triggered at every push to master (it then creates the `latest` tagged images), or manually (to create images with a version tag). This is far from being optimal.

I could restrict the push to master to ignore changes in `wiki`, `docs` or `README.md` to prevent creating images with no changes.

Another solution would be to trigger `Docker image` only when the `Versioning` workflow ends successfully or when `lib/versioning.yml` file is changed. Then, `Docker image` could retrieve the version number (in the `lib/versioning.yml` file) and create the images with their version and latest tags.
Nevertheless, to prevent creating new images, with new versions, every time a push is made in `master` branch, even if one simply corrected a typo in the wiki, it would probably better to restrict the execution of `Versioning` to relevant changes.

What is your opinion?

AttackingOrDefending:
I think we should just run the workflow only when the `lib/versioning.yml` file is changed.

fathzer:
> I think we should just run the workflow only when the `lib/versioning.yml` file is changed.

Done 👍 
I also modified the workflow to tag images with the version found in `lib/versioning.yml` assuming that this file will always contain a line starting with *lichess_bot_version:* followed by the version number.

fathzer:
> Can you fix the errors in the workflows (Python Build and Mypy)?
> 
> I have made some changes [here](https://github.com/AttackingOrDefending/lichess-bot/tree/docker2) that will fix the errors if you want to use it, but you can fix the errors in a different way if you want.

Hi,
I pushed some fixes that seems to work (It works in my environment, but I'm not very confident when modifying python code).
In order to fix `MyPy` workflow, I used another solution I've found [here](https://github.com/python/mypy/issues/1153#issuecomment-253842414). I made that because it seemed to me that in your proposal, the name of `game_specific_options` function should be changed to `game_specific_options_docker` when using Docker.
But, being totally incompetent in Python, maybe I just didn't understand anything and my solution is not good at all.
Feel free to tell me if that's the case.

fathzer:
> > Can you fix the errors in the workflows (Python Build and Mypy)?
> > I have made some changes [here](https://github.com/AttackingOrDefending/lichess-bot/tree/docker2) that will fix the errors if you want to use it, but you can fix the errors in a different way if you want.
> 
> Hi, I pushed some fixes that seems to work (It works in my environment, but I'm not very confident when modifying python code). In order to fix `MyPy` workflow, I used another solution I've found [here](https://github.com/python/mypy/issues/1153#issuecomment-253842414). I made that because it seemed to me that in your proposal, the name of `game_specific_options` function should be changed to `game_specific_options_docker` when using Docker. But, being totally incompetent in Python, maybe I just didn't understand anything and my solution is not good at all. Feel free to tell me if that's the case.

Ooooops, Please ignore this message. It seems nothing is working!!!
I will have a look ...

fathzer:
> Ooooops, Please ignore previous comment. It seems nothing is working anymore!!! I will have a look ...

Ok ... I fixed some missing white spaces and empty lines. It should be better now.

AttackingOrDefending:
You can see the errors here. You should probably add a `# type: ignore` also to `engine_wrapper.py`. Can you test changing `# type: ignore` to `# type: ignore[import-untyped]`?

fathzer:
> You can see the errors here. You should probably add a `# type: ignore` also to `engine_wrapper.py`. 

done

> Can you test changing `# type: ignore` to `# type: ignore[import-untyped]`?

It didn't work. I had to put `# type: ignore[import-untyped,no-redef]` to fix it (I hope it is not a problem).

AttackingOrDefending:
`# type: ignore[import-untyped,no-redef]` is still better than `# type: ignore`. By changing the name to `game_specific_options_docker` as I did in my branch you can fix the `no-redef` error, so you will only need to put `# type: ignore[import-untyped]`.

fathzer:
I don't understand what is the problem with ignoring `no-redef` errors on lines concerned by what look like a myPi false positive to me.
Nevertheless, I tried your proposal [here](https://github.com/fathzer/lichess-bot/tree/fixes_wkflw2) but, unfortunately, [wasn't able to have `Python build` workflow work](https://github.com/fathzer/lichess-bot/actions/workflows/python-build2.yml).

AttackingOrDefending:
Removing [line 101](https://github.com/fathzer/lichess-bot/blob/fixes_wkflw2/lib/model.py#L101) should fix the issue with `python-build`.

> I don't understand what is the problem with ignoring no-redef errors on lines concerned by what look like a myPi false positive to me.

We try to limit the amount of `# type: ignore` comments we use. We have used a similar approch [here](https://github.com/lichess-bot-devs/lichess-bot/blob/master/lib/engine_wrapper.py#L624-L638) for the same reason.

fathzer:
> Removing [line 101](https://github.com/fathzer/lichess-bot/blob/fixes_wkflw2/lib/model.py#L101) should fix the issue with `python-build`.

Yes 👍 

> > I don't understand what is the problem with ignoring no-redef errors on lines concerned by what look like a myPi false positive to me.
> 
> We try to limit the amount of `# type: ignore` comments we use. We have used a similar approch [here](https://github.com/lichess-bot-devs/lichess-bot/blob/master/lib/engine_wrapper.py#L624-L638) for the same reason.

Ok, so I've modified your proposal in `model.py` to remove the `# type: ignore` comment.
Probably something similar could be done in `engine_wrapper.py` ... by someone who knows Python (I think I've already written too much Python during this life ;-)). Then I would remove the copy of `extra_game_handlers.py` in `lib/extra_game_handlers_failover.py`.
 
> When I tried using the homemade engines that come with `lichess-bot` (e.g. `FirstMove`) without providing a custom `homemade.py` file, I got an error:
> 
> ```
> 2024-04-20 16:00:14                              ModuleNotFoundError: No module                     
> 2024-04-20 16:00:14                              named 'homemade'
> ```

If you don't provide any `homemade.py` in the volume mounted, the `homemade.py` link of the container is linked to nothing.
If you want `homemade.py` to have a failover, your Python code should be modified to look for this failover. On the other hand, I can simply add a warning about the fact that providing a `homemade.py` file is mandatory to use a home made engine.

AttackingOrDefending:
Is there any flag that you can add to [line 17](https://github.com/fathzer/lichess-bot/blob/docker-image/docker/Dockerfile#L17), to only move the file if it exists? If there isn't just add a warning and we can fix this in a later PR.

fathzer:
> Is there any flag that you can add to [line 17](https://github.com/fathzer/lichess-bot/blob/docker-image/docker/Dockerfile#L17), to only move the `homemade.py` file if it exists? If there isn't just add a warning and we can fix this in a later PR.

Unfortunately, the Dockerfile content (except the `CMD` line) is executed at build time and is written indelibly in the image.
I will add the warning.

fathzer:
> > We try to limit the amount of `# type: ignore` comments we use. We have used a similar approch [here](https://github.com/lichess-bot-devs/lichess-bot/blob/master/lib/engine_wrapper.py#L624-L638) for the same reason.
> 
> [...] something similar could be done in `engine_wrapper.py` by someone who knows Python (I think I've already written too much Python during this life ;-)). Then I would remove the copy of `extra_game_handlers.py` in `lib/extra_game_handlers_failover.py`.

I've just had an idea to do that easily, and to avoid polluting your code with try/except around imports of `extra_game_handlers.py` (the problem with `homemade.py` would reamain). I will check if it works today ...

fathzer:
> I've just had an idea [...]

I've pushed the implementation of the following idea:
Handling `extra_game_handlers` fail over in many places in the code is (in my opinion) not a good practice.

Instead, I implemented a proxy for the `extra_game_handler` functions in `lib/extra_game_handlers_wrapper` (maybe `lib/extra_game_handlers_proxy` would be a better name, but *wrapper* looks familiar in your code).

This proxy contains the ugly try/except hacks and is used wherever `extra_game_handlers` is useful, with a *standard* clean import.

There's also no longer a copy of `extra_game_handler` in `lib/extra_game_handlers_failover.py`. Now, local and Docker hosted instances run the exact same code.

PS: The problem with missing `homemade.py` remains (at least until a future PR :-)).

AttackingOrDefending:
That was a nice change. Now we don't have to use `# type: ignore` comments.

@MarkZH We have to first create an account on Docker Hub and the Github container registry before we can merge this.

AttackingOrDefending (approved):

AttackingOrDefending:
Can you change `secrets.DOCKER_PWD` to `secrets.DOCKER_TOKEN`?

MarkZH:
@fathzer Thanks. I got it working.

MarkZH (approved):

fathzer:
> Can you change `secrets.DOCKER_PWD` to `secrets.DOCKER_TOKEN` and `GB_REPO_PWD` to `GB_REPO_TOKEN`?

done :-)

AttackingOrDefending:
@MarkZH Should I merge this?

MarkZH:
> @MarkZH Should I merge this?

Go ahead.

PR #939, opened 2024-04-06
NanoGPTEngine Wrapper Complete

vincentlinzhu:
## Type of pull request:
- [ ] Bug fix
- [ X] Feature
- [ ] Other

## Description:
- Added in an argument to the main arg_parser to take in a weight file and use that as the model for evaluation against other bots.
- Implemented a homemade engine called NanoGPTEngine that wraps the model passed in from the weight file. 
- That engine uses NanoGptPlayer module to make moves in a chess game. The module uses the model to decide on the best move to play in the current board/game state.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ lichess_bot_auto_logs/*
 *.log
 *.txt
 TEMP/*
+venv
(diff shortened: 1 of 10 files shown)


MarkZH:
We won't be merging these changes into lichess-bot because they are specific to your engine. They would break lichess-bot for all other users.

Feel free to keep your fork customized to your engine to continue  your work. It looks very interesting.

PR #949, opened 2024-04-30
Remove file change condition

AttackingOrDefending:
## Type of pull request:
- [x] Bug fix
- [ ] Feature
- [ ] Other

## Description:

Since the commit is made thorugh github actions, github doesn't rerun the workflows, so the docker is not built.

## Related Issues:

Related to #936 

## Checklist:

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,8 +4,6 @@ on:
   push:
     branches:
       - master
-    paths:
-      - lib/versioning.yml
 
 env:
   GB_REGISTRY: ghcr.io
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,9 +1,10 @@
 name: Docker CI
 
 on:
-  push:
-    branches:
-      - master
+  workflow_run:
+    workflows: [Versioning]
+    types:
+      - completed
 
 env:
   GB_REGISTRY: ghcr.io
@@ -11,6 +12,7 @@ env:
 jobs:
   build:
     runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
     steps:
       - uses: actions/checkout@v4
       


fathzer:
Hi,

Maybe adding `/.github/workflows/docker.yml` is better than removing the whole file condition?  

The idea behind this file condition is a first step to prevent several strictly identical images with different version numbers.
It Assumes that the `lib/versioning.yml` is a safe way to track new versions. Unfortunately, if I understand correctly how the `versioning.yml` workflow works, it is currently not: A new version number is generated each time there's a push into the `master` branch, even if it is a correction in the wiki.

Nevertheless, if the file condition remains, it simply remains to fix the `versioning.yml` action to prevent incrementing `lib/versioning.yml` when nothing changes in the application (probably with an appropriate file condition on `versioning.yml` workflow).

But maybe I'm wrong about the meaning of `lib/versioning.yml`?

fathzer:
> Maybe adding `/.github/workflows/docker.yml` is better than removing the whole file condition?

Ooops, this will not work, as it seems the push made by `versioning` workflow does not trigger any workflow (Sorry, I tested it changing myself the `lib/versioning.yml` file).
I will test a solution based on `versioning` workflow completion ...

fathzer:
In order to have `docker CI` running when `Versioning` succeeds, you should replace the following in `docker.yml`:

```
on:
  push:
    branches:
      - master
    paths:
      - lib/versioning.yml

env:
  GB_REGISTRY: ghcr.io
  
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
```

by this:

```
on:
  workflow_run:
    workflows: ["Versioning"]
    types:
      - completed

env:
  GB_REGISTRY: ghcr.io
  
jobs:
  build:
    runs-on: ubuntu-latest
    if: ${{ github.event.workflow_run.conclusion == 'success' }}
    steps:
      - uses: actions/checkout@v4
```

Warning:
- This will only work on the default branch (extract from the [documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run): *Note: This event will only trigger a workflow run if the workflow file is on the default branch.*).
- Don't forget the `if: ${{ github.event.workflow_run.conclusion == 'success' }}` in the `build` job; It prevents the job from running if the `Versioning` workflow is triggered but fails.

AttackingOrDefending:
I implemented your change. I removed the quotes because the examples in the docs didn't have them.

MarkZH:
Ready to merge?

AttackingOrDefending:
Yes.

MarkZH (approved):

PR #952, opened 2024-05-04
Add more type hints

### LINKED ISSUE #921: Improving type annotations
Many places in the code are annotated with `Any` types that could use some more specific type annotations. This is especially true when parsing user's configuration files and JSON messages from lichess.org.

AttackingOrDefending:
## Type of pull request:
- [ ] Bug fix
- [x] Feature
- [ ] Other

## Description:

Adds more type hints using `TypedDict`.

## Related Issues:

closes #921 

## Checklist:


AttackingOrDefending:
There are 8 errors remaining.
```
lib\lichess.py:327: error: Incompatible types in assignment (expression has type "dict[str, Any]", variable has type "UserProfileType")  [assignment]
lib\lichess.py:397: error: Incompatible return value type (got "dict[str, Any]", expected "PublicDataType")  [return-value]
lib\engine_wrapper.py:306: error: Incompatible types in assignment (expression has type "TypedDict({'score'?: PovScore, 'pv'?: list[Move], 'depth'?: int, 'seldepth'?: int, 'time'?: float, 'nodes'?: int
, 'nps'?: int, 'tbhits'?: int, 'multipv'?: int, 'currmove'?: Move, 'currmovenumber'?: int, 'hashfull'?: int, 'cpuload'?: int, 'refutation'?: dict[Move, list[Move]], 'currline'?: dict[int, list[Move]], 'ebf'?: float, 'wdl'?: PovWdl, 'string'?: str})", variable has type "MoveInfoType")  [assignment]
lib\engine_wrapper.py:372: error: Returning Any from function declared to return "str"  [no-any-return]
lib\engine_wrapper.py:372: error: TypedDict key must be a string literal; expected one of ("Evaluation", "Winrate", "Hashfull", "Nodes", "Speed", ...)  [literal-required]
lib\engine_wrapper.py:372: error: TypedDict key must be a string literal; expected one of ("Evaluation", "Pv", "Depth", "Seldepth", "Movetime", ...)  [literal-required]
lib\engine_wrapper.py:374: error: TypedDict key must be a string literal; expected one of ("Evaluation", "Pv", "Depth", "Seldepth", "Movetime", ...)  [literal-required]
lib\engine_wrapper.py:393: error: Incompatible types in assignment (expression has type "dict[str, Any]", variable has type "ReadableMoveInfoType")  [assignment]
Found 8 errors in 2 files (checked 18 source files)
```

After I clean the mess in in `get_stats` in `engine_wrapper`, this could be merged. I expect to end with around 4 errors that I don't think I can fix, but we could add an ignore.

AttackingOrDefending:
The only uses of `Any` that are left are on `engine_wrapper.py` on the definition of homemade engines, and on `config.py` on the raw dict.

lib/model.py
```diff
@@ -6,25 +6,25 @@
...
         self.rated = challenge_info["rated"]
         self.variant = challenge_info["variant"]["key"]
         self.perf_name = challenge_info["perf"]["name"]
         self.speed = challenge_info["speed"]
-        self.increment: int = challenge_info.get("timeControl", {}).get("increment")
-        self.base: int = challenge_info.get("timeControl", {}).get("limit")
-        self.days: int = challenge_info.get("timeControl", {}).get("daysPerTurn")
+        self.increment: int = challenge_info.get("timeControl", {}).get("increment", -1)
+        self.base: int = challenge_info.get("timeControl", {}).get("limit", -1)
+        self.days: int = challenge_info.get("timeControl", {}).get("daysPerTurn", -1)
```
MarkZH, line 27: Why is -1 used as a default value here? This conflicts with the check in `is_supported_time_control()` below, which checks for `None`. I think changing the type of these members to `Optional[int]` is better.

lib/engine_wrapper.py
```diff
@@ -922,16 +924,17 @@ def get_opening_explorer_move(li: LICHESS_TYPE, board: chess.Board, game: model.
                               ) -> tuple[Optional[str], chess.engine.InfoDict]:
     """Get a move from lichess's opening explorer."""
     wb = "w" if board.turn == chess.WHITE else "b"
-    time_left = msec(game.state[f"{wb}time"])
+    time_left = msec(game.state["wtime"]) if board.turn == chess.WHITE else msec(game.state["btime"])
```
MarkZH, line 927: Why was this change made (and similar changes throughout like line 984)?

AttackingOrDefending: Because `game.state` only supports `wtime` and `btime` and not all `{text}time` options. We know that `wb` has a value of `"w"` or `"b"`, so `f"{wb}time"` will be either `wtime` or `btime`, but mypy doesn't know this. I could attempt to fix this using [literals](https://mypy.readthedocs.io/en/stable/literal_types.html#parameterizing-literals), like this.
```
# OPTION A
wb: Literal["w", "b"] = "w" if if board.turn == chess.WHITE else "b"
time_left = msec(game.state[f"{wb}time"])

# OPTION B
wbtime: Literal["wtime", "btime"] = "wtime" if if board.turn == chess.WHITE else "btime"
time_left = msec(game.state[wbtime])
```
I don't know if these options work, but I can try later.

MarkZH: If the type is `Literal` then the assigned value has to be a literal. This works:
```python
    is_white = board.turn == chess.WHITE
    wbtime: Literal["wtime", "btime"] = "wtime" if is_white else "btime"
    time_left = msec(game.state[wbtime])
```
Since we need `wtime` and `btime` everywhere, I think a function would work better and keep the code simple with many fewer `Literal` declarations:
```python
def wbtime(board: chess.Board) -> Literal["wtime", "btime"]:
    return "wtime" if board.turn == chess.WHITE else "btime"
```
Similar functions for `winc`/`binc` and others would be useful, as well.

AttackingOrDefending: I added the function for `wbinc` in `engine_wrapper` even though it's not used there, just so it is next to `wbtime`.

MarkZH (approved):

MarkZH:
Merge if there are no more changes.

PR #967, opened 2024-05-26
Allow all commands that start with `!eval` to overcome spam rules

eduherminio:
## Type of pull request:
- [ ] Bug fix
- [x] Feature
- [ ] Other

### Context

When watching bot games, chat spam become very strict when trying to send the same command multiple times, which usually happens for `!eval` command.
[This seems to be the piece of code](https://github.com/tsoj/lila/blob/b1f33443d61aba16dd0d66a678ea6fec77b52f3d/modules/security/src/main/Flood.scala#L19C7-L19C19) responsible of detecting a message as spam. It basically detects duplicated messages (thanks [@tsoj](https://github.com/tsoj)!)

There was an attempt to improve this on lichess side [Relaxing spam rules for when being a normal user alone with just BOT accounts in chat](https://github.com/lichess-org/lila/pull/12554), but it didn't make it.

### Proposal

We can improve spectator experience on our side by allowing multiple commands to return our bot's eval.
The easiest way I could think of is doing so for every command that starts with (`!eval`), i.e. `!eval1234`, `!evalplease`, `!eval with spaces`.

This change has been applied to [my bot](https://lichess.org/@/Lynx_BOT) and can be verified right now by watching any of his games and entering those commands one after each other.

## Checklist:

- [ ] I have added necessary documentation (if applicable).
- [ ] The changes pass all existing tests.

image

### DIFF
--- a/lib/conversation.py
+++ b/lib/conversation.py
@@ -67,6 +67,7 @@ class Conversation:
         :param cmd: The command to react to.
         """
         from_self = line.username == self.game.username
+        is_eval = cmd.startswith("eval")
         if cmd == "commands" or cmd == "help":
             self.send_reply(line, "Supported commands: !wait (wait a minute for my first move), !name, !howto, !eval, !queue")
         elif cmd == "wait" and self.game.is_abortable():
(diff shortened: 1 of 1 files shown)


eduherminio:
If this moves forward, happy to update any related documentation before or after merging.

lib/conversation.py
```diff
@@ -67,6 +67,7 @@ def command(self, line: ChatLine, cmd: str) -> None:
         :param cmd: The command to react to.
         """
         from_self = line.username == self.game.username
+        is_eval = cmd.startswith("eval")
         if cmd == "commands" or cmd == "help":
             self.send_reply(line, "Supported commands: !wait (wait a minute for my first move), !name, !howto, !eval, !queue")
```
MarkZH, line 72: I think it would be good to replace the string here with something like: "Supported commands: !wait (wait a minute for my first move), !name, !howto, !eval (or any text starting with !eval), !queue"

Make sure the resulting string is less than 140 characters.

eduherminio: Done!

MarkZH (approved):

AttackingOrDefending (approved):

PR #978, opened 2024-06-18
Append block list from challenge to matchmaking

### LINKED ISSUE #975: Avoid having to duplicate `block_list` in challenge and matchmaking sections
Right now, one needs to duplicate the `block_list` for both challenge and matchmaking sections if there are certain accounts they want their BOTs to avoid.

- Making the matchmaking block list default to the challenge one when left empty could be a good starting solution.
- General inheritance would be a more desirable target, allowing matchmaking to extend the blocked list from challenge section (use case: adding there engines that don't want to receive challenges from you, but you have no reason to block otherwise). A flag could be added to control this behavior (such has _inherit_challenge_block_list_) and when set to true, both lists could be merged for matchmaking.

MarkZH:
## Type of pull request:
- [ ] Bug fix
- [x] Feature
- [ ] Other

## Description:

Adds an option to the configuration file to add the entries in the `challenge: block_list` to the `matchmaking: block_list`. This avoids having to copy one block list into the other.

## Related Issues:

Closes #975

## Checklist:

N/A

### DIFF
--- a/config.yml.default
+++ b/config.yml.default
@@ -206,6 +206,7 @@ matchmaking:
 # block_list:                      # The list of bots that will not be challenged
 #   - user1
 #   - user2
+  include_challenge_block_list: false  # Do not challenge bots in the challenge: block_list in addition to the matchmaking block list.
 
 # overrides:                       # List of overrides for the matchmaking specifications above. When a challenge is created, either the default specification above or one of the overrides will be randomly chosen.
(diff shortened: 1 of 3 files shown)


AttackingOrDefending:
Shouldn't we set a default value for `include_challenge_block_list` in `config.py`?

AttackingOrDefending (approved):

Issue #920, opened 2024-03-04
Add tests for specific functions


MarkZH:
Right now, the only tests are integration tests that perform a complete run of `lichess_bot_main()` to check that there are no errors communicating with an engine and with a mocked version of the lichess.org servers. Individual functions could use unit tests, such as:

- database functions like `get_lichess_cloud_move()`
- `Lichess` class methods to make sure they still return expected responses (bot scopes, profile data, bot opponent lists, etc.)

ricor07:
Hello, I'd like to join the project by working on this issue.

AttackingOrDefending:
Thank you for your interest in contributing to the project! Feel free to work on this issue. Let us know if you have any questions or need clarification on the details.

natsu0529:
Hi, I'd like to add unit tests for Lichess methods, focusing on token scope validation, get_profile(), get_ongoing_games(), get_online_bots(), and is_online(). I'll mock all API calls so the tests don't require a token or network access.

PR #982, opened 2024-06-19
Add more tests

AttackingOrDefending:
## Type of pull request:
- [ ] Bug fix
- [ ] Feature
- [x] Other

## Description:

Add some tests for external moves. A lot more should be added before #920 can be closed.
Also fixes a bug in `engine_wrapper` found while creating these tests.

## Related Issues:

#920

## Checklist:

### DIFF
--- a/lib/engine_wrapper.py
+++ b/lib/engine_wrapper.py
@@ -1068,15 +1068,15 @@ def get_lichess_egtb_move(li: LICHESS_TYPE, game: model.Game, board: chess.Board
                 dtm *= -1
             logger.info(f"Got move {move} from tablebase.lichess.ovh (wdl: {wdl}, dtz: {dtz}, dtm: {dtm}) for game {game.id}")
         else:  # quality == "suggest":
-            best_wdl = name_to_wld[data["moves"][0]["category"]]
+            best_wdl = name_to_wld[data["moves"][0]["category"]] * -1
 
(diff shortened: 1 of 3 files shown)


lib/engine_wrapper.py
```diff
@@ -1068,15 +1068,15 @@ def get_lichess_egtb_move(li: LICHESS_TYPE, game: model.Game, board: chess.Board
...
 
             def good_enough(possible_move: LichessEGTBMoveType) -> bool:
-                return name_to_wld[possible_move["category"]] == best_wdl
+                return name_to_wld[possible_move["category"]] * -1 == best_wdl
 
             possible_moves = list(filter(good_enough, data["moves"]))
             if len(possible_moves) > 1:
                 move_list = [move["uci"] for move in possible_moves]
-                wdl = best_wdl * -1
+                wdl = best_wdl
```
MarkZH, line 1079: Am I wrong, or do these code changes result in no change in behavior?

AttackingOrDefending: This doesn't change anything. I just though that the code would be easier to understand because the correct wdl is with the ` * -1`, but we can obviously test with the wrong wdl and correct it at the end as we are currently doing. The bug was in line 1105.

MarkZH (approved):

Issue #1003, opened 2024-08-08
lichess-bot stops sending or accepting challenges


JimViebke:
**Describe the bug**

lichess-bot often sits idle after a game, without creating a challenge or acknowledging most of the incoming challenges (restarting the script manually seems to temporarily resolve this).

The timeline, per the attached [recent.log](https://github.com/user-attachments/files/16537069/recent.log):
- At 20:53:02, lichess-bot challenges blank_bot to a game. blank_bot accepts and the game is played.
- At 21:20:07, the game finishes, and lichess-bot says "Next challenge will be created after 21:21:07". I don't think it ever creates a challenge.
- At 21:31:54, lichess-bot receives a challenge from daddyshark_bot, and appears to ignore it (I think it should have accepted the challenge per the config file). This challenge does not show up in the logs again, so it is unclear if the challenge was ever cancelled.
- At 21:38:49, lichess-bot receives a challenge from chessxbot, and appears to ignore it (again, I think it should have been accepted per the config). The opponent cancels the challenge about half a minute later.
- At 22:30:45, lichess-bot receives a challenge from nakshatra3, and correctly declines it for being unrated.
- At 22:50:55, lichess-bot receives a challenge from chessxbot. As above, it appears this challenge is ignored, and the opponent cancels the challenge about half a minute later.

**To Reproduce**
1. Run lichess-bot 2024.8.6.1 with `venv\Scripts\activate` and `py lichess-bot.py -v`, and the config as specified in the attached log.

**Desktop:**
 - OS: Windows 10
 - Python Version: 3.12
 - Terminal: Windows PowerShell

(I'm not sure if this is related to my previous issue, #991. That thread links to a couple of issues and commits that could be relevant.)

AttackingOrDefending:
Setting `challenge.concurrency` to 1 should fix the problem. Matchmaking with `challenge.concurrency = 0` has some weird properties.

How the problem was caused:
[...]

PR #1004, opened 2024-08-08
Fix matchmaking bug with concurrency set to 0

AttackingOrDefending:
## Type of pull request:
- [x] Bug fix
- [ ] Feature
- [ ] Other

## Description:

The matchmaker would challenge other bots even when concurrency was set to 0, leading to the problems described in #1003. This PR fixes this.

## Related Issues:

#1003 

## Checklist:

### DIFF
--- a/lib/config.py
+++ b/lib/config.py
@@ -290,6 +290,9 @@ def validate_config(CONFIG: CONFIG_DICT_TYPE) -> None:
             config_assert(online_section.get("move_quality") != "suggest" or not online_section.get("enabled"),
                           f"XBoard engines can't be used with `move_quality` set to `suggest` in {subsection}.")
 
+    config_warn(CONFIG["challenge"]["concurrency"] > 0, "With challenge.concurrency set to 0, the bot won't accept or create "
+                                                        "any challenges.")
+
     config_assert(CONFIG["challenge"]["sort_by"] in ["best", "first"], "challenge.sort_by can be either `first` or `best`.")
     config_assert(CONFIG["challenge"]["preference"] in ["none", "human", "bot"],
                   "challenge.preference should be `none`, `human`, or `bot`.")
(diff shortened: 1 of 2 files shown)


MarkZH:
So, after this fix, setting `challenge: concurrency: 0` results in no challenges being sent or accepted. In other words, the bot does nothing. Should we print a warning and exit in this case?

MarkZH:
Something like this in config.py:
```python
config_assert(CONFIG["challenge"]["concurrency"] > 0, "challenge.concurrency must be greater than zero to play any games.")
```

AttackingOrDefending:
I used `config_warn` instead of `config_assert` because setting concurrency to 0 has some possible uses:
1. The bot's owner can create challenges through the lichess website and the bot will play the games (useful for testing).
2. After restarting lichess-bot, when you want your bot to just finish all the already started games.

AttackingOrDefending:
Also, just to note, it is still possible to make a bot not accept any challenges but still create ones, by setting `min_base` to a value larger than `max_base` (e.g. `60` when `max_base` is 30).

MarkZH:
> I used `config_warn` instead of `config_assert` because setting concurrency to 0 has some possible uses:
> 
> 1. The bot's owner can create challenges through the lichess website and the bot will play the games (useful for testing).
> 
> 2. After restarting lichess-bot, when you want your bot to just finish all the already started games.

Point 1 makes sense. For point 2, the user can set `quit_after_all_games_finish: true`.

> Also, just to note, it is still possible to make a bot not accept any challenges but still create ones, by setting `min_base` to a value larger than `max_base` (e.g. 60 when `max_base` is 30).

This should probably have a warning, as well as any other configuration with `min_` and `max_`. But, that can be another PR.

MarkZH (approved):

PR #1019, opened 2024-09-17
Simple lichess-bot.py for code organization

MarkZH:
## Type of pull request:
- [ ] Bug fix
- [ ] Feature
- [x] Other

## Description:

This is primarily a code organization PR. It moves the code in lichess-bot.py into the lib folder, leaving a simple entry point for the program.

Changes:

1. Move all code in lichess-bot.py to lib/lichess_bot.py.
2. lichess-bot.py now only calls a start function
3. Since lib.lichess_bot is now a valid name to import, `importlib` in test_bot.py is no longer needed. This allows more accurate type checking.

## Related Issues:

N/A

## Checklist:

N/A


MarkZH:
Looking at the individual commits should give a better idea of the changes in this PR.

It seems to me that the largest code file should be in the lib folder, but I'm open to disagreement.

test_bot/test_bot.py
```diff
@@ -23,7 +23,7 @@
 from lib.types import CONFIG_DICT_TYPE
 if "pytest" not in sys.modules:
     sys.exit(f"The script {os.path.basename(__file__)} should only be run by pytest.")
-lichess_bot = importlib.import_module("lichess-bot")
+import lib.lichess_bot as lichess_bot
```
AttackingOrDefending, line 26: Is there any reason we don't use `from lib import lichess_bot`?

AttackingOrDefending:
I agree with moving the code to `lib`. I was thinking of also adding a `main.py` file that contains the same code as `lichess-bot.py` so it would be a clear entry point (we would of course keep `lichess-bot.py` for backwards compatibility). What do you think about adding a `main.py`?

test_bot/test_bot.py
MarkZH: No, just forgot the syntax.

MarkZH:
I don't agree with making a main.py file. The most natural entry point for lichess-bot is lichess-bot.py. I've only seen main files in compiled languages like C or C++, but even there the final executable will be named for the program.

AttackingOrDefending (approved):

PR #1023, opened 2024-09-23
Accept engines that require an interpreter

### LINKED ISSUE #1022: Enable Running engines which are runnable jar files
My engine is written in Scala and has no binary (exe) but a runnable jar file

Instead of dir and name of the engine, I suggest to have a run-command variable, where i can enter something like: 
run-command = "java -jar myengine.jar [parameters]"

Best Regards
Jakob Erber

AttackingOrDefending:
## Type of pull request:
- [ ] Bug fix
- [x] Feature
- [ ] Other

## Description:

Allow users to provide arguements before and after the engine name (e.g. when a user is running a `jar` file as a chess engine).
[engine_options](https://github.com/lichess-bot-devs/lichess-bot/blob/master/config.yml.default#L83) could be removed as `after_name_parameters` does the same thing but for a more general use case. On the other hand, `engine_options` is easier to use and some people may depend on it, so if we remove it, we should probably deprecate it first.

## Related Issues:

closes #1022 

## Checklist:


MarkZH:
I don't think `before_name_parameters` and `after_name_parameters` are good names. The problem that this PR should solve is when a user's chess engine requires an interpreter to run. So, I think two engine subsections are needed:
```yaml
engine:
  interpreter: java
  interpreter_options:
    - "-jar"
```
The other `engine` subsections (`dir`, `name`, `engine_options`) remain the same. The command within `engine_wrapper.create_engine()` is then built like this:
```python
commands = []
if cfg.interpreter:
  commands.append(cfg.interpreter)
  commands.extend(cfg.interpreter_options or [])
commands.append(engine_path)
if cfg.engine_options:
  for k, v in cfg.engine_options.items():
    commands.append(f"--{k}={v}" if v is not None else f"--{k}")
```
I don't think there are any other cases besides interpreters for things going before the file containing the engine.

MarkZH:
Good. Now all that's needed is to document this in the wiki.

I would also comment out `interpreter_options:` in config.yml.default.

AttackingOrDefending:
For some reason the test fails, while I can run `buggy_engine.py` normally as a bot.

test_bot/test_bot.py
```diff
@@ -297,12 +297,8 @@ def engine_path(CONFIG: CONFIG_DICT_TYPE) -> str:
         name: str = CONFIG["engine"]["name"]
         return os.path.join(dir, name)
 
-    if platform == "win32":
-        CONFIG["engine"]["name"] = "buggy_engine.bat"
-    else:
-        CONFIG["engine"]["name"] = "buggy_engine"
-        st = os.stat(engine_path(CONFIG))
-        os.chmod(engine_path(CONFIG), st.st_mode | stat.S_IEXEC)
+    CONFIG["engine"]["name"] = "buggy_engine.py"
```
MarkZH, line 300: The `buggy_engine` is used for both sides of the game. The later call to `engine_path(CONFIG)` on line 305 fails to get the right name for the executable due to the change in this line. Within `engine_path()`, the `name` variable needs to be either `buggy_engine.bat` or `buggy_engine` depending on the `platform` variable. Right now, it returns `buggy_engine.py`, which can't be run by directly calling it.

.github/workflows/python-test.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         key: ${{ matrix.os }}-engines
     - name: Test with pytest
       run: |
-        pytest --log-cli-level=10
+        pytest --log-cli-level=10 -s
```
MarkZH, line 41: What does this do?

test_bot/test_bot.py
MarkZH: Here's what I came up with. There's probably better ways.
```python
    def engine_path(CONFIG: CONFIG_DICT_TYPE) -> str:
        dir: str = CONFIG["engine"]["dir"]
        name: str = CONFIG["engine"]["name"].removesuffix(".py")
        if platform == "win32":
            name = name + ".bat"
        return os.path.join(dir, name)
```

.github/workflows/python-test.yml
AttackingOrDefending: It disables output capturing by pytest (see [here](https://stackoverflow.com/questions/14405063/how-to-see-normal-print-output-created-during-pytest-run)). I changed it to get more debug information which may have helped with fixing the tests.

test_bot/test_bot.py
```diff
@@ -294,15 +294,16 @@ def test_buggy_engine() -> None:
...
     def engine_path(CONFIG: CONFIG_DICT_TYPE) -> str:
         dir: str = CONFIG["engine"]["dir"]
-        name: str = CONFIG["engine"]["name"]
+        name: str = CONFIG["engine"]["name"].removesuffix(".py")
+        if platform == "win32":
+            name += ".bat"
+        else:
+            st = os.stat(os.path.join(dir, name))
+            os.chmod(os.path.join(dir, name), st.st_mode | stat.S_IEXEC)
         return os.path.join(dir, name)
```
MarkZH, line 303: The repetition of `os.path.join()` can be removed by putting `path = os.path.join(dir, name)` above `if platform == "win32":`.
```python
        path = os.path.join(dir, name)
        if platform == "win32":
            path += ".bat"
        else:
            st = os.stat(path)
            os.chmod(path, st.st_mode | stat.S_IEXEC)
        return path
```

PR #1035, opened 2024-10-12
Delay engine and book downloads until needed for a test

MarkZH:
## Type of pull request:
- [ ] Bug fix
- [ ] Feature
- [x] Other

## Description:

This PR delays downloading the chess engines used for testing until the test that uses them actually runs. This way, the test script can start faster and run the downloads in parallel. Plus, any tooling that reads the test file (VS Code plugins that track test completion) does not have to wait for engine/book downloads.

Another change is to skip a test when an engine or book download fails instead of failing it. The [Sjeng website](https://sjeng.org/) seems to have been down for several weeks now. GitHub actions still works because it's retained a cached copy.

Finally, a logger instance local to test_bot.py is used instead of importing the one from lichess_bot.py.

## Related Issues:

N/A

## Checklist:

N/A

### DIFF
--- a/test_bot/test_bot.py
+++ b/test_bot/test_bot.py
@@ -205,8 +205,7 @@ def run_bot(raw_config: CONFIG_DICT_TYPE, logging_level: int, opponent_path: str
 def test_sf() -> None:
     """Test lichess-bot with Stockfish (UCI)."""
     if platform != "linux" and platform != "win32":
-        assert True
-        return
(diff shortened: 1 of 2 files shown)


AttackingOrDefending:
I think we should give tests around 15 more seconds to finish to account for the extra time taken to download the engines.

AttackingOrDefending (approved):

PR #1051, opened 2024-12-02
Add tests for `lib.lichess` and `lib.model`

AttackingOrDefending:
## Type of pull request:
- [ ] Bug fix
- [ ] Feature
- [x] Other

## Description:

Adds more tests. For testing lichess I use [`BadSunfish`](https://lichess.org/@/BadSunfish) which is a bot that I created 3 years but didn't end up using.

Note: The `lichess` tests are skipped when the token is missing, which should happen on all PRs as they don't have access to environment variables.

## Related Issues:

#920 
More tests can still be added so I won't close it yet.

## Checklist:

### DIFF
--- a/test_bot/test_lichess.py
+++ b/test_bot/test_lichess.py
@@ -1,10 +1,13 @@
+"""Tests for the lichess communication."""
+
 from lib import lichess
 import logging
 import os
 import pytest
 
 
-def test_lichess():
+def test_lichess() -> None:
+    """Test the lichess communication."""
     token = os.getenv("LICHESS_BOT_TEST_TOKEN")
     if token is None:
         pytest.skip("Lichess-bot test token must be set.")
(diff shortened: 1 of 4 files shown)


test_bot/test_lichess.py
```diff
@@ -0,0 +1,49 @@
...
+import pytest
+
+
+def test_lichess() -> None:
+    """Test the lichess communication."""
+    token = os.getenv("LICHESS_BOT_TEST_TOKEN")
+    if token is None:
+        pytest.skip("Lichess-bot test token must be set.")
+    li = lichess.Lichess(token, "https://lichess.org/", "0.0.0", logging.DEBUG, 3)
+    assert len(li.get_online_bots()) > 20
```
MarkZH, line 15: Why 20? While there are usually a large number of bots online at any given time, I don't think we should be testing something that's out of our control. Maybe a test like this would be more reliable:
```python
online_bots = li.get_online_bots()
assert isinstance(list, online_bots)
assert any(filter(lambda bot: bot['username'] == "BadSunfish", online_bots))
```

AttackingOrDefending: A problem is that it takes some time until bots show up in the `online_bots` list, so `assert any(filter(lambda bot: bot['username'] == "BadSunfish", online_bots))` will fail most of the times. It is quite hard to create a meaningful test that is completely under our control.

MarkZH (approved):

PR #1063, opened 2024-12-30
Replace flake8/flake8-docstrings with ruff

MarkZH:
## Type of pull request:
- [ ] Bug fix
- [ ] Feature
- [x] Other

## Description:

Flake8-docstrings is deprecated because pydocstyle
(https://github.com/pycqa/pydocstyle) is deprecated. The
checker/linter/formatter ruff is recommended in its place.

## Related Issues:

N/A

## Checklist:

N/A

### DIFF
--- a/.github/workflows/python-build.yml
+++ b/.github/workflows/python-build.yml
@@ -32,10 +32,3 @@ jobs:
       run: |
         # Check for python syntax errors and inconsistent code style.
         ruff check --config test_bot/ruff.toml
-    - name: Lint with flake8-markdown
-      run: |
-        flake8-markdown "*.md"
-        flake8-markdown "wiki/*.md"
-    - name: Lint with flake8-docstrings
-      run: |
-        flake8 . --count --max-line-length=127 --statistics --select=D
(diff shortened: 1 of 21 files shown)


MarkZH:
We'll need to check whether everything that was covered by flake8 is still covered by ruff. Rule sets are listed [here](https://docs.astral.sh/ruff/rules/).

AttackingOrDefending:
When I linted with `--select ALL` I saw that it has some nice additional warnings like ` Q000 [*] Single quotes found but double quotes preferred`, so I would suggest enabling all rules and just disabling a few if we want (like `PLR0913 Too many arguments in function definition (6 > 5)`, `BLE001 Do not catch blind exception: Exception` and `FBT001 Boolean-typed positional argument in function definition`).

AttackingOrDefending:
Since we replaced `flake8` with `ruff` I think we should also remove `flake8-markdown`. It doesn't matter if `ruff` doesn't have an alternative for this as I don't think we have any python code on out markdown files.

AttackingOrDefending (approved):

PR #1066, opened 2025-01-04
Use a trivial engine for tests

MarkZH:
## Type of pull request:
- [x] Bug fix
- [ ] Feature
- [x] Other

## Description:

Instead of a hobbled stockfish engine for game tests, use a trivial engine that just picks the first legal move on a board. This should avoid the random timeouts in tests when games go on too long due to the hobbled test engine still making reasonable moves. Plus, it should cut down the amount of logging done during tests.

## Related Issues:

N/A

## Checklist:

N/A

### DIFF
--- a/test_bot/test_bot.py
+++ b/test_bot/test_bot.py
@@ -28,11 +28,11 @@ from lib import lichess_bot
 platform = sys.platform
 archive_ext = "zip" if platform == "win32" else "tar"
 file_extension = ".exe" if platform == "win32" else ""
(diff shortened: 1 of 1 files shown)


MarkZH:
I'm not actually sure if this will solve the timeout problem since it happens so seldomly. But the only thing I can think of that would lead to a timeout error is the test game going too long. The various engines are cached, so there's no time taken to download them.

AttackingOrDefending (approved):

AttackingOrDefending:
I think it's worth the try if this will solve the random timeouts.

PR #1069, opened 2025-01-06
More efficient move calculation for Gaviota and Syzygy

adbar:
## Type of pull request:
- [ ] Bug fix
- [ ] Feature
- [x] Other

## Description:

This PR simplifies the calculation of moves using Gaviota and Syzygy databases, I noticed inefficient lines like values computed twice. The external tests currently don't pass due to a network issue but the rest is fine. This code should run more efficiently and be easier to maintain.

## Checklist:

### DIFF
--- a/lib/engine_wrapper.py
+++ b/lib/engine_wrapper.py
@@ -23,6 +23,7 @@ from lib.types import (ReadableType, ChessDBMoveType, LichessEGTBMoveType, OPTIO
                        COMMANDS_TYPE, MOVE, InfoStrDict, InfoDictKeys, InfoDictValue, GO_COMMANDS_TYPE, EGTPATH_TYPE,
                        ENGINE_INPUT_ARGS_TYPE, ENGINE_INPUT_KWARGS_TYPE)
 from extra_game_handlers import game_specific_options
+from operator import itemgetter
 from typing import Any, Optional, Union, Literal, cast
 from types import TracebackType
 LICHESS_TYPE = Union[lichess.Lichess, test_bot.lichess.Lichess]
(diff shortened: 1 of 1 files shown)


lib/engine_wrapper.py
```diff
@@ -1149,28 +1150,30 @@ def get_syzygy(board: chess.Board, game: model.Game,
...
             tablebase.add_directory(path)
 
         try:
             moves = score_syzygy_moves(board, dtz_scorer, tablebase)
+            converted_moves = {move: dtz_to_wdl(dtm) for move, dtm in moves.items()}
 
-            best_wdl = max(map(dtz_to_wdl, moves.values()))
-            good_moves = [(move, dtz) for move, dtz in moves.items() if dtz_to_wdl(dtz) == best_wdl]
+            best_wdl = max(converted_moves.values())
+            good_moves = [(move, dtz) for move, dtz in converted_moves.items() if dtz == best_wdl]
```
AttackingOrDefending, line 1166: `converted_moves` contains pairs of `(move, wdl)` so `good_moves` contains pairs of `(move, wdl)` instead of `(move, dtz)` leading to the bot choosing a suboptimal move.

lib/engine_wrapper.py
```diff
@@ -1222,25 +1225,29 @@ def get_gaviota(board: chess.Board, game: model.Game,
...
 
         try:
             moves = score_gaviota_moves(board, dtm_scorer, tablebase)
 
-            best_wdl = max(map(dtm_to_gaviota_wdl, moves.values()))
-            good_moves = [(move, dtm) for move, dtm in moves.items() if dtm_to_gaviota_wdl(dtm) == best_wdl]
-            best_dtm = min([dtm for move, dtm in good_moves])
+            converted_moves = {move: dtm_to_gaviota_wdl(dtm) for move, dtm in moves.items()}
+            best_wdl = max(converted_moves.values())
+            good_moves = [(move, dtm) for move, dtm in converted_moves.items() if dtm == best_wdl]
```
AttackingOrDefending, line 1249: Same problem here. good_moves contains pairs of (move, estimated wdl) instead of (move, dtm).

lib/engine_wrapper.py
adbar: I made this change because I noticed that `dtz_to_wdl` is applied twice, first to compute the maximum value and second to compute again the value of each move. Since nothing else changes it seems unnecessary.

`converted_moves` is actually made of entries of the type `move: dtz_to_wdl(dtm)`, so the second line should be `(move, wdl) for move, wdl in converted_moves.items() …`

Would that solve the problem or am I wrong?

AttackingOrDefending: While the variables have incorrect names, renaming them won't help. If you want to avoid recalculation you will have to change `converted_moves = {move: dtz_to_wdl(dtm) for move, dtm in moves.items()}` to `converted_moves = {move: (dtz, dtz_to_wdl(dtz)) for move, dtz in moves.items()}` to store both the dtz and the wdl, and then choose according to what's needed.

adbar: Mypy raises errors if I use your suggestion. I'm not sure what to do, I could remove the lines we don't agree on so that we can move on with the PR and tackle the issue in another one. You can also directly edit the PR if you want.

AttackingOrDefending: I agree with removing them for now. The change from copying and pushing to pushing and popping seems to be 20-30% faster, so it would be nice to have it.

AttackingOrDefending:
Also, can you update the branch to include the commits that are now in main, so that the tests won't fail?

adbar:
@AttackingOrDefending I just removed the lines you mentioned and integrated the changes from the master branch. The speedup with pushing and popping is nice to hear!

AttackingOrDefending (approved):

MarkZH (approved):

PR #1075, opened 2025-01-10
timer: import submodules only and add tests

adbar:
## Type of pull request:
- [ ] Bug fix
- [ ] Feature
- [x] Other

## Description:

This PR updates the timer class to make the code more concise and adds tests.

## Related Issues:

Addresses #920.

## Checklist:

- [x] I have read and followed the [contribution guidelines](/docs/CONTRIBUTING.md).
- [x] I have added necessary documentation (if applicable).
- [x] The changes pass all existing tests.

PR #1079, opened 2025-01-13
add slots in all classes

adbar:
## Type of pull request:
- [ ] Bug fix
- [ ] Feature
- [x] Other

## Description:

Since we're at it I thought we could add slots in all classes (the TypedDict classes are not concerned).

## Related Issues:

See discussion in PR #1075.

## Checklist:

### DIFF
--- a/lib/config.py
+++ b/lib/config.py
@@ -14,6 +14,8 @@ logger = logging.getLogger(__name__)
 class Configuration:
     """The config or a sub-config that the bot uses."""
 
+    __slots__ = ("config",)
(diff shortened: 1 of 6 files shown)


AttackingOrDefending:
I would also prefer not adding `__slots__`, as they make the code less readable in my opinion. #1069 was merged as it made some readability changes (e.g. changing `elif` to `if`) and improved the speed without loss in the readability of that code.

adbar:
OK, I understand.

PR #1082, opened 2025-01-29
Fix some ruff errors and simplify LICHESS_TYPE

AttackingOrDefending:
## Type of pull request:
- [ ] Bug fix
- [ ] Feature
- [x] Other

## Description:

Fixes some ruff errors and removes them from being ignored. I'm not sure if some changes (like removing `else` after `return`) are beneficial, so I'm open to reverting them back.

Also, makes `Lichess` in `test_bot.lichess` inherit from `lib.lichess.Lichess` to simplify away the `LICHESS_TYPE` variable.

## Related Issues:

None.

## Checklist:

### DIFF
--- a/lib/engine_wrapper.py
+++ b/lib/engine_wrapper.py
@@ -13,7 +13,6 @@ import time
 import random
 import math
 import contextlib
-import test_bot.lichess
 from collections import Counter
 from collections.abc import Callable
 from lib import model, lichess
@@ -26,7 +25,6 @@ from extra_game_handlers import game_specific_options
 from operator import itemgetter
 from typing import Any, Optional, Union, Literal, cast
 from types import TracebackType
-LICHESS_TYPE = Union[lichess.Lichess, test_bot.lichess.Lichess]
 
 
 logger = logging.getLogger(__name__)
@@ -132,7 +130,7 @@ class EngineWrapper:
     def play_move(self,
                   board: chess.Board,
                   game: model.Game,
-                  li: LICHESS_TYPE,
+                  li: lichess.Lichess,
                   setup_timer: Timer,
                   move_overhead: datetime.timedelta,
                   can_ponder: bool,
@@ -193,8 +191,7 @@ class EngineWrapper:
                     game_ender = li.abort if game.is_abortable() else li.resign
                     game_ender(game.id)
                     return
-                else:
-                    raise
+                raise
 
         # Heed min_time
         elapsed = setup_timer.time_since_reset()
@@ -283,8 +280,7 @@ class EngineWrapper:
         """
         if self.comment_start_index < 0:
             return -1
-        else:
-            return move_stack_index - self.comment_start_index
+        return move_stack_index - self.comment_start_index
 
     def comment_for_board_index(self, index: int) -> InfoStrDict:
         """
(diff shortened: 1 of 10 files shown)


lib/lichess.py
```diff
@@ -2,7 +2,7 @@
 import json
 import requests
 from urllib.parse import urljoin
-from requests.exceptions import ConnectionError, HTTPError, ReadTimeout
+from requests.exceptions import ConnectionError as RequestsConnectionError, HTTPError, ReadTimeout
```
MarkZH, line 5: Why is this renamed?

AttackingOrDefending: Because of the error `"A004" # Import is shadowing a Python builtin`. Python has a `ConnectionError` itself which can be used without being imported, so by importing the new error from `requests` we overrode it.

MarkZH:
All of the changes are good, except that I agree with your hesitation about changing the if-elif-else structures. Different ways of writing them can have extra meaning to people, even if the Python interpreter treats them the same.

---

 (Apologies for the pedantry)

TL;DR: I suggest putting `RET505` and `RET508` back in the ignore list, but also reviewing the if-blocks it flags for consideration.

---

One thing ruff seems to really like is turning every `if-elif-else` that has return statements into [guard patterns](https://en.wikipedia.org/wiki/Guard_(computer_science)). For example, the following functions are all equivalent:
```python
def average1(numbers: list[float]) -> float:
    if numbers:
        s = sum(numbers)
        n = len(numbers)
        return s/n
    else:
        raise ValueError("Cannot find average of empty list.")

def average2(numbers: list[float]) -> float:
    if numbers:
        s = sum(numbers)
        n = len(numbers)
        return s/n

    raise ValueError("Cannot find average of empty list.")

def average3(numbers: list[float]) -> float:
    if not numbers:
        raise ValueError("Cannot find average of empty list.")

    s = sum(numbers)
    n = len(numbers)
    return s/n
```
Statements in a sequence are the easiest to understand since they don't have branches or loops. So, we should place the most important parts of a function within the fewest control structures (minimizing indentation).

The first function, `average1`, puts the normal path on the same footing as the error path, which doesn't seem right since the normal path is way more common.

The second function, `aveage2` (that ruff prefers), makes it look like the normal path that calculates the average is the special case (this is the reverse of the guard pattern).

The third function, `average3`, has the correct form of the guard pattern and is the one I would choose.

In this PR, I would agree with the change to the if-block starting on line 283 of `lib/engine_wrapper.py` (`if self.comment_start_index < 0:`). The `return -1` is basically an error condition, so this is a correct guard clause pattern.

One change I don't agree with is this (starting on line 352 of `lib/engine_wrapper`):
```python
        minutes, seconds = divmod(number, 60)
        if minutes >= 1:
            return f"{minutes:0.0f}m {seconds:0.1f}s"
        else:
            return f"{seconds:0.1f}s"
```
to
```python
        minutes, seconds = divmod(number, 60)
        if minutes >= 1:
            return f"{minutes:0.0f}m {seconds:0.1f}s"
        return f"{seconds:0.1f}s"
```
Formatting a time duration of more than a minute is not a special case. These two cases are just different ways of handling different data and so should not be put in different structure.

There is one example of an if-block that ruff thinks should be changed, but I disagree with the ruff-suggested change: lines 188 to 194 of `lib/engine_wrapper.py`: the `raise` command should be put in the if-block with the condition reversed. The handling of a bad move could then be unindented.

The changes like
```python
if a:
    return a_result
elif b:
    return b_result
else:
    return c_result
```
to
```python
if a:
    return a_result
if b:
    return b_result
return c_result
```
are fine.

Use your judgement when choosing structures.

AttackingOrDefending:
I changed the structures to what I considered most readable.

lib/engine_wrapper.py
```diff
@@ -193,8 +191,7 @@ def play_move(self,
                     game_ender = li.abort if game.is_abortable() else li.resign
                     game_ender(game.id)
                     return
-                else:
-                    raise
+                raise
```
MarkZH, line 194: The if-condition should be reversed to match the guard pattern:
```python
if not any(isinstance(e, BadMove) for e in error.args):
    raise

# Logging bad move and returning
```

PR #1090, opened 2025-02-19
Restart if there is a network exception

### LINKED ISSUE #1088: Soft-fail on connection drop
Just a thought, maybe when the lichess-bot script is unable to reach the lichess servers, it could enter a sleep loop where it keeps checking (fail, sleep 10s, check again, etc.) instead of exiting. I think this could be a nice feature to improve stability of bots, just a little idea.

Thanks!

AttackingOrDefending:
## Type of pull request:
- [ ] Bug fix
- [x] Feature
- [ ] Other

## Description:

Restart lichess-bot if an exception is raised due to network errors. This also makes the bot continue ruuning is the WiFi stops working for some time.

## Related Issues:

closes #1088 

## Checklist:

### DIFF
--- a/lib/lichess_bot.py
+++ b/lib/lichess_bot.py
@@ -1256,6 +1256,9 @@ def start_program() -> None:
             except RequestException:
                 restart = True
                 logger.exception("Restarting lichess-bot due to a network error:")
+            if terminated or force_quit:
+                logger.info("Termination requested - stopping restart cycle")
+                break
             time.sleep(10 if should_restart() else 0)
     except Exception:
         logger.exception("Quitting lichess-bot due to an error:")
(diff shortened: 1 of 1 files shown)


AttackingOrDefending:
The first network error is caught by [this](https://github.com/AttackingOrDefending/lichess-bot/blob/connection/lib/lichess_bot.py#L387) which sets restart to True. After lichess-bot restarts it will fail on the token test and then `Restarting lichess-bot due to a network error:` will show up. lichess-bot currently catches the first network error but terminates when getting the second network error on token test.

I just found out that with this PR, lichess-bot won't be able to terminate if offline, so that should be fixed. It's quite possible that token test was added after the restarting functionality which is why it wasn't taken care of.

AttackingOrDefending:
The problem with terminating is now fixed, but it still takes some time until the network timeout is raised but this issue also exists in the master branch as if you start lichess-bot while offline and try to press Ctrl+C while the token test is happening, you will still have to wait until it finishes for lichess-bot to terminate.

MarkZH (approved):

MarkZH:
Since this PR improves on master, I'm approving it. But, the fact that it takes a long time to break out of the token test loop when there is no network connection is probably worth a fix--probably by making a custom `api_post()` method that bails out earlier with a smaller `max_time` argument in the `backoff` decoration.

PR #1093, opened 2025-03-14
Exit lichess-bot promptly without waiting

AttackingOrDefending:
## Type of pull request:
- [x] Bug fix
- [ ] Feature
- [ ] Other

## Description:

lichess-bot now exits immediately when receiving a SIGINT without waiting for 60 seconds. Now the maximum time wait is 15 seconds because of the timeout in `get_game_stream` and `get_event_stream`. That can also be changed.

I opted to create a class `Stop` and make all functions use that class so the variables are shared.

Also fixes the ruff errors.

## Related Issues:

#1090 

## Checklist:

### DIFF
--- a/lib/engine_wrapper.py
+++ b/lib/engine_wrapper.py
@@ -94,7 +94,7 @@ class EngineWrapper:
         self.engine: Union[chess.engine.SimpleEngine, FillerEngine]
         self.scores: list[chess.engine.PovScore] = []
         self.draw_or_resign = draw_or_resign
-        self.go_commands = Configuration(cast("GO_COMMANDS_TYPE", options.pop("go_commands", {})) or {})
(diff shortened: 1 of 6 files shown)


MarkZH:
Can you move the ruff error fixes to another PR? It would make it easier to evaluate the changes. If that causes errors, you can add a rule exception to the ruff.toml file.

MarkZH:
Actually, adding the type quoting changes to #1094 would make more sense.

Do you agree that quoting types in `cast()` is beneficial, or should we disable `TC006`?

AttackingOrDefending:
I think it would be better to add an exception as the code looks less clean in my opinion after adding the quotes.

MarkZH (approved):

Issue #1107, opened 2025-05-10
Different openings for white and black


CallOn84:
Hi developers.

I was wondering if it's possible to have different opening books for the white and black sides?

MarkZH:
How would these books be different? Wouldn't any opening book have the best moves for both sides?

CallOn84:
> How would these books be different? Wouldn't any opening book have the best moves for both sides?

This is just my experience using the polyglot book feature. I don't want to set the move pick to be the best moves, as doing so would mean that there's not enough variety when the bot is playing. However, the `weight_random` and `min_weight` don't always work particularly well because the weights vary depending on the position hash. Furthermore, the best move in an opening doesn't necessarily mean the bot prefers it.

AttackingOrDefending:
Would adding a normalization option to the opening books section help? For now, I'm thinking something like:
- No normalization (current)
- Make the sum of all weights equal to 100 (so `new_weight = weight / sum(weights) * 100`)
- Make the largest weight equal to 100 (so `new_weight = weight / max(weights) * 100`)

This should make using the min_weight option easier.

CallOn84:
That might be good. I can test it with Elite Leela if you want to see how it performs

CallOn84:
> Let us know how the latest change works for you.

How does it work config.yml wise?

PR #1109, opened 2025-05-15
Normalize book weights

AttackingOrDefending:
## Type of pull request:
- [ ] Bug fix
- [x] Feature
- [ ] Other

## Description:

Adds the ability to normalize the weights of an opening book, so that `min_weight` can be used more easily across different books.

## Related Issues:

#1107 

## Checklist:


MarkZH:
I don't understand how this PR changes anything. Scaling the weights shouldn't matter, since all that matters is their relative sizes. Multiplying every weight by the same number should result in the same probability.

Also, it looks like the new random choices in the `get_move_from_entries()` function behave the same as the [`find()`](https://python-chess.readthedocs.io/en/latest/polyglot.html#chess.polyglot.MemoryMappedReader.find) (`best_move`), [`choice()`](https://python-chess.readthedocs.io/en/latest/_modules/chess/polyglot.html#MemoryMappedReader.choice) (`uniform_random`), and [`weighted_choice()`](https://python-chess.readthedocs.io/en/latest/_modules/chess/polyglot.html#MemoryMappedReader.weighted_choice) (`weighted_random`) methods of the python-chess polyglot reader.

After reading the linked issue, are these changes about making `min_weight` have a consistent meaning? In that case, it would be easier to scale `min_weight`. Something like this:
```python
min_weight = polyglot_cfg.min_weight
normalization = polyglot_cfg.normalization
entries = reader.find_all(board)
scalar = (100 if normalization == "none"
          else sum(entry.weight for entry in entries) if normalization == "sum"
          else max(entry.weight for entry in entries) if normalization == "max")
min_weight = min_weight*scalar/100
# I think this is equivalent to the change in the PR.

# The rest of the function is unchanged.

if selection == "weighted_random":
    move = reader.weighted_choice(board).move
elif selection == "uniform_random":
    move = reader.choice(board, minimum_weight=min_weight).move
elif selection == "best_move":
    move = reader.find(board, minimum_weight=min_weight).move
```

AttackingOrDefending:
The goal was to normalize the weights of an opening book, as they can be in the range of 0-65535, and even in the same opening book some positions may have a sum of weights `<= 10` and others `> 100000`. This makes `min_weight` hard to use as a value of 20 wouldn't leave any move in the first position but would leave all the moves in the second. Indeed, scaling `min_weight` instead of the weights of the entries seems to be an easier way to accomplish the same goal.

MarkZH:
Is it possible that the maximum score will be negative? If so, the `scalar` should be calculated with the absolute value. Otherwise, it would change the sign of the `min_weight`.

The documentation for `min_weight` should have a reference to the `normalization` section.

How are checkmates handled in books? If they have an infinite score, then do the book functions correctly handle that?

AttackingOrDefending:
The weights in polyglot opening books are unsigned 16-bit integers so they take values from 0 to 65535.

lib/engine_wrapper.py
```diff
@@ -754,6 +754,12 @@ def get_book_move(board: chess.Board, game: model.Game,
             try:
                 selection = polyglot_cfg.selection
                 min_weight = polyglot_cfg.min_weight
+                normalization = polyglot_cfg.normalization
+                entries = list(reader.find_all(board))
```
MarkZH, line 758: To reduce redundancy, this line can be changed to `weights = [entry.weight for entry in reader.find_all(board)]`. Then the next line becomes shorter.

PR #1136, opened 2025-09-25
Fix lichess API test

AttackingOrDefending:
## Type of pull request:
- [x] Bug fix
- [ ] Feature
- [ ] Other

## Description:

There were changes to the lichess api which made a test fail.

## Related Issues:

None

## Checklist:


MarkZH:
I think it would be better to write a pair of very simple engines for testing--one for UCI, one for Xboard. This way, we would be relying on these outside engines working in a specific way.

The test engines would be similar to the [buggy engine](https://github.com/lichess-bot-devs/lichess-bot/blob/master/test_bot/buggy_engine.py) in that they would be very simple by playing the same game every time (no searching). They would only implement the bare bones of the two protocols. Since we would have complete control over these engines, we could use them for testing things like 
- sending options to the engine from configuration files, and
- checking various log messages reliably like for white winning, black winning, and draws.

MarkZH (approved):

PR #1137, opened 2025-09-26
Handle new error message when challenging bots

### LINKED ISSUE #1135: New lichess ratelimit applies
I just added [a new limit for lichess bots](https://github.com/lichess-org/lila/blob/master/modules/bot/src/main/BotLimit.scala). They can now play a maximum of 100 games against other bots per day.
Games against humans remain unlimited.

Why the change? To reduce wasted resources due to bots playing several hundreds of games against other bots, every day, using automatic match-making. No-one ever looks at these games. Several bots have hundreds of thousands of such games, each.

I understand that playing other bots is useful to test an engine, so it's still possible, up to 100 games per day. I think it's largely sufficient - in fact that limit might be changed to 64 or 50, at some point.

So, when creating a challenge against another bot, it's now possible to receive a `429` response with a body like this:

```json
{
    "error": "You played 100 games against other bots today, please wait before challenging another bot.",
    "ratelimit": {
        "key": "bot.vsBot.day",
        "seconds": 86398
    }
}
```
You can use `ratelimit.seconds` to know when to resume bot challenges, and `ratelimit.key` to know what's going on without parsing the message, and display your own message.

You may also receive a `400` response if the bot you are challenging has played too many bot games:

```json
{
  "error": "<Opponent-name> played 100 games against other bots today, please wait until <UTC-date> to challenge them."
}
```

This means you can still challenge other bots - just not that one.

I think it would be helpful for lichess-bot to understand this new limit and prevent its users from being ratelimited, and possibly IP banned.

Let me know if you have any questions, here or on https://discord/gg/lichess

MarkZH:
Detect when the rate limiting status code (429) is due to issuing a challenge after playing more than 100 games in a day. Use the message in the response to delay issuing another challenge.

## Type of pull request:
- [ ] Bug fix
- [ ] Feature
- [x] Other

## Description:

The biggest change is setting the time delay according to the `["ratelimit"]["seconds"]` field in the response.

## Related Issues:

Closes #1135

## Checklist:

### DIFF
--- a/lib/lichess.py
+++ b/lib/lichess.py
@@ -61,6 +61,11 @@ stop = Stop()
 class RateLimitedError(RuntimeError):
     """Exception raised when we are rate limited (status code 429)."""
 
+    def __init__(self, message: str, timeout: datetime.timedelta) -> None:
+        """Create a rate-limited error with the time left until the rate limit expires."""
+        super().__init__(message)
+        self.timeout = timeout
+
 
 def is_new_rate_limit(response: requests.models.Response) -> bool:
     """Check if the status code is 429, which means that we are rate limited."""
@@ -263,7 +268,8 @@ class Lichess:
         path_template = ENDPOINTS[endpoint_name]
         if self.is_rate_limited(path_template):
             raise RateLimitedError(f"{path_template} is rate-limited. "
-                                   f"Will retry in {sec_str(self.rate_limit_time_left(path_template))} seconds.")
+                                   f"Will retry in {sec_str(self.rate_limit_time_left(path_template))} seconds.",
+                                   self.rate_limit_time_left(path_template))
         return path_template
 
     def set_rate_limit_delay(self, path_template: str, delay_time: datetime.timedelta) -> None:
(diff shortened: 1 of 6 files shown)


MarkZH:
I'm submitting this as a draft since I have more changes planned. Plus, I want to make sure that I'm on the right track in implementing the change to the bot API.

My next change is probably getting rid of `Matchmaking.min_wait_time` and the increasing delay calculations in `Matchmaking. update_daily_challenge_record()`. If the user wants to use all of their daily challenges in a short time, they should be able to since the new error message is very explicit.

MarkZH:
Next: add a message for short `matchmaking: challenge_timeout` configurations to warn users that they will run out of daily games quickly.

AttackingOrDefending (approved):

ornicar:
> I don't think players will care if their 100 games/day limit is used up quickly or not.

Maybe they could care. Imagine you just made a tweak to your bot, and want to see it in action, but all 100 games were played because lichess-bot rushes them asap

PR #1141, opened 2025-10-05
Use Timers instead of bools to filter opponents to allow temporary stops to challenges

MarkZH:
## Type of pull request:
- [ ] Bug fix
- [x] Feature
- [ ] Other

## Description:

Instead of preventing the challenging of another bot that declines a challenge for the rest of the lichess-bot run, default to a one-day timeout. Block lists still work, but are implemented as a 10-year timeout.

The primary reason for this PR is an upcoming change in lichess. Challenging another bot that has already played 100 games today will result in a 400 status code and an error message with information about when the bot's timeout will expire.

## Related Issues:

This commit to lila tells bots how long to wait before challenging another bot that has reached the 100-bot-game per day limit: https://github.com/lichess-org/lila/commit/53f2c4d2f321da1d4cd05b7c30d417dd1e902a49

## Checklist:

### DIFF
--- a/lib/lichess.py
+++ b/lib/lichess.py
@@ -268,9 +268,9 @@ class Lichess:
         """
         path_template = ENDPOINTS[endpoint_name]
         if self.is_rate_limited(path_template):
-            raise RateLimitedError(f"{path_template} is rate-limited. "
-                                   f"Will retry in {sec_str(self.rate_limit_time_left(path_template))} seconds.",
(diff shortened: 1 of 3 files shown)


MarkZH:
I think this is ready for testing beyond my computer. There is a bit of redundancy. The player timeout after 100 games is implemented in both the `Lichess` class and the `Matchmaking` class. This is probably unavoidable given their different functions.

Issue #1158, opened 2025-11-27
syzygy usage by lichess-bot plays wrong losing moves


abulmo:
**Describe the bug**
In my config.yml file I use syzygy endgame table through lichess-bot interface like this:

_lichess_bot_tbs: # The tablebases list here will be read by lichess-bot, not the engine.
    syzygy:
      enabled: true # Whether or not to use local syzygy endgame tablebases.
      paths: # Paths to Syzygy endgame tablebases.
        - "/home/richard/progs/chess/tb"
      max_pieces: 6 # Maximum number of pieces in the endgame tablebase.
      move_quality: "best" # One of "best" or "suggest" (it takes all the moves with the same WDL and tells the engine to only consider these; will move instantly if there is only 1 "good" move)._

In the following game: [https://lichess.org/pj5eG32b/black], My program Dumb ( dumb_by_abulmo-bot on Lichess), reached a winning position into the endgame table after 70...Rxb3. But lichess-bot then played the next move by itself 71...b5 leading to a draw.

**Expected behavior**
It should have played 71...Rb2+ or 71...Rb4 (the two winning moves with the lowest dtz) instead of 71...b5. 

**Logs**
screen output from lichess_bot
[terminal-output.txt](https://github.com/user-attachments/files/23787024/terminal-output.txt)
[config.log](https://github.com/user-attachments/files/23786817/config.log)
last 1000 lines of lichess.log
[lichess-bot.log](https://github.com/user-attachments/files/23786819/lichess-bot.log)
uci communication & Search log from my Engine
[Dumb-206365.log](https://github.com/user-attachments/files/23786840/Dumb-206365.log)

**Desktop (please complete the following information):**
 - OS:Fedora Linux 43
 - Python Version: 3.14
 - Terminal: bash

**Additional context**
I checked  than my KRPPvKB.rtw and KRPPvKB.rtz files are correct.

PR #1161, opened 2025-12-03
Fix syzygy

AttackingOrDefending:
## Type of pull request:
- [x] Bug fix
- [ ] Feature
- [ ] Other

## Description:

The 0.5 we would add for zeroing moves so that we would count 0.5 as theit dtz instead of the dtz after the move was being added to drawn positions as well.

## Related Issues:

#1158 

## Checklist:


AttackingOrDefending:
For 1210, I made the required change to 1209. For dtm, I suggest we keep it for the 50-move-rule although we could remove the entire ` + int(math.copysign(board.halfmove_clock, dtm) if dtm else 0)`.

MarkZH (approved):

PR #1164, opened 2025-12-12
Assume game is active until lichess says otherwise

### LINKED ISSUE #1163: Bot exceeds `concurrency = 1` and accepts multiple simultaneous challenges
Even with concurrency = 1 set in the configuration, the bot still accepts multiple challenges at the same time. This seems to be a concurrency-limit bug in lichess-bot-bridge.

When several bots challenge almost simultaneously, the bot accepts 2–3 challenges concurrently. In the screenshot example, it accepted games from SybBot, FataliiBot, and turkjs while it was already playing one game. Since my hardware cannot run multiple engine instances, two of those games were lost on time.

Image

This issue happens frequently:

The bot starts a single game.

Multiple new challenges arrive close together.

concurrency = 1 is ignored and all challenges are accepted.

Engine overloads and crashes.

Bot cannot reconnect to the earlier games and loses them on time.

Configuration:
concurrency = 1

Expected behavior:
When running with concurrency = 1, every incoming challenge beyond the first should be ignored or declined until the current game finishes.

Actual behavior:
If challenges arrive within a tight time window, the bot accepts all of them, bypassing the concurrency limit.

This is causing repeated timeouts and rating loss. A strict enforcement of the concurrency limit is urgently needed, especially for users running on limited hardware.

MarkZH:
## Type of pull request:
- [x] Bug fix
- [ ] Feature
- [ ] Other

## Description:

When a communication error occurs during a game, the function `game_is_active()` checks with lichess if the current game is still ongoing before trying to reconnect. If that function errors, then the game is aborted locally and never reconnects.

With this change, a failure to retrieve the list of active games results in the function returning `True`.

## Related Issues:

Closes #1163

## Checklist:

### DIFF
--- a/lib/lichess.py
+++ b/lib/lichess.py
@@ -401,13 +401,16 @@ class Lichess:
         self.set_user_agent(profile["username"])
         return profile
 
-    def get_ongoing_games(self) -> list[GameType]:
-        """Get the bot's ongoing games."""
-        ongoing_games: list[GameType] = []
+    def get_ongoing_games(self) -> list[GameType] | None:
+        """
+        Get the bot's ongoing games.
+
+        If an error occurs when retreiving the games, None is returned.
+        """
         with contextlib.suppress(Exception):
             response = cast(dict[str, list[GameType]], self.api_get_json("playing"))
-            ongoing_games = response["nowPlaying"]
-        return ongoing_games
+            return response["nowPlaying"]
+        return None
 
     def resign(self, game_id: str) -> None:
         """Resign a game."""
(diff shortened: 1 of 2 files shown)


MarkZH:
It seems like the error requires an error in sending a move that causes lichess-bot to reach [here](https://github.com/lichess-bot-devs/lichess-bot/blob/62c7237604791d4fa209987c1a3b031f4ac30c2a/lib/lichess_bot.py#L740) and then another error that causes the [active games list](https://github.com/lichess-bot-devs/lichess-bot/blob/62c7237604791d4fa209987c1a3b031f4ac30c2a/lib/lichess.py#L406-L410) to come back empty.

Issue #1200, opened 2026-04-24
watch_control_stream silently hangs on TCP CLOSE-WAIT — bot disappears from /api/bot/online without reconnecting


Etschmia:
## Summary

When the Lichess server closes the event stream connection (sending a TCP FIN),
the bot's socket enters **CLOSE-WAIT** state. `iter_lines()` blocks indefinitely
without raising an exception, so the reconnect loop in `watch_control_stream`
never iterates. The bot process stays alive and appears healthy, but holds no
active stream connection — causing it to vanish from `/api/bot/online` silently.

Related to #151 (which covers the exception path), but this is the silent path:
no `ChunkedEncodingError`, no log output, no restart.

## Root cause

`lib/lichess_bot.py`, `watch_control_stream`:

```python
while not stop.terminated:
    try:
        with li.get_event_stream() as response:
            lines = response.iter_lines()
            for line in lines:   # ← blocks here indefinitely on CLOSE-WAIT
                ...
    except Exception:
        error = traceback.format_exc()
        break                    # ← exception path exits loop → triggers restart
                                 #   but CLOSE-WAIT raises no exception at all

When Lichess sends FIN (e.g. after a game ends, or on a server-side
timeout), the kernel ACKs it and the socket enters CLOSE-WAIT. At the
application layer, requests/urllib3 does not surface this as an
exception — iter_lines() simply never returns the next item. The for
loop stalls, the while loop never re-enters, and no reconnect happens.

The only recovery mechanism is check_online_status (line 358/426), which
fires at most once per hour — far too slow for a bot that is supposed
to be continuously available.

How to observe

# While the bot is idle after a game:
ss -tp | grep python
# Shows: CLOSE-WAIT  ...  python,pid=<main_pid>,fd=4
#        (instead of ESTAB)

# Confirm bot is absent from Lichess bot list:
curl -s "https://lichess.org/api/bot/online?nb=200" | grep -i <botname>
# → no output

Proposed fix

Add a read timeout to the streaming request so that a dead connection
raises ReadTimeout instead of hanging forever. requests supports this
via the timeout parameter as (connect_timeout, read_timeout):

# In li.get_event_stream() — set a read timeout slightly longer than
# Lichess's keepalive interval (~15 s empty-line pings):
response = session.get(url, stream=True, timeout=(5, 30))

With a 30-second read timeout, a CLOSE-WAIT connection raises
ReadTimeout within 30 s, the except Exception branch fires, and the
[...]

PR #1201, opened 2026-04-24
Reconnect control stream on network errors instead of terminating

Etschmia:
## Problem

`watch_control_stream` handles all exceptions with `break`, which exits the reconnect loop, sends `{"type": "terminated"}` to the control queue, and triggers a full bot restart (including a 10-second sleep in the outer restart loop). Any transient network error — including a read timeout on a TCP connection stuck in CLOSE_WAIT — therefore takes the bot offline for 10+ seconds.

Related discussion: #1200

## Fix

On exception, log at WARNING level and `continue` the loop (1-second backoff) instead of `break`. The stream reconnects immediately without interrupting active game processes or requiring a full restart.

Only `stop.terminated` (SIGINT) causes the loop to exit and send `"terminated"`, which is the intended shutdown path.

## Before / After

```python
# before
except Exception:
    error = traceback.format_exc()
    break                            # kills reconnect loop → full restart

# after
except Exception:
    if stop.terminated:
        break
    logger.warning(f"Control stream error, reconnecting:\n{traceback.format_exc()}")
    time.sleep(1)                    # brief backoff, then reconnect
```

## Notes

- `time` and `traceback` are already imported at the top of `lichess_bot.py`
- The `error` field in the `"terminated"` message was only ever populated on the exception path; it is now always `None` since exceptions are handled by reconnecting
- Existing tests pass unchanged since `watch_control_stream` is exercised through the fake Lichess server which closes streams cleanly (no exception path hit)

### DIFF
--- a/lib/lichess_bot.py
+++ b/lib/lichess_bot.py
@@ -125,12 +125,10 @@ def watch_control_stream(control_queue: CONTROL_QUEUE_TYPE, li: lichess.Lichess)
                     else:
                         control_queue.put_nowait({"type": "ping"})
         except Exception:
-            if stop.terminated:
-                break
             logger.warning(f"Control stream error, reconnecting:\n{traceback.format_exc()}")
             time.sleep(1)
 
-    control_queue.put_nowait({"type": "terminated", "error": None})
+    control_queue.put_nowait({"type": "terminated"})
 
 
 def do_correspondence_ping(control_queue: CONTROL_QUEUE_TYPE, period: datetime.timedelta) -> None:
@@ -381,7 +379,6 @@ def lichess_bot_main(li: lichess.Lichess,
 
             if event["type"] == "terminated":
                 stop.restart = True
-                logger.debug(f"Terminating exception:\n{event['error']}")
                 control_queue.task_done()
                 break
 
--- a/lib/lichess_bot.py
+++ b/lib/lichess_bot.py
@@ -126,10 +126,12 @@ def watch_control_stream(control_queue: CONTROL_QUEUE_TYPE, li: lichess.Lichess)
                     else:
                         control_queue.put_nowait({"type": "ping"})
         except Exception:
-            error = traceback.format_exc()
-            break
+            if stop.terminated:
+                break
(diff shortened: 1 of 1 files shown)


AttackingOrDefending:
Can you fix the ruff errors?

MarkZH:
I've let this run for almost three days on my computer, and the exception path was never taken. This PR seems fine, but I don't have a way to test it with actual lichess sessions.

lib/lichess_bot.py
```diff
@@ -126,10 +125,12 @@ def watch_control_stream(control_queue: CONTROL_QUEUE_TYPE, li: lichess.Lichess)
                     else:
                         control_queue.put_nowait({"type": "ping"})
         except Exception:
-            error = traceback.format_exc()
-            break
+            if stop.terminated:
```
MarkZH: This branch seems redundant with the while-loop condition. Can it be removed?

lib/lichess_bot.py
```diff
@@ -126,10 +125,12 @@ def watch_control_stream(control_queue: CONTROL_QUEUE_TYPE, li: lichess.Lichess)
...
+                break
+            logger.warning(f"Control stream error, reconnecting:\n{traceback.format_exc()}")
+            time.sleep(1)
 
-    control_queue.put_nowait({"type": "terminated", "error": error})
+    control_queue.put_nowait({"type": "terminated", "error": None})
```
MarkZH: If the error field is always None, then there's no point in sending it. Remove `, "error": None` and the `logger.debug()` call in line 382.

lib/lichess_bot.py
Etschmia: You're right, removed in 66b5167.

lib/lichess_bot.py
Etschmia: Good catch, removed both in 66b5167.

Etschmia:
Thanks for testing this! On my host the stuck-CLOSE_WAIT exception happened ~once a week — it likely needs flaky network conditions to surface, so a clean test environment won't see it.

MarkZH (approved):

AttackingOrDefending (approved):

PR #1209, opened 2026-05-19
Increase number of bots queried from lichess

### LINKED ISSUE #1208: Increase limit to Lichess online bots API endpoint request

When using matchmaking, lichess-bot uses the following URL to get the list of online bots. According to Lichess API, only the first 100 bots are returned by default.
Considering that most bots will be filtered out due to opponent_min_rating, opponent_max_rating or opponent_rating_difference parameters, this limits the reach for bots.
If there were min/max parameters in the endpoint, the request could use this to get only the desired bots. But the only available parameter is "nb", which sets the number of bots returned. Its default is 100.

So, the only solution right now seems to be increasing "nb", possibly to the max value (512).
Or set a new config.yml parameter to allow us set our desired limit.

Propose to Lichess so the API to be changed, adding min_rating and max_rating parameters.

**Additional context**

curl -H "Authorization: Bearer lip_xxxxxxxxxxxxxxxxxx" https://lichess.org/api/bot/online
Note the 100 lines limit.
curl -H "Authorization: Bearer lip_xxxxxxxxxxxxxxxxxx" https://lichess.org/api/bot/online?nb=512
Note the increased number of  lines.

MarkZH:
## Type of pull request:
- [ ] Bug fix
- [x] Feature
- [ ] Other

## Description:

Increase the number of bots queried from lichess from the default 100 to the maximum 512. This should allow for a wider variety of bots to challenge.

## Related Issues:

Closes #1208

## Checklist:

### DIFF
--- a/lib/lichess.py
+++ b/lib/lichess.py
@@ -431,7 +431,7 @@ class Lichess:
     def get_online_bots(self) -> list[UserProfileType]:
         """Get a list of bots that are online."""
         try:
-            online_bots_str = self.api_get_raw("online_bots")
+            online_bots_str = self.api_get_raw("online_bots", params={"nb": "512"})
             online_bots = list(filter(bool, online_bots_str.split("\n")))
             return list(map(json.loads, online_bots))
         except Exception:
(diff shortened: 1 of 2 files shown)


MarkZH:
I also added logging for the number of bots downloaded and the number of bots after filtering. This should help users to diagnose if their filters are overzealous.

AttackingOrDefending (approved):

PR #1217, opened 2026-06-20
feat: reserve game slots for human players (#1212)

### LINKED ISSUE #1212: New Feature: Reserve game slots for human players
With bots constantly challenging each other, it may be difficult for a human player to challenge a bot if all of it's game slots are constantly occupied.

This will require an extra setting: `challenge: games_reserved_for_humans`. This will be an integer from 0 to `challenge: concurrency` (default: 0) that will specify how many game slots should be left open for humans. If `challenge: concurrency` is 10 and `challenge: games_reserved_for_humans` is 4, then there will be at most 6 concurrent games against bots.

The `challenge: preference: "human"` configuration is not enough since, if there aren't any challenges from humans at the moment, then all of the game slots will be filled up with bot games.

**Additional context**
Idea from Discord message: https://discord.com/channels/280713822073913354/1420771373789352028/1511753570595377424

philipmk42:
## Type of pull request:
- [ ] Bug fix
- [x] Feature
- [ ] Other

## Description:

Implements games_reserved_for_humans (#1212) — a setting that keeps a number of game slots free for human challengers so other bots can't occupy every slot. Following @MarkZH's suggested design: a Player.bot_names class registry (populated in Player.__init__ from challenge/gameFull events, with an is_bot_name() classmethod) lets the code identify a bot by name, working around the gameStart event omitting the title. active_games is now a dict[str, str] (game ID → opponent name) so bot vs. human games can be told apart. accept_challenges gates bot challenges against concurrency - games_reserved_for_humans while still accepting humans up to full concurrency, and Matchmaking.challenge (which only challenges bots) is capped at the same bot limit. Setting documented in config.yml.default and validated in config.py.

## Related Issues:

Closes #1212

## Checklist:

### DIFF
--- a/lib/config.py
+++ b/lib/config.py
@@ -205,6 +205,7 @@ def insert_default_values(CONFIG: CONFIG_DICT_TYPE) -> None:
     set_config_default(CONFIG, "engine", "polyglot", key="min_weight", default=1)
     set_config_default(CONFIG, "engine", "polyglot", key="normalization", default="none")
     set_config_default(CONFIG, "challenge", key="concurrency", default=1)
+    set_config_default(CONFIG, "challenge", key="games_reserved_for_humans", default=0)
     set_config_default(CONFIG, "challenge", key="sort_by", default="best")
     set_config_default(CONFIG, "challenge", key="preference", default="none")
     set_config_default(CONFIG, "challenge", key="accept_bot", default=False)
--- a/lib/config.py
+++ b/lib/config.py
@@ -318,6 +318,15 @@ def validate_config(CONFIG: CONFIG_DICT_TYPE) -> None:
     config_warn(CONFIG["challenge"]["concurrency"] > 0, "With challenge.concurrency set to 0, the bot won't accept or create "
                                                         "any challenges.")
 
+    config_assert(0 <= CONFIG["challenge"]["games_reserved_for_humans"] <= CONFIG["challenge"]["concurrency"],
+                  "challenge.games_reserved_for_humans must be between 0 and challenge.concurrency.")
+
+    config_warn(CONFIG["challenge"]["games_reserved_for_humans"] == 0
+                or CONFIG["challenge"]["preference"] == "human",
+                'challenge.games_reserved_for_humans reserves slots for humans, but challenge.preference is not '
+                '"human", so human challenges may be stalled behind bot challenges in the queue. Set '
+                'challenge.preference to "human" to prioritize human challengers.')
+
     config_assert(CONFIG["challenge"]["sort_by"] in ["best", "first"], "challenge.sort_by can be either `first` or `best`.")
     config_assert(CONFIG["challenge"]["preference"] in ["none", "human", "bot"],
                   "challenge.preference should be `none`, `human`, or `bot`.")
--- a/lib/matchmaking.py
+++ b/lib/matchmaking.py
@@ -229,15 +229,18 @@ class Matchmaking:
         :param challenge_queue: The queue containing the challenges.
         :param max_bot_games: The maximum allowed number of simultaneous games against bots.
         """
-        max_games_for_matchmaking = (max_bot_games if self.matchmaking_cfg.allow_during_games
-                                     else min(1, max_bot_games))
+        # If matchmaking is not allowed while playing other games, don't create
+        # a challenge when any game (against a bot or a human) is in progress.
+        if not self.matchmaking_cfg.allow_during_games and active_games:
+            return
+
         # Count only games against bots: matchmaking only challenges bots, and
         # human games must not count toward the bot-game limit (otherwise a bot
         # whose slots are filled by humans would stop matchmaking even though
         # bot slots remain free).
         bot_game_count = (sum(1 for name in active_games.values() if model.Player.is_bot_name(name))
                           + len(challenge_queue))
-        if (bot_game_count >= max_games_for_matchmaking
+        if (bot_game_count >= max_bot_games
                 or (bot_game_count > 0 and self.last_challenge_created_delay.time_since_reset() < self.max_wait_time)
                 or not self.should_create_challenge()):
             return
--- a/lib/matchmaking.py
+++ b/lib/matchmaking.py
@@ -237,12 +237,8 @@ class Matchmaking:
         # Count only games against bots: matchmaking only challenges bots, and
         # human games must not count toward the bot-game limit (otherwise a bot
         # whose slots are filled by humans would stop matchmaking even though
-        # bot slots remain free). Opponent names may be title-prefixed (e.g.
-        # "BOT sseh-c") while bot_names holds bare usernames, so compare the
-        # bare username (the last whitespace-separated token).
-        bot_game_count = (sum(1 for name in active_games.values()
-                              if model.Player.is_bot_name(name.split()[-1] if name else name))
-                          + len(challenge_queue))
+        # bot slots remain free).
+        bot_game_count = model.Player.count_bot_games(active_games) + len(challenge_queue)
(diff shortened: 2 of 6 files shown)


lib/matchmaking.py
```diff
@@ -220,15 +220,17 @@ def get_random_config_value(self, config: Configuration, parameter: str, choices
...
-        :param active_games: The games that the bot is playing.
+        :param active_games: The games that the bot is playing (game ID -> opponent name).
         :param challenge_queue: The queue containing the challenges.
-        :param max_games: The maximum allowed number of simultaneous games.
+        :param max_bot_games: The maximum allowed number of simultaneous games against bots.
         """
-        max_games_for_matchmaking = max_games if self.matchmaking_cfg.allow_during_games else min(1, max_games)
+        max_games_for_matchmaking = (max_bot_games if self.matchmaking_cfg.allow_during_games
+                                     else min(1, max_bot_games))
         game_count = len(active_games) + len(challenge_queue)
```
MarkZH: I don't think this is right. Imagine the following configuration:
```
challenge:
  concurrency: 10
  games_reserved_for_humans: 4
```
So, `max_bot_games == 6`.

In the highlighted code, `game_count` will be the total number of games--both bot and human. So, if the bot is playing against 6 humans, it will not challenge a bot, even though there are four game slots open. I think the line needs to be `bot_game_count = count_bot_games(active_games) + len(challenge_queue)`--similar to `accept_challenges()` in `lichess_bot.py`.

lib/config.py
```diff
@@ -205,6 +205,7 @@ def insert_default_values(CONFIG: CONFIG_DICT_TYPE) -> None:
     set_config_default(CONFIG, "engine", "polyglot", key="min_weight", default=1)
     set_config_default(CONFIG, "engine", "polyglot", key="normalization", default="none")
     set_config_default(CONFIG, "challenge", key="concurrency", default=1)
+    set_config_default(CONFIG, "challenge", key="games_reserved_for_humans", default=0)
```
MarkZH, line 208: This configuration can interact with this one:
```
challenge:
  preference: "none"/"human"/"bot"
```
If `challenge.game_reserved_for_humans > 0`, there should be a warning if `challenge.preference != "human"`. If you look at `accept_challenges()` in `lichess_bot.py`, you'll see that no challenges are accepted if the front of the queue is a bot challenge and there are no more slots for bot games. If there are human challenges farther back in the queue, then those will be stalled. If `challenge.preference == "human"`, then human challenges will always be at the front, which is probably what the user who reserves games for humans probably wants.

MarkZH:
Also, please update the [Configuration wiki page](https://github.com/lichess-bot-devs/lichess-bot/blob/master/wiki/Configure-lichess-bot.md) with this new option.

lib/matchmaking.py
philipmk42: You're right, good catch — thanks. game_count counting both bot and human games means the reservation incorrectly blocks bot matchmaking when humans fill the slots. I'll change it to count only bot games (count_bot_games(active_games) + len(challenge_queue)), mirroring accept_challenges().

lib/config.py
philipmk42: Makes sense — I'll add a config warning when games_reserved_for_humans > 0 and challenge.preference != "human", since otherwise human challenges can get stalled behind bot ones in the queue. Thanks for pointing out the interaction.

philipmk42:
Will do — I'll update the Configuration wiki page with the new option once the implementation is settled.

philipmk42:
Thanks for the detailed review @MarkZH — both addressed:

- Matchmaking now counts only bot games (is_bot_name over active_games plus the queue) instead of total games, so human-filled slots no longer block bot matchmaking.
- Added a config warning when games_reserved_for_humans > 0 and challenge.preference != "human", since human challenges could otherwise stall behind bot ones in the queue.

I'll update the Configuration wiki page once this is settled.

philipmk42:
Added documentation for games_reserved_for_humans to the Configuration wiki page (wiki/Configure-lichess-bot.md), included in this PR.

lib/matchmaking.py
```diff
@@ -220,18 +220,25 @@ def get_random_config_value(self, config: Configuration, parameter: str, choices
...
         :param challenge_queue: The queue containing the challenges.
-        :param max_games: The maximum allowed number of simultaneous games.
+        :param max_bot_games: The maximum allowed number of simultaneous games against bots.
         """
-        max_games_for_matchmaking = max_games if self.matchmaking_cfg.allow_during_games else min(1, max_games)
-        game_count = len(active_games) + len(challenge_queue)
-        if (game_count >= max_games_for_matchmaking
-                or (game_count > 0 and self.last_challenge_created_delay.time_since_reset() < self.max_wait_time)
+        max_games_for_matchmaking = (max_bot_games if self.matchmaking_cfg.allow_during_games
+                                     else min(1, max_bot_games))
```
MarkZH: I don't think this is right. If `self.matchmaking_cfg.allow_during_games` is `False`, lichess-bot will still issue challenges even if there are active games against humans. I think the easiest way to handle this is:
```python
if not self.matchmaking_cfg.allow_during_games and active_games:
    return
```
Then, use `max_bot_games` for deciding when to challenge.

philipmk42: Good catch — you're right, with allow_during_games: false the bot would still issue challenges while in a game against a human. Fixed by returning early when allow_during_games is false and there are any active games, then using max_bot_games for the bot-game limit. Thanks.

MarkZH:
I tried running this branch and bot games with
```
challenge:
  concurrency: 10
  games_reserved_for_humans: 9
```
Bot games were not limited to one at a time.

I added some log messages and found this:
```
2026-06-27 02:13:15,193 lib.matchmaking (matchmaking.py:250) INFO active_games={'GZnHcpuw': 'BOT sseh-c', 'R9l6ojqK': 'BOT mochi_bot', '4YbjovqL': 'BOT turkjs', '2h1rnRJe': 'BOT Raven030'}
2026-06-27 02:13:15,194 lib.matchmaking (matchmaking.py:251) INFO model.Player.bot_names={'Genetic_Chess_Bot', 'uSunfish-l7', 'mochi_bot', 'ThrorBjorn', 'Humaia', 'Raven030', 'sseh-c', 'GiffiBot', 'is_BOSS', 'turkjs', 'Catie-2_1', 'bfiedler-bot'}
```
Notice that in `active_games`, every bot user name is prefixed with `BOT `, whereas names in `model.Player.bot_names` do not. Due to this mismatch, `count_bot_games()` and line 241 in matchmaking.py always return `0`.

philipmk42:
Great catch, and thanks for digging into the logs to pin it down. You're right — active_games stores the title-prefixed username ("BOT sseh-c") while Player.bot_names holds the bare username ("sseh-c"), so count_bot_games() never matched and always returned 0. Fixed by comparing the bare username (name.split()[-1]) in both count_bot_games() and the matchmaking count. Verified it counts correctly with your example config.

MarkZH:
Good. This works. One more improvement will complete this PR.

Instead of repeating the filtering code in `Matchmaker.challenge()` in line 243, move `count_bot_games()` from `lichess_bot.py` to `model.py` and call the function from both places.

philipmk42:
Done — moved count_bot_games into Player as a classmethod (alongside bot_names/is_bot_name, since it's about bot identity), and both accept_challenges and Matchmaker.challenge now call Player.count_bot_games(active_games). No more duplicated logic.

MarkZH (approved):

philipmk42:
Thanks for the review and approval! Let me know if there's anything else needed before merging.

AttackingOrDefending (approved):

PR #1226, opened 2026-07-19
Auto-block bots that decline challenges

### LINKED ISSUE #1220: Auto-Bot Blocklist Update Support
The current bot blocklist is ephemeral and reset whenever the bot resets. While the "online bot blocklist" mitigates this somewhat, currently this requires manual maintenance or an external script / provider to adjust the list.

Upon a bot refusing a challenge, if configured lichess-bot should edit the online bot blocklist and append the bot to the blocklist.

The current behavior and logging dictates that bots that reject challenges are only temporarily ignored ("Will not challenge BOT [...] to another game today"). Ideally this behavior should be configurable (especially this 1 day TTL), as most bots likely do not change their configurations that often.

This functionality could be implemented externally, but it seems reasonably appropriate for the bot itself to support this functionality.

Ideally, would be great to support S3, LFS, or Gists.

MarkZH:
## Type of pull request:
- [ ] Bug fix
- [x] Feature
- [ ] Other

## Description:

If the user's configuration has `matchmaking: permablock_after_declined_challenge: true`, then any bot that declines a challenge from the user's bot is added to a permanent block list that persists after lichess-bot is shut down.

## Related Issues:

Fixes #1220

## Checklist:

### DIFF
--- a/lib/config.py
+++ b/lib/config.py
@@ -231,6 +231,8 @@ def insert_default_values(CONFIG: CONFIG_DICT_TYPE) -> None:
     set_config_default(CONFIG, "matchmaking", key="block_list", default=[], force_empty_values=True)
     set_config_default(CONFIG, "matchmaking", key="online_block_list", default=[], force_empty_values=True)
     set_config_default(CONFIG, "matchmaking", key="include_challenge_block_list", default=False, force_empty_values=True)
+    set_config_default(
+        CONFIG, "matchmaking", key="permablock_after_declined_challenge", default=False, force_empty_values=True)
     default_filter = (CONFIG.get("matchmaking") or {}).get("delay_after_decline") or FilterType.NONE.value
     set_config_default(CONFIG, "matchmaking", key="challenge_filter", default=default_filter, force_empty_values=True)
     set_config_default(CONFIG, "matchmaking", key="allow_matchmaking", default=False)
--- a/lib/config.py
+++ b/lib/config.py
@@ -231,8 +231,7 @@ def insert_default_values(CONFIG: CONFIG_DICT_TYPE) -> None:
     set_config_default(CONFIG, "matchmaking", key="block_list", default=[], force_empty_values=True)
     set_config_default(CONFIG, "matchmaking", key="online_block_list", default=[], force_empty_values=True)
     set_config_default(CONFIG, "matchmaking", key="include_challenge_block_list", default=False, force_empty_values=True)
-    set_config_default(
-        CONFIG, "matchmaking", key="permablock_after_declined_challenge", default=False, force_empty_values=True)
+    set_config_default(CONFIG, "matchmaking", key="challenge_decliner_file_name", default=False, force_empty_values=True)
     default_filter = (CONFIG.get("matchmaking") or {}).get("delay_after_decline") or FilterType.NONE.value
     set_config_default(CONFIG, "matchmaking", key="challenge_filter", default=default_filter, force_empty_values=True)
     set_config_default(CONFIG, "matchmaking", key="allow_matchmaking", default=False)
--- a/lib/matchmaking.py
+++ b/lib/matchmaking.py
@@ -354,7 +354,7 @@ class Matchmaking:
         self.challenge_type_acceptable[(username, game_aspect)] = timeout_timer
         if add_to_file and self.local_block_list:
             block_expiration = datetime.datetime.strftime(timeout_timer.expiration(), self.datetime_format)
-            self.local_block_list.write_text(f"{self.username()},{game_aspect},{block_expiration}\n")
+            self.local_block_list.write_text(f"{username},{game_aspect},{block_expiration}\n")
 
     def should_accept_challenge(self, username: str, game_aspect: str) -> bool:
         """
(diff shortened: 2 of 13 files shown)


MarkZH:
Currently untested. I'd appreciate comments on whether this is a good approach to implementing this feature.

lib/matchmaking.py
```diff
@@ -50,6 +51,44 @@ def __init__(self, li: Lichess, config: Configuration, user_profile: UserProfile
...
+            with self.local_block_list.open(encoding="utf8") as source:
+                valid_lines: list[str] = []
+                for line_raw in source:
+                    line = line_raw.strip()
+                    if not line:
+                        continue
+
+                    _, _, expire_str = line.split(",")
+                    expiration = datetime.datetime.strptime(expire_str, self.datetime_format)
+                    if expiration < datetime.datetime.now():
```
AttackingOrDefending: This should be flipped I think, as it adds the expired lines into the `valid_lines` list.

lib/matchmaking.py
```diff
@@ -301,7 +345,11 @@ def add_challenge_filter(self, username: str, game_aspect: str, timeout: datetim
         challenge. If the parameter is empty, that is equivalent to adding the opponent to the block list.
         :param timeout: The amount of time to not challenge an opponent. If None, the default is a day.
         """
-        self.challenge_type_acceptable[(username, game_aspect)] = Timer(timeout or days(1))
+        timeout_timer = Timer(timeout or days(1))
+        self.challenge_type_acceptable[(username, game_aspect)] = timeout_timer
+        if add_to_file and self.local_block_list:
+            block_expiration = datetime.datetime.strftime(timeout_timer.expiration(), self.datetime_format)
+            self.local_block_list.write_text(f"{self.username},{game_aspect},{block_expiration}\n")
```
AttackingOrDefending: `username` is a method currently. I think switching to a property is also a good idea.

AttackingOrDefending:
> Currently untested. I'd appreciate comments on whether this is a good approach to implementing this feature.

I think there are a few more stuff but I now realize you would have seen them either way and just asked for the approach. The approach overall looks good. I would prefer avoiding the `try`/`except` blocks, but I am not sure if there is a cleaner way.

lib/config.py
```diff
@@ -231,6 +231,8 @@ def insert_default_values(CONFIG: CONFIG_DICT_TYPE) -> None:
     set_config_default(CONFIG, "matchmaking", key="block_list", default=[], force_empty_values=True)
     set_config_default(CONFIG, "matchmaking", key="online_block_list", default=[], force_empty_values=True)
     set_config_default(CONFIG, "matchmaking", key="include_challenge_block_list", default=False, force_empty_values=True)
+    set_config_default(
+        CONFIG, "matchmaking", key="permablock_after_declined_challenge", default=False, force_empty_values=True)
```
MSWS: I think making it more obvious on what this does would be good, maybe something like `blocklist_persist_file`; no value would be off, and otherwise you can configure the path to use (eg: `blocklist_persist_file=blocked_challenge_decliners.csv`.

lib/matchmaking.py
```diff
@@ -50,6 +53,27 @@ def __init__(self, li: Lichess, config: Configuration, user_profile: UserProfile
 
         self.online_block_list = OnlineBlocklist(self.matchmaking_cfg.online_block_list)
 
+        self.local_block_list: Path | None = None
+        self.datetime_format = "%Y-%m-%d %H:%M:%S.%f"
```
AttackingOrDefending: This isn't used anymore

test_bot/ruff.toml
```diff
@@ -12,7 +12,7 @@ ignore = [
     "COM812", # Trailing comma missing
     "D203", # Require blank line after class declaration before docstring
     "D212", # Start multiline docstring on same line as triple-quote
-    "DTZ005", # datetime without timezone
+    "DTZ", # handling datetimes without timezone
```
AttackingOrDefending: I think it now passes without this change

AttackingOrDefending:
Other than that I think it is good.

lib/config.py
MarkZH: Better?

AttackingOrDefending (approved):

lib/matchmaking.py
```diff
@@ -50,6 +53,26 @@ def __init__(self, li: Lichess, config: Configuration, user_profile: UserProfile
...
 
+        self.local_block_list: Path | None = None
+        self.permablock = bool(self.matchmaking_cfg.challenge_decliner_file_name)
+        if self.permablock:
+            self.local_block_list = Path(self.matchmaking_cfg.challenge_decliner_file_name)
+            self.read_local_block_list()
+
+    def read_local_block_list(self) -> None:
+        """Read the local block list file and reload blocks from previous session."""
+        if not self.local_block_list or not self.local_block_list.is_file():
```
MarkZH:
```python
if not self.local_block_list or not self.local_block_list.exists():
    return

if not self.local_block_list.is_file():
    raise Error
```

AttackingOrDefending (approved):

PR #1236, opened 2026-08-05
Add test coverage for pondering

thomasahle:
The test suite enables `ponder: true` via `config.yml.default`, but none of the test engines ever return a ponder move, so python-chess never starts a ponder search — `go ponder`, `ponderhit`, and the `stop` wind-down are never exercised by any test.

This matters in practice: pondering produces unusually tight command timing (a `position`/`go ponder` pair arrives immediately after the engine emits `bestmove`, and `ponderhit` arrives with no preceding `position`). We recently debugged a production bot whose engine had a race in exactly that window — commands arriving right after `bestmove` were occasionally dropped, causing silent forfeits on time. The current suite structurally can't catch that class of bug for any engine author.

This PR adds two tests following the existing conventions in `test_bot/`:

* `test_uci_ponder` with `ponder_engine.py` (modeled on `uci_engine.py`): predicts the opponent's scripted moves correctly, so every ponder search ends in a `ponderhit`. It responds to commands immediately, so the tight post-`bestmove` timing is exercised for real.
* `test_slow_ponder_engine` with `slow_ponder_engine.py` (in the spirit of `buggy_engine.py`): always predicts a move that is never played and is slow to react to the resulting `stop`, verifying a ponder miss and a sluggish ponder wind-down don't deadlock the game or lose on time.

Both games still end in the scripted scholar's-mate win and PGN write. `pytest`, `ruff check --config test_bot/ruff.toml`, and `mypy --strict .` all pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test_bot/ponder_engine.py
+++ b/test_bot/ponder_engine.py
@@ -1,0 +1,57 @@
+"""An engine that mimics a UCI engine that supports pondering."""
+
+import chess
+import typing
+
+if typing.TYPE_CHECKING:
+    from test_bot.test_games import scholars_mate
+else:
+    from test_games import scholars_mate
+
+assert input() == "uci"
+
+
+def send_command(command: str) -> None:
+    """Send UCI commands to lichess-bot without output buffering."""
+    print(command, flush=True)  # noqa: T201 (print() found)
+
+
+def bestmove_command(board: chess.Board) -> str:
+    """Choose the next scripted move and predict the opponent's reply for pondering."""
+    move_count = len(board.move_stack)
+    command = f"bestmove {scholars_mate[move_count]}"
+    if move_count + 1 < len(scholars_mate):
+        command += f" ponder {scholars_mate[move_count + 1]}"
+    return command
+
+
+send_command("id name UCI_Ponder_Test_Bot")
+send_command("id author lichess-bot-devs")
+send_command("uciok")
+
+board = chess.Board()
+pondering = False
+while True:
+    command, *remaining = input().split()
+    if command == "quit":
+        break
+    elif command == "isready":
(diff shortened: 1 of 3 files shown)


MarkZH:
Do you think there's a way to check if the ponder branch in the engines was actually taken? Perhaps by reading the logs?

thomasahle:
Yes, and good catch — the tests were only checking the outcome of the game, so they would have passed even if no ponder search ever started. I went with structural capture rather than log text: each ponder engine now records its search commands *from inside the branch that handles them* (to a file passed through the `engine_options` config), so an assertion on that list cannot pass unless the branch actually ran. Reading the logs would have been awkward here anyway — each game runs in its own process with queue-based logging, so `caplog` in the pytest process never sees python-chess's engine traffic — and pinned log substrings are more brittle than the command stream itself. `ponder_engine.py` must now produce exactly `go, go, go ponder, ponderhit, go ponder, ponderhit`: no pondering after the first move, then two ponder searches that both end in a `ponderhit`. `slow_ponder_engine.py` must produce exactly `go, go, go ponder, stop, go, go ponder, stop, go`: every guess misses, so each ponder search is stopped and replaced by a real search. As a check that the assertions bite, flipping `ponder` to `false` in either test collapses the log to four plain `go` commands and fails. Added in cc6154d; ruff, `mypy --strict`, and the full test suite are green, and the ponder tests were stable over repeated runs.

AttackingOrDefending:
I think we can remove slow ponder (making the original engine have both ponderhits and misses) and perhaps merge ponder with uci engine to reduce duplication. What do you think @MarkZH ?

PR #1240, opened 2026-08-11
Don't let a failed chat message cancel the move the engine owes

thomasahle:
## Type of pull request:
- [x] Bug fix
- [ ] Feature
- [ ] Other

## Description:

This PR originally bundled three unrelated production fixes. I have split it: it now contains only the chat fix, and the other two are #1241 (games past `challenge.concurrency` sit unplayed) and #1242 (the event stream goes silent and the bot never notices). The three are independent and can be reviewed, changed or merged in any order. Nothing had been reviewed yet, so nothing is lost by the restructuring — apologies for the churn.

### The failure

We run two bots, sunfish-engine and sunfish-nnue-engine, with a greeting configured that turned out to be longer than 140 characters. Lichess answered 400 to the chat POST on every greeted game.

`chat()` logged that an over-long message "will not be sent" and then sent it anyway. The resulting `HTTPError` propagated out of `say_hello()` into `play_game()`'s stream loop — after the game state that prompted the engine's move had already been consumed from the stream. The move was never made, and the state that would have prompted it was gone. Every greeted game died the same way, aborted by lichess "by lack of activity": <https://lichess.org/h4tDwluR>, <https://lichess.org/ZbT4vfAs>.

The over-long greeting is our own misconfiguration. The bug is that a cosmetic chat message could cancel the move the engine owed.

### What changed

Two changes in `chat()`, nine lines:

* An over-long message is now actually skipped — the existing warning ends with a `return` instead of falling through to the POST.
* Request errors are caught in `chat()` itself. Chat is cosmetic; it must never propagate an exception into the game loop.

The second change is the one that matters, and it holds for any chat failure, not just the over-long case: a flaky connection at the moment the bot says "Good luck!" should not cost it the game.

## Related Issues:

None.

## Checklist:

* `pytest` — 55 passed, 1 skipped (`test_lichess`, which needs `LICHESS_BOT_TEST_TOKEN`). Two of those are new, in `test_bot/test_lichess.py`: an over-long message is not POSTed, and a failing POST does not raise. Both fail before the change and pass after.
* `mypy --strict .` — clean.
* Running in production on both bots since 2026-08-11.

Happy to restructure this to your preference — for instance if you would rather have the exception caught in `say_hello()`/`Conversation` than in `chat()`, or want the caught exception list narrowed or widened.

### DIFF
--- a/lib/lichess.py
+++ b/lib/lichess.py
@@ -390,9 +390,17 @@ class Lichess:
             logger.warning(f"This chat message is {len(text)} characters, which is longer "
                            f"than the maximum of {MAX_CHAT_MESSAGE_LEN}. It will not be sent.")
             logger.warning(f"Message: {text}")
+            return
 
         data = {"room": room, "text": text}
-        self.api_post("chat", game_id, data=data)
+        try:
+            self.api_post("chat", game_id, data=data)
+        except (HTTPError, ReadTimeout, RemoteDisconnected, RequestsConnectionError) as e:
+            # Chat is cosmetic. A failed chat message must never propagate an
+            # exception into the game loop, where it would cancel the move the
+            # engine owes (the state that prompted the move has already been
+            # consumed from the game stream, so the move would never be made).
+            logger.warning(f"Could not send chat message to the {room} room of game {game_id}: {e}")
 
     def abort(self, game_id: str) -> None:
         """Aborts a game."""
--- a/lib/lichess.py
+++ b/lib/lichess.py
@@ -396,10 +396,6 @@ class Lichess:
         try:
             self.api_post("chat", game_id, data=data)
         except (HTTPError, ReadTimeout, RemoteDisconnected, RequestsConnectionError) as e:
-            # Chat is cosmetic. A failed chat message must never propagate an
-            # exception into the game loop, where it would cancel the move the
-            # engine owes (the state that prompted the move has already been
-            # consumed from the game stream, so the move would never be made).
             logger.warning(f"Could not send chat message to the {room} room of game {game_id}: {e}")
 
     def abort(self, game_id: str) -> None:
--- a/test_bot/test_lichess.py
+++ b/test_bot/test_lichess.py
@@ -66,7 +66,7 @@ def test_over_long_chat_message_is_not_sent() -> None:
     def record_api_post(path: str, *_args: object, **_kwargs: object) -> None:
         posted.append(path)
 
-    li.api_post = record_api_post  # type: ignore[method-assign, assignment]
+    li.api_post = record_api_post  # type: ignore[assignment]
 
     li.chat("game_id", "player", "x" * (lichess.MAX_CHAT_MESSAGE_LEN + 1))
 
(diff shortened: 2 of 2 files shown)


thomasahle:
Heads-up: I have split this PR. It now contains only the chat fix (one commit, rebased on current master). The other two fixes moved to #1241 (abort games past `challenge.concurrency`) and #1242 (restart when the event stream goes silent), one commit each.

The abort fix in #1241 is not the version that was here: we ran the original in production and its abort-refused fallback turned out to be dangerous, so that path now leaves the game alone and logs loudly instead. The reasoning is in that PR's description.

Nothing had been reviewed yet, so nothing should be lost — sorry for the notification churn.

lib/lichess.py
```diff
@@ -390,9 +390,17 @@ def chat(self, game_id: str, room: str, text: str) -> None:
...
                            f"than the maximum of {MAX_CHAT_MESSAGE_LEN}. It will not be sent.")
             logger.warning(f"Message: {text}")
+            return
 
         data = {"room": room, "text": text}
-        self.api_post("chat", game_id, data=data)
+        try:
+            self.api_post("chat", game_id, data=data)
+        except (HTTPError, ReadTimeout, RemoteDisconnected, RequestsConnectionError) as e:
+            # Chat is cosmetic. A failed chat message must never propagate an
```
AttackingOrDefending: I think we should remove this comment.

test_bot/test_lichess.py
```diff
@@ -57,6 +58,33 @@ def test_challenge_429_without_retry_after_uses_exponential_backoff() -> None:
...
 
+def test_over_long_chat_message_is_not_sent() -> None:
+    """A message longer than the limit is rejected by lichess, so it must not be sent at all."""
+    li = lichess_without_init()
+    posted: list[str] = []
+
+    def record_api_post(path: str, *_args: object, **_kwargs: object) -> None:
+        posted.append(path)
+
+    li.api_post = record_api_post  # type: ignore[method-assign, assignment]
```
AttackingOrDefending: `method-assign` can be removed from the comment.

thomasahle:
Both addressed: the explanatory comment block is removed (the warning log line carries the intent), and the type-ignore is trimmed to `assignment` only.

lib/lichess.py
```diff
@@ -390,9 +390,13 @@ def chat(self, game_id: str, room: str, text: str) -> None:
             logger.warning(f"This chat message is {len(text)} characters, which is longer "
                            f"than the maximum of {MAX_CHAT_MESSAGE_LEN}. It will not be sent.")
             logger.warning(f"Message: {text}")
+            return
 
         data = {"room": room, "text": text}
-        self.api_post("chat", game_id, data=data)
+        try:
+            self.api_post("chat", game_id, data=data)
+        except (HTTPError, ReadTimeout, RemoteDisconnected, RequestsConnectionError) as e:
```
AttackingOrDefending, line 398: I think it should include rate limit errors as well.

AttackingOrDefending:
Looks good overall. I am a bit unsure if we need these tests or if they should be there, but don't have any better ideas now.

PR #1241, opened 2026-08-12
Abort games that exceed challenge.concurrency instead of queuing them

thomasahle:
## Type of pull request:
- [x] Bug fix
- [ ] Feature
- [ ] Other

## Description:

Split out of #1240, which bundled three unrelated production fixes into one PR. The three are independent and can be reviewed, changed or merged in any order. See also #1240 (a failed chat message cancels the move the engine owes) and #1242 (the event stream goes silent and the bot never notices).

### The failure

`start_game_thread()` hands every game to the multiprocessing pool, whose size is `challenge.concurrency` (+1). `accept_challenges()` respects that limit, but bare `gameStart` events bypass it: an opponent accepting one of our outgoing matchmaking challenges just as an incoming challenge is accepted, or games replayed on reconnect. The extra game is then queued behind the games already holding the pool's processes and sits there unplayed until lichess abandons it. The bot is online, its clock is running, and it never moves. We hit this repeatedly running two bots on one host.

### What changed

* `games_in_progress` tracks which games actually hold a pool process. This is not the same set as `active_games`, which also counts accepted challenges whose game has not started yet.
* When a game starts and no process is free, it is aborted immediately. An aborted game is unrated and harmless; a game left to sit until lichess abandons it is neither.
* `accept_challenges()` gains a `pending_outgoing_challenges` argument, so an unanswered outgoing matchmaking challenge reserves a game slot. That closes the race at the source in the common case; the abort is the backstop for the rest.

### The abort-refused path, and why it does nothing

This is the part we got wrong twice in production, so it is worth spelling out.

Lichess answers 400 to an abort once the game has moves. The first version of this patch fell back to **playing** the overflow game with the pool's reserve process. In production that slow-served two games on one core — exactly the harm the concurrency limit exists to prevent.

The second version **resigned** it instead. That was worse. It resigned two games that were alive and going fine — <https://lichess.org/x4E0Q6js> and <https://lichess.org/bQQnPWSb> — because a second lichess-bot instance on the same account had made the moves that made them unabortable.

The conclusion that survived contact with production: a refused abort means the game has moves *this process never made*. Under correct single-serving that is impossible, so it is evidence that something else is serving this account, and then neither playing nor resigning is safe. So the code attempts the abort (the polite, ratingless exit), and if the abort is refused it logs an ERROR naming the game and saying precisely that, frees the slot, and leaves the game alone. Whoever is making those moves keeps the game.

The abort is wrapped so that no exception can escape `start_game_thread()`. An exception there would propagate into the caller and cost the *other* game the move it owes, which is the failure mode fixed in #1240.

## Related Issues:

None.

## Checklist:

* `pytest` — 55 passed, 1 skipped (`test_lichess`, which needs `LICHESS_BOT_TEST_TOKEN`). Three of those are new, in `test_bot/test_start_game.py`: a game within the limit is played, a game past the limit is aborted, and an overflow game whose abort is refused is neither played nor resigned.
* `mypy --strict .` — clean.
* Running in production on two bots since 2026-08-11.

I am happy to restructure this to your preference: rename things, drop the `pending_outgoing_challenges` half if you would rather have only the abort backstop, thread `games_in_progress` through the call chain instead of keeping it at module level next to `correspondence_games_to_start`, or fold the tests in elsewhere.

### DIFF
--- a/lib/lichess_bot.py
+++ b/lib/lichess_bot.py
@@ -344,6 +344,7 @@ def lichess_bot_main(li: lichess.Lichess,
     max_games = config.challenge.concurrency
     reserved_for_humans = min(config.challenge.games_reserved_for_humans, max_games)
     max_bot_games = max_games - reserved_for_humans
+    games_in_progress.clear()  # A restart abandons the old pool and the games it was playing.
 
     one_game_completed = False
 
@@ -387,6 +388,7 @@ def lichess_bot_main(li: lichess.Lichess,
 
             if event["type"] == "local_game_done":
                 active_games.pop(event["game"]["id"], None)
+                games_in_progress.discard(event["game"]["id"])
                 matchmaker.game_done()
                 log_proc_count("Freed", active_games)
                 one_game_completed = True
@@ -422,7 +424,8 @@ def lichess_bot_main(li: lichess.Lichess,
                                              play_game_args,
                                              active_games,
                                              max_games)
-            accept_challenges(li, challenge_queue, active_games, max_games, max_bot_games)
+            accept_challenges(li, challenge_queue, active_games, max_games, max_bot_games,
+                              int(bool(matchmaker.challenge_id)))
             matchmaker.challenge(active_games, challenge_queue, max_bot_games)
             check_online_status(li, user_profile, last_check_online_time)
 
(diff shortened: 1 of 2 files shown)


AttackingOrDefending:
Queuing it and having lichess abort it is better than aborting the game ourselves, as we have the possibility of starting it if another game finishes early.

Issue #677, opened 2023-03-22
Python version deprecation tracking


MarkZH:
This issue is for discussing the scheduling and motivations for deprecating old Python versions.

### Deprecating 3.8 (May 1, 2023)

- [ ] Finish type hinting with `Queue`.

### Deprecating 3.9 (TBD)

- [ ] In timer.py, replace `time.time()` with `time.monotonic()` so timing is not affected by clock changes (DST, e.g.). This function is available on [MacOS starting with version 3.10](https://docs.python.org/3/library/time.html#time.monotonic).

AttackingOrDefending:
With deprecating `3.8`, we can use `list[dict[str, int]]` instead of `List[Dict[str, int]]`, so we can remove `from typing import List`.
With deprecating `3.9`, we can use `int | str` instaed of `Union[int, str]`, and `int | None` instead of `Optional[int]`.

MarkZH:
With respect to `time.monotonic()`, I misinterpreted the Python docs. The [documentation](https://docs.python.org/3/library/time.html#time.monotonic) states that

> Changed in version 3.10: On macOS, the function is now system-wide.

I had thought this meant that the function wasn't guaranteed to be available on macOS. However, according to [this cpython issue](https://github.com/python/cpython/issues/102740), the definition of "system-wide" is

> [...] the clock is the same for all processes on the system and the reference point does not change after start-up time.

Meaning, `time.monotonic()` is available for macOS, but the timers on different processes cannot be used together until 3.10. Since all of the timers used in lichess-bot never leave their originating thread or process, this limitation is irrelevant.

However, `time.monotonic()` has much lower time resolution than `time.perf_counter()`, so the issue is moot and `perf_counter()` is a better choice for timing purposes.

AttackingOrDefending:
With deprecating `3.9` we can use [`TypeAlias`](https://docs.python.org/3/library/typing.html#typing.TypeAlias), so we would change `MOVE = Union[PlayResult, list[Move]]` to `MOVE: TypeAlias = Union[PlayResult, list[Move]]`.

With deprecating `3.10` we can use [`NotRequired`](https://docs.python.org/3/library/typing.html#typing.NotRequired) and [`Required`](https://docs.python.org/3/library/typing.html#typing.Required) in `TypedDict`s.

[...]

PR #1248, opened 2026-08-31
Use NotRequired for optional TypedDict keys

MohKardan:
## Type of pull request:
- [ ] Bug fix
- [ ] Feature
- [x] Other

## Description:

This implements the **"Use NotRequired and Required in TypedDicts"** checklist item from the *Deprecating 3.10* section of #677.

- All `total=False` TypedDicts (18 classes in `lib/lichess_types.py` and `PlayGameArgsType` in `lib/lichess_bot.py`) now use explicit `NotRequired[...]` annotations for every optional key. This is semantically identical, but makes the optionality of each key explicit and allows keys that are always present to be marked `Required` in the future.
- `NotRequired` is imported from `typing` on Python 3.11+ and from `typing_extensions` on Python 3.10. `typing_extensions` is added as a dependency with an environment marker (`python_version < "3.11"`) until 3.10 support is dropped on 2026-10-31 (see my comment on #677). If maintainers prefer waiting until the 3.10 deprecation date passes, switching the imports to plain `typing` and dropping the dependency is a trivial change.

Validation:
- `mypy --strict .` passes on both Python 3.10 and 3.11+ (no issues in 31 source files).
- `pytest`: 53 passed, 1 skipped (the online token test, as expected without `LICHESS_BOT_TEST_TOKEN`).
- No behavior change: all these TypedDicts are populated by casting lichess API JSON responses, so keys are genuinely optional at runtime.

Marked as draft pending feedback on the `typing_extensions` question in #677.

## Related Issues:

Part of #677

## Checklist:


MarkZH:
Since we'll be deprecating Python 3.10, you don't need to check for whether the python version is 3.11 or greater. That's what [this](https://github.com/lichess-bot-devs/lichess-bot/blob/df7e730de58cc3ef2f1415a0dc2eeda842d39167/lib/lichess_bot.py#L1242) code is for.

What you need to do is change the [github workflows](https://github.com/lichess-bot-devs/lichess-bot/tree/master/.github/workflows) so that the oldest version of python testing is for 3.11. Look for:
```
matrix:
  python: ["3.10", ..., "3.14"]
```
and change it to
```
matrix:
  python: ["3.11", ..., "3.15"]
```

Once we approve of your pull request, we won't merge it until the deprecation date.

MohKardan:
Thanks for the review, @MarkZH! All done:

- `NotRequired` is now imported directly from `typing` (removed the `sys.version_info` checks) and the `typing_extensions` dependency is dropped.
- Updated the CI matrices in `python-build.yml`, `python-test.yml` and `mypy.yml` so the oldest tested version is 3.11 and the newest is 3.15.

One small heads-up: `actions/setup-python` may not resolve `"3.15"` until the stable Python 3.15 release in October, so that job could stay red until then. If you prefer, I can switch it to `"3.15-dev"` in the meantime — happy to adjust either way.

And understood — no rush on merging until the 3.10 deprecation date. Marking the PR ready for review.

MohKardan:
Confirmed the diagnosis: `actions/setup-python` can't resolve `"3.15"` yet — the `Set up Python 3.15` step failed, and that failure cancelled all the other matrix jobs (fail-fast is the default).

Switched the matrices to `"3.15-dev"` as suggested above (cceccde), so CI should be green now. Once Python 3.15 is stable in October, I'll change it back to `"3.15"` — or feel free to when merging.

AttackingOrDefending:
Removing `total=False` and adding `NotRequired` is a good start. I think after that we can go in one of two ways:
- Changing `NotRequired` to `Required` for the values needed by lichess-bot to run
- Changing `NotRequired` to `Required` for the values always returned by lichess

I believe option 1 is simpler and easier to maintain, although 2 is more accurate. Can be done in this PR or another one.

